PR #2, opened 2020-09-07
Scholarship list page and component implementation 

edwingl10:
- implemented the scholarship list page and component files
- made some folders in the src directory for better organization 
- made a test scholarship file that contains a list of scholarships and their respective information 
- added some basic styles to the scholarship list page


src/App.js
```diff
@@ -1,25 +1,11 @@
...
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
+    <ScholarshipListPage />
```
josueetcom, line 8: Since we'll probably have a separate home page, could we instead use [React Router](https://reactrouter.com/web/guides/quick-start) to only to route to the Scholarships page if the path matches `/scholarships` or so?

src/pages/ScholarshipListPage.js
```diff
@@ -0,0 +1,26 @@
...
+        id={scholarship.id}
+        name={scholarship.name}
+        amount={scholarship.amount}
+        deadline={scholarship.deadline}
+        website={scholarship.website}
+        />
+    );
+}
+
+function ScholarshipListView(){
```
josueetcom, line 17: *ScholarshipsListView

here and elsewhere where a list of scholarships is referred to

src/components/ScholarshipList.js
```diff
@@ -0,0 +1,17 @@
...
+    return(
+        <div class="scholarship-row">
+            <ul>
+                <li>{props.name}</li>
+                <li>{props.amount}</li>
+                <li>{props.deadline}</li>
+                <li>favorte btn</li>
+            </ul>
+        </div>
+    );
```
josueetcom, line 14: Maybe instead of taking in individual scholarships, this component can take in a list of scholarships and return the entire list? I'm looking at the [Basic List Component](https://reactjs.org/docs/lists-and-keys.html#basic-list-component) section from the [List and Keys](https://reactjs.org/docs/lists-and-keys.html) concepts page. 

I think also we might also need [Keys](https://reactjs.org/docs/lists-and-keys.html) on each of the items.

src/components/scholarshipList.css
```diff
@@ -0,0 +1,12 @@
...
+    display: flex;
+    justify-content: space-between;
+}
+.scholarship-row:after{
+    display: block;
+    height: 6px;
+    background-color: #F5B700;
+    content:"";
+    width: 100%;
+}
```
josueetcom, line 12: (Feel free to address this in a follow-up PR instead) You could look into [Material UI](https://material-ui.com/)'s components to see if there's something there that helps display the content more cleanly. There's a Data Display section, that includes things like [Lists](https://material-ui.com/components/lists/) and [Tables](https://material-ui.com/components/tables/) which might come in handy.

josueetcom (changes requested):
Thanks for getting this going! I just have a few nits about overall structure.

Also, I noticed that the repo was forked. I've added you to the organization so you can create branches/commits on the main repo instead

PR #8, opened 2020-09-17
Contact/about pages, scholarship filters

edwingl10:
- created contact page (go to /contact)
- created about page (go to /about)
- non functional filter can be found in scholarships page


src/pages/Contact.jsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import TextField from '@material-ui/core/TextField';
+import TextareaAutosize from '@material-ui/core/TextareaAutosize';
+import Button from '@material-ui/core/Button';
+import { Link } from 'react-router-dom';
+
+function Contact() {
+  return (
+    <div>
```
josueetcom, line 9: Instead of a `<div>` have you looked into using a [`<Container>`](https://material-ui.com/components/container/#container) for these pages?

src/pages/About.jsx
```diff
@@ -0,0 +1,69 @@
...
+        </Grid>
+        <Grid item xs={10} sm={6} md={4}>
+          <AboutCard img={testPic} name="Gonzalo Lara" description="Eiusmod tempor ipsum nostrud quis commodo enim enim voluptate cillum laboris esse esse velit tempor. " />
+        </Grid>
+        <Grid item xs={10} sm={6} md={4}>
+          <AboutCard img={testPic} name="Sergio Mejia" description="Anim dolor ad exercitation eu. Do minim ullamco ea esse qui duis incididunt." />
+        </Grid>
+        <Grid item xs={10} sm={6} md={4}>
+          <AboutCard img={testPic} name="Alejandro Loaiza" description="Ad laboris duis nostrud eiusmod incididunt adipisicing ut. Anim qui deserunt incididunt ullamco." />
+        </Grid>
```
josueetcom, line 55: I suggest using a loop here to generate the repetitive <Grid> elements. You could extract the data out into an object outside the function, e.g.:
```js
const developers = {
  "Josue Rios": {
    img: testPic,
    description: "some description"
  },
  "Edwin Lopez": {
    // properties
  },
  // others
}
```

src/pages/About.jsx
```diff
@@ -0,0 +1,69 @@
...
+        In cupidatat amet eu excepteur reprehenderit quis enim id velit labore
+        duis culpa. Ex culpa occaecat velit consequat eu.
+        Labore eiusmod pariatur occaecat aliqua reprehenderit laboris.
+        Officia qui ad do magna excepteur proident laborum ea laborum consectetur aliqua pariatur.
+      </p>
+
+      <h2>Meet the Developers</h2>
+      <Grid container justify="space-evenly" spacing={3}>
+        <Grid item xs={10} sm={6} md={4}>
+          <AboutCard img={testPic} name="Josue Rios" description="Deserunt proident dolor ad anim exercitation duis et laboris ipsum nisi cillum." />
```
josueetcom, line 42: When it comes to populating information like this, I think it's generally better if we instead leave a description of what it's supposed to contain and/or file [issues](https://github.com/beyondhb1079/s4us/issues) to figure out the details later. E.g. the description could be something like `TODO: About Josue Rios` or `TODO(#9): About Josue Rios` and so forth. I've filed issues for all of us to come up with a 2-3 sentence description of ourselves.

src/components/Filter.jsx
```diff
@@ -0,0 +1,69 @@
...
+    <div>
+      <FormControl variant="outlined" className={classes.formControl}>
+        <InputLabel>Major</InputLabel>
+        <Select value={major} onChange={selectMajor} label="Major">
+          <MenuItem value=""><em>None</em></MenuItem>
+          <MenuItem value="Chemical Eng.">Chemical Eng.</MenuItem>
+          <MenuItem value="Computer Science">Computer Science</MenuItem>
+          <MenuItem value="Data Science">Data Science</MenuItem>
+        </Select>
+      </FormControl>
```
josueetcom, line 42: These three dropdowns all look very similar as well as their items. I suggest refactoring this into its own component, e.g. `filterDropdown` or `FilterDropdown` that takes in a property (maybe called `items`?) of the format `{key: value}`, e.g. `{12: "High School Senior"}` and loops through those to generate the `<MenuItem>`. The component doesn't need to live in a separate file but if you think it could be used outside of this `Filter` component then I'd place it in its own file.

src/components/Filter.jsx
```diff
@@ -0,0 +1,69 @@
...
+import Select from '@material-ui/core/Select';
+
+const useStyles = makeStyles((theme) => ({
+  formControl: {
+    margin: theme.spacing(1),
+    minWidth: 120,
+  },
+}));
+
+function Filter() {
```
josueetcom, line 16: Maybe a better name might be `FilterBar`?

josueetcom (changes requested):
I've done an initial pass. One more suggestion I have is to have separate branches and PRs for unrelated changes. E.g. You could have a PR adding the About page, another one adding the Contact Us page, and another one adding the Scholarships Filter. This has several advantages:
- You and your reviewers have less to think about per PR since they're smaller.
- Smaller PRs means there's less comments/feedback per PR, allowing the more straightforward to get merged sooner.
- Helps isolate controversial changes in a single PR. E.g. If the About and Contact Us page look fine but there are still some comments on Scholarship filters, those two pages could've been merged in separate PRs.

src/pages/Contact.jsx
```diff
@@ -0,0 +1,50 @@
...
+      <h3>Reach Out</h3>
+      <p>
+        We are happy to answer any of your questions. Fill out the form and we&apos;ll be in
+        touch as soon as possible.
+      </p>
+      <form>
+        <div><TextField id="standard-basic" label="full name" /></div>
+        <div><TextField id="standard-basic" label="email" /></div>
+        <div><TextField id="standard-basic" label="subject" /></div>
+        <div><TextareaAutosize rowsMin={10} placeholder="Message" /></div>
```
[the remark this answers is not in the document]
edwingl10: So we'll just display one Multiline TextField where the user can input all their message content?

src/pages/Scholarships.jsx
```diff
@@ -1,14 +1,16 @@
 import React from 'react';
 import scholarships from '../testdata/scholarships';
 import ScholarshipList from '../components/ScholarshipList';
+import FilterBar from '../components/FilterBar';
 
-function ScholarshipListPage() {
+function Scholarships() {
   return (
     <div>
```
josueetcom, line 8: Please use a `<Container>` here as well

src/pages/Contact.jsx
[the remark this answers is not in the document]
josueetcom: Yeah, you should be able to specify the number of rows like in the right most examples of the link above.

josueetcom (changes requested):
Just minor changes then we should be able to get this merged.

gonzalo1330:
I took a look at the deploy and it might be nice to have the [Start Your Search, About, Contact] on the header? To be able to always go back and forth between them instead of having to go to the homepage to find those links.

src/components/Header.jsx
```diff
@@ -1,9 +1,26 @@
...
 import LoginDialog from './LoginDialog';
 
+const buttonStyles = {
+  marginRight: '5px',
+};
+
 function Header() {
   return (
     <div className="header">
+      <Link to="/scholarships" style={buttonStyles}>
```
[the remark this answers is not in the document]
josueetcom: Same for the link below

josueetcom (approved):

PR #33, opened 2020-09-22
Fixed Login Button bug

gonzalo1330:


src/components/LoginButton.jsx
```diff
@@ -0,0 +1,35 @@
+import React, { useState } from 'react';
+import firebase from 'firebase';
+import Button from '@material-ui/core/Button';
+import LoginDialog from './LoginDialog';
+
+export default function LoginButton() {
+  const { currentUser } = firebase.auth(); // determines how to initialize isSignedIn
+  const [isSignedIn, setLoggedIn] = useState(!!currentUser);
+  const [loginDialogOpen, setDialogOpen] = useState(false); // initialized to false
```
josueetcom, line 9: Please fix these to use the `const [X, setX]` naming pattern seen throughout the [React docs](https://reactjs.org/docs/hooks-state.html).

src/components/LoginButton.jsx
```diff
@@ -0,0 +1,35 @@
...
+  const signUserOut = () => firebase.auth().signOut();
+  const showLoginDialog = () => setDialogOpen(true);
+
+  return (
+    <div>
+      <Button
+        variant="outlined"
+        color="secondary"
+        onClick={isSignedIn ? signUserOut : showLoginDialog}
+      >
```
josueetcom, line 29: Does this fit in a single line without breaking the line limit? If so I would do that

src/components/LoginButton.jsx
```diff
@@ -0,0 +1,35 @@
...
+        {isSignedIn ? 'Sign out' : 'Login'}
+      </Button>
+      <LoginDialog open={loginDialogOpen} onClose={false} />
```
josueetcom, line 32: `onClose` probably shouldn't be false. The dialog should probably close when the user tries to close the dialog without logging in.

src/components/LoginDialog.jsx
```diff
@@ -1,33 +1,48 @@
...
-import FirebaseLoginScreen from './FirebaseLoginScreen';
+import firebase from 'firebase';
+import StyleFirebaseAuth from 'react-firebaseui/StyledFirebaseAuth';
+import PropTypes from 'prop-types';
 
-export default function LoginDialog() {
-  const [open, setOpen] = React.useState(false);
-  const handleClickOpen = () => setOpen(true);
-  const handleClose = () => setOpen(false);
+class LoginDialog extends React.Component {
```
josueetcom, line 11: This can be written as a [function component](https://reactjs.org/docs/components-and-props.html)

src/components/LoginDialog.jsx
```diff
@@ -1,33 +1,48 @@
...
       </Dialog>
-    </div>
-  );
+    );
+  }
 }
+
+LoginDialog.propTypes = {
+  open: PropTypes.element.isRequired,
+  onClose: PropTypes.element.isRequired,
```
josueetcom, line 45: Based on the [PropTypes](https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes) page these should be of type `PropTypes.func`

src/pages/About.jsx
```diff
@@ -20,7 +20,11 @@ const developers = [
   {
     name: 'Gonzalo Lara',
     img: testPic,
-    description: 'Todo: about Gonzalo Lara',
+    description: 'My passion for computers has lead me to pursue a career in Software '
+      + 'Engineering. I am currently a UWB CSS student completing my final year in the '
+      + 'Computer Science and Software Engineering program. I am 24 years old, bilingual '
+      + 'and currently a part of the DACA program.  I intend to contribute as much as '
+      + 'possible to building a better tomorrow.',
```
josueetcom, line 27: Better to leave this in the other PR so that it gets merged quickly. It's also not relevant to the Login Button fix.

josueetcom (changes requested):

src/components/TestData.jsx
```diff
@@ -0,0 +1,21 @@
...
+  render() {
+    return (
+      <div>
+        <h1>Scholarship List from Firestore!</h1>
+      </div>
+    )
+  }
+}
+
+export default TestData;
```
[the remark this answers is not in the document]
gonzalo1330: Yeah its outdated and deleted from this PR

src/pages/Home.jsx
```diff
@@ -1,9 +1,11 @@
 import React, { } from 'react';
+import Data from '../components/TestData';
```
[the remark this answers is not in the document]
gonzalo1330: Same as comment above

src/components/LoginButton.jsx
gonzalo1330: Fixed

src/components/LoginButton.jsx
gonzalo1330: Ok I had it like that but when I did the format document it broke them into this, ill fix it manually

src/components/LoginButton.jsx
```diff
@@ -0,0 +1,35 @@
+import React, { useState } from 'react';
+import firebase from 'firebase';
+import Button from '@material-ui/core/Button';
+import LoginDialog from './LoginDialog';
+
+export default function LoginButton() {
+  const { currentUser } = firebase.auth(); // determines how to initialize isSignedIn
```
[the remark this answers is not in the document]
gonzalo1330: Hm, made the change and it cause a lint error and its fixable via lint but convert it to `const { currentUser } = firebase.auth();`

src/components/LoginButton.jsx
gonzalo1330: Yeah I ran into this issue that it does not close if the user does not want to login, any recommendation for how to fix?

src/components/LoginDialog.jsx
gonzalo1330: Doing it that way causes lint errors

src/pages/About.jsx
gonzalo1330: Ok I did that

src/components/LoginDialog.jsx
josueetcom: What lint errors does it give?

src/components/LoginButton.jsx
josueetcom: Instead of `false` pass it a function that'll close the login dialog

src/components/LoginButton.jsx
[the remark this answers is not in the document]
josueetcom: What was the lint error?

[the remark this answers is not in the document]
josueetcom: If it was a `destructuring-props` issue then I would just use `firebase.auth().currentUser` where `currentUser` is used since It's only used in one place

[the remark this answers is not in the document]
josueetcom: Please don't mark the conversation resolved unless the comment was actually addressed. In this case, there's still open questions.

src/components/LoginButton.jsx
josueetcom: Generally speaking, it's also not good practice to mark a PR comment as fixed unless the fix can actually be seen in the code on here

gonzalo1330: *changed

src/components/LoginDialog.jsx
gonzalo1330: 44:3  error  propType "open" is not required, but has no corresponding defaultProps declaration     react/require-default-props
  45:3  error  propType "onClose" is not required, but has no corresponding defaultProps declaration  react/require-default-props

src/components/LoginButton.jsx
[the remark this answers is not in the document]
gonzalo1330: 7:9  error  Use object destructuring  prefer-destructuring

src/components/LoginButton.jsx
gonzalo1330: Ok nice, I passed in this function -> `const handleClose = () => setLoginDialogOpen(false);` and it now works when you click outside the dialog it'll close

src/components/LoginDialog.jsx
gonzalo1330: lines 11-13 make them as a function?

src/components/LoginDialog.jsx
edwingl10: Instead of using element as the PropType, use func
`open: PropTypes.func.isRequired` 
`onClose: PropTypes.func.isRequired`

gonzalo1330: Got it and no lint error, thanks Edwin

src/components/LoginDialog.jsx
gonzalo1330: Oh wait you mean `LoginDialog` to be written as a function rather than a class?

edwingl10: And you would place lines 11 - 13 inside of your LoginDialog function

josueetcom (approved):

PR #41, opened 2020-09-24
Destructure scholarship properties

josueetcom:
This should make it easier to use those values when we change things up later


edwingl10 (approved):
Ahh yes, way cleaner!

PR #51, opened 2020-09-26
refactored brand name

edwingl10:
- can be found in src/config/constants.js

### DIFF
--- a/src/config/constants.js
+++ b/src/config/constants.js
@@ -1,4 +1,2 @@
-const constants = {
-  brandName: 'UndocuScholars',
-};
-export default constants;
+/* eslint-disable import/prefer-default-export */
+export const BRAND_NAME = 'Undocuscholars';
--- a/src/config/constants.js
+++ b/src/config/constants.js
@@ -1,0 +1,4 @@
+const constants = {
+  brandName: 'UndocuScholars',
+};
+export default constants;
--- a/src/pages/About.jsx
+++ b/src/pages/About.jsx
@@ -41,10 +41,10 @@ function About() {
   return (
     <Container>
       <h1 style={{ textAlign: 'center' }}>
-        {`About ${BRAND_NAME}`}
+        About {BRAND_NAME}
       </h1>
       <h2>
-        {`Scholarship Search With ${BRAND_NAME}`}
+        Scholarship Search With {BRAND_NAME}
       </h2>
       <p>
         Est elit eu mollit reprehenderit enim nisi ea velit mollit dolore.
@@ -64,7 +64,7 @@ function About() {
       </p>
 
       <h2>
-        {`How ${BRAND_NAME} Works`}
+        How {BRAND_NAME} Works
       </h2>
       <p>
         Laboris amet nostrud amet anim fugiat deserunt ad laboris dolore sint eiusmod.
--- a/src/pages/About.jsx
+++ b/src/pages/About.jsx
@@ -5,6 +5,7 @@ import Button from '@material-ui/core/Button';
 import Container from '@material-ui/core/Container';
 import AboutCard from '../components/AboutCard';
 import testPic from '../logo.svg';
+import constants from '../config/constants';
 
 /* eslint-disable max-len */
 const developers = [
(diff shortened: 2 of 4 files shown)


src/pages/About.jsx
```diff
@@ -58,7 +67,13 @@ function About() {
         Reprehenderit enim aute commodo cupidatat ut ipsum ea minim magna velit.
       </p>
 
-      <h2>How UndocuScholar Works</h2>
+      <h2>
+        How
+        {' '}
+        {constants.brandName}
+        {' '}
+        Works
```
josueetcom, line 75: Can't this just be `How {constants.brandName} Works`?

src/pages/Contact.jsx
```diff
@@ -28,7 +29,11 @@ function Contact() {
 
       <h3>Learn About Us</h3>
       <p>
-        You can learn more about UndocuScholar and the developers
+        You can learn more about
+        {' '}
+        {constants.brandName}
+        {' '}
+        and the developers
```
josueetcom, line 36: Same question here

src/config/constants.js
```diff
@@ -0,0 +1,4 @@
+const constants = {
+  brandName: 'UndocuScholars',
+};
+export default constants;
```
josueetcom, line 4: I recommend disabling the single export per file rule here and just exporting the constant:
```suggestion
/* eslint-disable import/prefer-default-export */
export const BRAND_NAME = 'Undocuscholars';
```
Otherwise it's possible to set `constants.brandName` even though it should be a constant (the only thing constant about a JS object is the reference).

Also, per [23.10](https://airbnb.io/javascript/#naming--uppercase) of the AirBnB style guide exported constants should be uppercased. It helps visually see when a constant is being used.

src/pages/About.jsx
```diff
@@ -39,8 +40,16 @@ const developers = [
...
-      <h2>Scholarship Search With UndocuScholar</h2>
+      <h1 style={{ textAlign: 'center' }}>
+        About
+        {' '}
+        {constants.brandName}
+      </h1>
+      <h2>
+        Scholarship Search With
+        {' '}
+        {constants.brandName}
```
josueetcom, line 51: Similar question here as my other comment later in the file

src/pages/About.jsx
edwingl10: I had it like that but the lint --fix changes it to that format. I'll change it back.

src/config/constants.js
edwingl10: I changed it because of the lint error but now that I know how to disable it, I'll revert it

src/pages/About.jsx
edwingl10: should i disable 'react/jsx-one-expression-per-line', if not I get a lint error

josueetcom: Hmm, if we do that it might make HTML subelements harder to read later. Maybe instead in `package.json` in the `eslintConfig` section you can add this:
```json
    "rules":{
      "quotes": ["error", "single", { "allowTemplateLiterals": true }]
    }
```
And then use a template here to make this read more easily and occupy less lines:
```js
{`How {constants.brandName} Works`}
```

src/config/constants.js
josueetcom: Yeah, the rule is there to help encourage single exports per file but it doesn't make sense in the case of this file.

src/pages/About.jsx
```diff
@@ -39,8 +40,12 @@ const developers = [
 function About() {
   return (
     <Container>
-      <h1 style={{ textAlign: 'center' }}>About UndocuScholar</h1>
-      <h2>Scholarship Search With UndocuScholar</h2>
+      <h1 style={{ textAlign: 'center' }}>
+        {`About ${BRAND_NAME}`}
```
josueetcom, line 44: This actually doesn't look as clean as I thought 😦 . Maybe instead templates like this we can just use normal JSX but disable the one expression per line rule? In the `eslintConfig.rules` section you can add `"react/jsx-one-expression-per-line": "off"`

josueetcom: This might even fit in one line with the `<h1>` tags

josueetcom (approved):
This looks good! Thanks!

josueetcom (approved):

Issue #61, opened 2020-09-28
Scholarships details page is crashing


josueetcom:
E.g. https://dreamerscholars.web.app/scholarships/3

PR #67, opened 2020-09-29
scholarships bug fix

edwingl10:


src/pages/ScholarshipDetails.jsx
```diff
@@ -10,18 +11,20 @@ function ScholarshipDetailsPage() {
     return (<h1>Scholarship Not Found</h1>);
   }
 
-  const [name, amount, deadline, website, school, year, description] = scholarship;
```
josueetcom, line 13: Ohh was this the bug??

josueetcom (approved):

josueetcom:
Update the description to say **fixes #61**. GitHub will then automatically close the issue when this PR is merged

src/pages/ScholarshipDetails.jsx
edwingl10: Yes, The {} were needed for destructuring.

josueetcom: Dang, the reason I had used `[]` was to avoid creating the new lines. Perhaps we can set **object-curly-newline.ObjectPattern** to `"never"` in the ESLint config so it allows this to be a single line with curly braces? I'm looking at
https://eslint.org/docs/rules/object-curly-newline#objectexpression-and-objectpattern

edwingl10: Oh yeah I was thinking of using that

PR #74, opened 2020-10-06
Add TypeScript to the project

### LINKED ISSUE #73: Add TypeScript support to project
So we can have strongly typed models. JS/JSX will continue to be supported. But if we want to we can fully migrate all the code over to TS/TSX later

josueetcom:
Also added a prototype Scholarship class.

Fixes #73 

Please leave a comment on anything you find confusing or want to learn more about

### DIFF
--- a/src/models/Scholarship.ts
+++ b/src/models/Scholarship.ts
@@ -1,0 +1,29 @@
+import scholarships from '../testdata/scholarships';
+
+interface ScholarshipProps {
+  // TODO(https://github.com/beyondhb1079/s4us/issues/56):
+  // Update this to reflect the schema
+  name: string;
+  description: string;
+  deadline: string;
+  website: string;
+  school: string;
+  year: string;
+}
+
+export default class Scholarship {
+  id?: string;
+  data: ScholarshipProps;
+
+  constructor(data: ScholarshipProps) {
+    this.data = { ...data };
+  }
+
+  static get(id: string): Scholarship | undefined {
+    const scholarshipList = scholarships as { [id: string]: ScholarshipProps };
+    if (id in scholarshipList) {
+      return new Scholarship(scholarshipList[id]);
+    }
+    return undefined;
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ package-lock.json
 .env.development.local
 .env.test.local
 .env.production.local
+src/react-app-env.d.ts
 
 npm-debug.log*
 yarn-debug.log*
--- a/package.json
+++ b/package.json
@@ -8,7 +8,12 @@
     "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^11.0.4",
     "@testing-library/user-event": "^12.1.6",
+    "@types/jest": "^26.0.14",
+    "@types/node": "^14.11.2",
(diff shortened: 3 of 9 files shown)


src/models/Scholarship.ts
```diff
@@ -0,0 +1,21 @@
+interface ScholarshipProps {
+  // TODO(https://github.com/beyondhb1079/s4us/issues/56):
+  // Update this to reflect the schema
+  id?: string;
```
edwingl10, line 4: Is the use of the ? to check if we provided an id, and if not it doesn't create that field?

src/models/Scholarship.ts
```diff
@@ -0,0 +1,21 @@
...
+  name: string;
+  description: string;
+  deadline: string;
+  website: string;
+  school: string;
+  year: string;
+}
+
+export default class Scholarship {
+  id?: string;
```
edwingl10, line 14: How come we need another id when the ScholarshipProps object already has one?

src/models/Scholarship.ts
```diff
@@ -0,0 +1,21 @@
...
+  data: ScholarshipProps;
+
+  constructor(data: ScholarshipProps) {
+    this.data = { ...data };
```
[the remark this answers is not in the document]
edwingl10: ahh so in this case we are copying the information from the data parameter in the constructor into the data field?

src/models/Scholarship.ts
josueetcom: `?` makes an [optional property](https://www.typescriptlang.org/docs/handbook/advanced-types.html#optional-parameters-and-properties) here, which means it can be null/undefined. I imagine we're going to want to do that for almost every property except the minimum required ones that must be present.

edwingl10 (approved):

src/models/Scholarship.ts
josueetcom: What I initially had in this PR was a bit offset of what I intended. What I mean to have here is an id property separate from the the construction of a scholarship (since the ID is generated, not provided). The latest version of this PR reflects that.

src/models/Scholarship.ts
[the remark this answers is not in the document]
josueetcom: Yes that is correct!

edwingl10 (approved):

src/models/Scholarship.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  static get(id: string): Scholarship | undefined {
+    const scholarshipList = scholarships as { [id: string]: ScholarshipProps };
```
edwingl10, line 23: I was having a little trouble making sense of this but if I understand correctly, the 'as' is telling the compiler to interpret the scholarship as an object with key of type string and value of type ScholarshipProps. 
I'm wondering why the id is wrapped with []

josueetcom: Yeah, it's a little confusing to read. I'm not that familiar with TypeScript so I'm not sure if this is the right way to write this but it looks like this is what's called an [Index Type](https://www.typescriptlang.org/docs/handbook/advanced-types.html#index-types-and-index-signatures). TypeScript also has Map and Record types as well though so maybe one of those is better for this case. We can submit this as is though and as we learn more about TypeScript correct this as needed

PR #77, opened 2020-10-10
User class

edwingl10:
not entirely sure how to write the get method yet.

### DIFF
--- a/src/models/User.ts
+++ b/src/models/User.ts
@@ -1,0 +1,28 @@
+interface UserProps {
+    age: number;
+    schoolYear: number;
+    role: string;
+    gender: string;
+    scholarshipPreferences: {
+        type: string,
+        major: string,
+        amount: number, 
+        school: string,
+        grade: number,
+        location: string
+    };
+    otherSettings: {
+        notifications: boolean,
+        deadlineNotifications: boolean,
+        relatedNotifications: boolean
+    }
+}
+
+export default class User{
+    id?:string;
+    data: UserProps
+
+    constructor(data:UserProps){
+        this.data = { ...data };
+    }
+}


src/models/User.ts
```diff
@@ -1,28 +1,32 @@
...
+  data: UserProps;
 
-    constructor(data:UserProps){
-        this.data = { ...data };
-    }
-}
\ No newline at end of file
+  constructor(data:UserProps) {
+    this.data = { ...data };
+  }
```
josueetcom, line 31: Hmm, I think this constructor should be private as it doesn't make sense for to create users. Rather, what we can do is have a `getCurrentUser()` method that loads data based on [`firebase.auth().currentUser`](https://firebase.google.com/docs/auth/web/manage-users#get_the_currently_signed-in_user) and another `get(id: string)` to fetch data for a different user.

src/models/User.ts
```diff
@@ -1,28 +1,32 @@
...
+  role: Role;
+  gender: Gender;
+  scholarshipPreferences: {
+    type: ScholarshipType,
+    major: string,
+    amount: number,
+    school: string,
+    schoolYear: Grade,
+    location: string
+  };
```
josueetcom, line 17: I would create a separate interface type for this and the `otherSettings` type

src/models/User.ts
```diff
@@ -1,28 +1,32 @@
...
-        school: string,
-        grade: number,
-        location: string
-    };
-    otherSettings: {
-        notifications: boolean,
-        deadlineNotifications: boolean,
-        relatedNotifications: boolean
-    }
+  age: number;
```
josueetcom, line 6: This will get outdated within a year. Maybe instead store the birth date? Or if we don't want to ask for a specific date we can at least ask for the birth year and possibly the birth month too

src/models/enums.ts
```diff
@@ -0,0 +1,25 @@
+export enum Grade{
```
josueetcom, line 1: I suggest adding a default `Unknown` value in case we encounter an unknown value.

src/models/enums.ts
```diff
@@ -0,0 +1,25 @@
...
+  Reader,
+  Contributor,
+  Moderator,
+  Admin
+}
+export enum Gender{
+  Male,
+  Female,
+  Other
+}
```
josueetcom, line 19: Also an `Unknown` value here as none of these values are good defaults.

src/models/enums.ts
```diff
@@ -0,0 +1,25 @@
...
+export enum ScholarshipType{
+  Academic,
+  Sports,
+  Community,
+  Organization
+}
```
josueetcom, line 25: `Unknown` value here too

src/models/enums.ts
```diff
@@ -0,0 +1,25 @@
+export enum Grade{
+  HighschoolJunior,
+  HighschoolSenior,
+  CollegeFreshman,
+  CollegeSophmore,
+  CollegeJunior,
+  CollegeSenior
+}
+export enum Role{
```
josueetcom, line 9: I suggest leaving this in `User.ts` as it's specific to that

src/models/User.ts
edwingl10: Having them in the same file would be ok, right?

src/models/User.ts
```diff
@@ -1,32 +1,73 @@
...
+interface ScholarshipPreferences{
+  type: ScholarshipType;
+  major: string[] | undefined;
+  amount: {
+    type: ScholarshipAmount | undefined;
+    min: number | undefined;
+    max: number | undefined;
   }
+  school: string[];
+  schoolYear: Grade;
```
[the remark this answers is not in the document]
edwingl10: My idea behind this was if a user was also interested in scholarships for next year. 
But I think it makes more sense using the filters for that.

src/models/User.ts
```diff
@@ -0,0 +1,73 @@
...
+  type: ScholarshipType;
+  major: string[] | undefined;
+  amount: {
+    type: ScholarshipAmount | undefined;
+    min: number | undefined;
+    max: number | undefined;
+  }
+  school: string[];
+  schoolYear: Grade;
+  location: string[];
```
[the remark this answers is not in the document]
edwingl10: For sure, I'll change it.

src/models/User.ts
```diff
@@ -1,32 +1,73 @@
...
+  uid: string | null,
+}
+interface ScholarshipPreferences{
+  type: ScholarshipType;
+  major: string[] | undefined;
+  amount: {
+    type: ScholarshipAmount | undefined;
+    min: number | undefined;
+    max: number | undefined;
   }
```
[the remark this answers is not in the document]
edwingl10: If it's FIXED, either the min,max or both can store the amount.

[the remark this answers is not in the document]
josueetcom: Hmm alright. I think both storing the value is fine

src/models/User.ts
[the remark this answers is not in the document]
josueetcom: Ohh I see, this might be relevant then if they're considering multiple schools where `Grade` is different. Otherwise I would expect this to just be the current Grade+1

josueetcom (commented):
I see most of comments I left marked as resolved. I don't see any changes though, did you mean to push what you have?

edwingl10:
> I see most of comments I left marked as resolved. I don't see any changes though, did you mean to push what you have?

I'm still working on it. I marked them resolve so I can keep track of which ones I've fixed

josueetcom:
Ah, in that case it's still good to push what you have. In the **Reviewers** section of the PR there's a **Re-request review** button you can click when it's ready for me to take another look

src/models/User.ts
```diff
@@ -0,0 +1,67 @@
+import firebase from 'firebase';
+import {
+  Grade, ScholarshipType, Gender, ScholarshipAmount, States,
+} from './enums';
+
+export enum Role{
```
josueetcom, line 6: I'm surprised the linter doesn't complain about the missing space before the curly brace. Does `yarn lint --fix` do this?
```typescript
export enum Role {
```

src/models/enums.ts
```diff
@@ -0,0 +1,83 @@
+export enum Grade{
+  UNKNOWN,
+  HIGHSCHOOL_JUNIOR,
+  HIGHSCHOOL_SENIOR,
+  COLLEGE_FRESHMAN,
+  COLLEGE_SOPHOMORE,
+  COLLEGE_JUNIOR,
+  COLLEGE_SENIOR
+}
+export enum Gender{
```
josueetcom, line 10: I'm also surprised it didn't complain about the spacing between `enum`s. Does `yarn lint --fix` fix this?

src/models/enums.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+export enum ScholarshipAmount{
+  UNKNOWN,
+  RANGE,
+  FULLRIDE,
+  FIXED
+}
+
+export enum States{
```
josueetcom, line 31:
```suggestion
export enum State{
```

src/models/User.ts
edwingl10: Usually `lint --fix` adds the space. I forgot to run the command so most of my code will have lint errors, I'll fix them with the next push.

src/models/enums.ts
```diff
@@ -0,0 +1,83 @@
+export enum Grade {
+  UNKNOWN,
+  HIGHSCHOOL_JUNIOR,
+  HIGHSCHOOL_SENIOR,
+  COLLEGE_FRESHMAN,
+  COLLEGE_SOPHOMORE,
+  COLLEGE_JUNIOR,
+  COLLEGE_SENIOR,
+}
+export enum Gender {
```
josueetcom, line 10: Please add a space before this

src/models/enums.ts
```diff
@@ -0,0 +1,83 @@
...
+  UNKNOWN,
+  MALE,
+  FEMALE,
+  OTHER,
+}
+export enum ScholarshipType {
```
josueetcom, line 16: here too.

src/models/User.ts
```diff
@@ -0,0 +1,81 @@
...
+  gender: Gender;
+  school?: string;
+  state?: string;
+  majors?: string[];
+  scholarshipPrefs: ScholarshipPreferences;
+  notificationSettings: NotificationSettings;
+}
+
+interface ScholarshipPreferences {
+  type: ScholarshipType[];
```
josueetcom, line 36: Should this be a single type?

src/models/User.ts
```diff
@@ -0,0 +1,81 @@
...
+  major?: string[];
```
josueetcom, line 37: *majors

src/models/User.ts
```diff
@@ -0,0 +1,81 @@
...
+  amount: {
+    type: ScholarshipAmount;
+    min?: number;
+    max?: number;
+  };
```
josueetcom, line 42: maybe `ScholarshipAmount` should be this type? And then `ScholarshipAmountType` a separate type

src/models/User.ts
```diff
@@ -0,0 +1,81 @@
...
+  State,
+} from './enums';
+
+export enum Role {
+  READER,
+  CONTRIBUTOR,
+  MODERATOR,
+  ADMIN,
+}
+export enum NotificationType {
```
josueetcom, line 17: Please add a space before this line.

src/models/User.ts
```diff
@@ -0,0 +1,81 @@
...
+interface NotificationSettings {
+  notifications: NotificationType;
+  deadlineNotifications: NotificationType;
+  relatedNotifications: NotificationType;
+}
+
+export const converter: firestore.FirestoreDataConverter<UserData> = {
+  toFirestore: (data: UserData) => ({
+    ...data,
+  }),
```
josueetcom, line 55: This can be written
```js
  toFirestore: (data: UserData) => ({...data}),
```

src/models/User.ts
```diff
@@ -0,0 +1,81 @@
...
+    return { ...data, scholarshipPrefs, notificationSettings } as UserData;
+  },
+};
+
+export default class User extends FirestoreModel<UserData> {
+  static get collection(): firestore.CollectionReference<UserData> {
+    return firestore().collection('users').withConverter(converter);
+  }
+
+  private constructor(id?: string, data?: UserData) {
```
josueetcom, line 69: I would simply leave this as 
```suggestion
  private constructor(id: string, user?: firebase.User) {
```
We are not generating IDs for users. Similarly since we're not creating new users there is no data to pass in to save a new user. We can at most expect to read other user's data.

josueetcom: Though this does make me question the design decision I made around `new Scholarship(undefined, data)` and `new Scholarship(id)`. I wonder if instead of having constructors like that if static methods like `Scholarship.new(data)` and `Scholarship.get(id)` would read better. Or maybe even having a separate `Scholarships` class and making those methods read `Scholarships.get(id)` and `Scholarships.new()` 🤔

src/models/User.ts
```diff
@@ -0,0 +1,81 @@
...
+    super(ref, data ?? ({} as UserData));
+  }
+
+  static getCurrentUser(): firebase.User | undefined {
+    const user = firebase.auth().currentUser;
+    if (user) {
+      return user;
+    }
+    return undefined;
+  }
```
josueetcom, line 80: You want to return a `User` here, so probably something like:
```ts
  static currentUser(): User | undefined {
    const user = firebase.auth().currentUser;
    return user ?  new User(user.uid, user) : undefined;
  }
```

package.json
```diff
@@ -16,6 +16,7 @@
     "eslint-config-airbnb-typescript": "^12.0.0",
     "eslint-plugin-jest": "^24.0.2",
     "firebase": "^7.19.1",
+    "prettier": "^2.1.2",
```
josueetcom, line 19: not needed. it's already in `devDependencies`

josueetcom (changes requested):

PR #78, opened 2020-10-10
scholarship form page

edwingl10:
- basic scholarship form page


src/pages/ScholarshipForm.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+function ScholarshipForm() {
+  return (
+    <Container>
+      <h1>Submit a Scholarship</h1>
+      <FormControl>
+        <TextField id="standard-basic" label="scholarship name *" />
+        <TextField id="standard-basic" label="Deadline *" />
+        <TextField id="standard-basic" label="amount *" />
+        <TextField id="standard-basic" label="website url *" />
```
josueetcom, line 15: Is there a way to have the labels appear above the text field? I'm looking at https://material-ui.com/components/text-fields/#form-props

Also, the form should be wrapped in a [<form>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form) element

src/App.jsx
```diff
@@ -18,6 +19,7 @@ function App() {
     <Router>
       <Header />
       <Switch>
+        <Route path="/scholarships/new" component={ScholarshipForm} />
```
josueetcom, line 22: Can we add a button for this in the header? Otherwise we have to remember the url path and it's not discoverable

src/pages/ScholarshipForm.jsx
```diff
@@ -1,20 +1,28 @@
...
-      <FormControl>
-        <TextField id="standard-basic" label="scholarship name *" />
-        <TextField id="standard-basic" label="Deadline *" />
-        <TextField id="standard-basic" label="amount *" />
-        <TextField id="standard-basic" label="website url *" />
+      <form>
+        <TextField className={classes.fieldStyle} id="name" label="Scholarship Name" type="text" InputLabelProps={{ shrink: true }} fullWidth />
+        <TextField className={classes.fieldStyle} id="deadline" label="Deadline" type="text" InputLabelProps={{ shrink: true }} fullWidth />
+        <TextField className={classes.fieldStyle} id="amount" label="Amount" type="text" InputLabelProps={{ shrink: true }} fullWidth />
+        <TextField className={classes.fieldStyle} id="website" label="Website URL" type="text" InputLabelProps={{ shrink: true }} fullWidth />
```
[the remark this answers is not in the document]
josueetcom: This was resolved but I still see the old style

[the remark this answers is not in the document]
edwingl10: I haven't pushed my changes, still working on it.

src/pages/ScholarshipForm.jsx
```diff
@@ -4,26 +4,153 @@ import TextField from '@material-ui/core/TextField';
 import Button from '@material-ui/core/Button';
 import { makeStyles } from '@material-ui/core/styles';
 
-const useStyles = makeStyles(({
+const ScholarshipFields = {
+  1: {
```
josueetcom, line 8: Rather than numbers I would use the key/id of each of these elements

src/pages/ScholarshipForm.jsx
```diff
@@ -0,0 +1,202 @@
...
+  KeyboardDatePicker,
+} from '@material-ui/pickers';
+import { ScholarshipFields } from './formFields';
+
+const useStyles = makeStyles({
+  fieldStyle: {
+    margin: '10px 0;',
+  },
+  formControlStyle: {
+    width: '40%',
```
josueetcom, line 27: I think this should be wider or left alone. On mobile it looks too skinny

src/pages/ScholarshipForm.jsx
```diff
@@ -0,0 +1,202 @@
...
+      case 'amountSub':
+        return displayAmountFields();
+
+      default:
+        return <div />;
+    }
+  }
+
+  return (
+    <Container style={{ width: '40%' }}>
```
josueetcom, line 185: This may not look great on mobile devices. I suggest maybe leaving this as is or finding a different way to set width for just desktop

src/pages/ScholarshipForm.jsx
josueetcom: I think you want
```suggestion
    maxWidth: 'md',
```

josueetcom: Based on https://material-ui.com/api/container/#css

package.json
```diff
@@ -3,19 +3,23 @@
...
     "@testing-library/user-event": "^12.1.6",
     "@types/node": "^14.11.2",
     "@types/react": "^16.9.49",
     "@types/react-dom": "^16.9.8",
+    "date-fns": "^2.16.1",
     "eslint-config-airbnb": "^18.2.0",
     "eslint-config-airbnb-typescript": "^12.0.0",
     "eslint-plugin-jest": "^24.0.2",
     "firebase": "^7.19.1",
+    "prettier": "^2.1.2",
```
josueetcom, line 22: Not needed. It's already in the `devDependencies` section

src/pages/formFields.js
```diff
@@ -0,0 +1,109 @@
+export const ScholarshipFields = {
+  name: {
+    key: 1,
```
josueetcom, line 3: `key` is not needed here. The `key` in list elements can simply be the key here (`name`)

src/pages/ScholarshipForm.jsx
```diff
@@ -0,0 +1,202 @@
...
+      />
+    ));
+  }
+
+  function displayField(fieldKey) {
+    const { key } = ScholarshipFields[fieldKey];
+    const { label } = ScholarshipFields[fieldKey];
+    const { type } = ScholarshipFields[fieldKey];
+
+    switch (type) {
```
josueetcom, line 105: Rather than having a `displayField` with a switch statement I recommend having the [elements](https://en.reactjs.org/docs/rendering-elements.html) themselves be the value in the `formFields` mapping and creating Function components for any common ones. E.g. 
```jsx
function TextField(props) {
  // Returns a <TextField> according to the passed in props
}

const formFields = {
  name: <TextField your custom props/>, // Alternatively TextField({ custom props })
  date: <MuiPickersUtilsProvider utils={DateFnsUtils} ...>
           <KeyboardDatePicker  .../>
        </MuiPickersUtilsProvider>,
  ...
};
```

src/pages/ScholarshipForm.jsx
```diff
@@ -0,0 +1,202 @@
...
+      case 'text':
+      case 'number':
+        return (
+          <TextField
+            key={key}
+            id={fieldKey}
+            label={label}
+            className={classes.fieldStyle}
+            type={type === 'text' ? 'text' : 'number'}
```
josueetcom, line 114:
```suggestion
            type={type}
```

src/pages/ScholarshipForm.jsx
```diff
@@ -0,0 +1,202 @@
...
+        InputLabelProps={{ shrink: true }}
+        fullWidth
+        InputProps={{ inputProps: { min: 0, step: 50 } }}
+        value={formFieldStates[field]}
+        onChange={(e) =>
+          handleFieldChange(field, parseInt(e.target.value, 10) || '')
+        }
+      />
+    ));
+  }
```
josueetcom, line 98: I recommend writing a separate `<ScholarshipAmountInput>` component instead that knows how to render this stuff. It could take in the current scholarship amount and also an onChange handler that takes in a `ScholarshipAmount` object.

josueetcom (changes requested):
Left some initial feedback. Mostly on the structure.

src/pages/ScholarshipForm.jsx
edwingl10: I didn't think of that but it looks cleaner! How would I pass in State and onChange handler to the elements though?

src/pages/ScholarshipForm.jsx
josueetcom: This wasn't updated

josueetcom (commented):
I left some comments, but seeing how fragile it is to add all these different form fields, I think it'd be best to split these up into multiple PRs instead (we can additionally break up the related issue into smaller digestable issues). E.g.
- Basic form page with the required fields
- Separate PRs for each input _type_. It may make sense to group some of the changes (e.g. eligibility requirements) but one field is complicated enough maybe it should be its own PR.

src/pages/ScholarshipForm.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <Container style={{ width: '40%' }}>
+      <h1>Submit a Scholarship</h1>
+      <form noValidate autoComplete="off">
+        {Object.entries(ScholarshipFields).map(
+          ([field, component]) => component
+        )}
```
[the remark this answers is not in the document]
edwingl10: I tried that way and it didn't render anything. Maybe it's because I didn't forward the refs? Regardless, I'm trying it again and will let you know my status on it.

src/pages/formFields.jsx
```diff
@@ -0,0 +1,154 @@
...
+  const { id, label, type } = properties;
+  return (
+    <TextField
+      key={id}
+      id={id}
+      label={label}
+      type={type}
+      InputLabelProps={{ shrink: true }}
+      fullWidth
+      InputProps={{ inputProps: { min: 0, step: 50 } }}
```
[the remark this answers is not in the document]
edwingl10: the `min: 0` is so that the value amount of scholarships can't be below 0. The `step:50` is so that when a user clicks on the up/down arrow for changing values, it increases/decreases by 50

src/pages/ScholarshipForm.jsx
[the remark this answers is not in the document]
josueetcom: What about
```jsx
        {Object.entries(ScholarshipFields).map(
          ([name, Field]) => <Field id={name} key={name}/>
        )}
```

If that doesn't work then I think it's because `field` is an object rather than a class. Hmm, in that case I would forget about this suggestion.

src/pages/formFields.jsx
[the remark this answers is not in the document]
josueetcom: That should only be relevant for amount fields though. I would leave that for the Amount text field only. Or better yet, a separate `ScholarshipAmountTextField` function component.

src/pages/formFields.jsx
```diff
@@ -0,0 +1,154 @@
...
+
+function TextFieldComp(properties) {
+  const { id, label, type } = properties;
+  return (
+    <TextField
+      key={id}
+      id={id}
+      label={label}
+      type={type}
+      InputLabelProps={{ shrink: true }}
```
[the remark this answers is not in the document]
edwingl10: Isn't this so that the label appears on top of the `TextField`

PR #88, opened 2020-10-14
Create base model interface and Firestore model class

josueetcom:
The `FirestoreModel` class does the heavy lifting! The subclasses just need to implement a friendlier contructor.

### DIFF
--- a/src/models/FirestoreModel.ts
+++ b/src/models/FirestoreModel.ts
@@ -1,0 +1,50 @@
+import { firestore } from 'firebase';
+import Model from './Model';
+
+export default class FirestoreModel<T> implements Model<T> {
+  constructor(private readonly ref: firestore.DocumentReference<T>, public readonly data: T) {}
+
+  public get id(): string {
+    return this.ref.id;
+  }
+
+  get(): Promise<FirestoreModel<T>> {
+    return new Promise((resolve, reject) => {
+      this.ref.get()
+        .then((doc: firestore.DocumentSnapshot<T>) => {
+          if (!doc.exists) {
+            reject(new Error(`${this.ref.path} not found`));
+            return;
+          }
+          resolve(new FirestoreModel<T>(doc.ref, doc.data() as T));
+        })
+        .catch((error) => reject(error));
+    });
+  }
+
+  save(): Promise<FirestoreModel<T>> {
+    return new Promise((resolve, reject) => {
+      this.ref.set(this.data)
+        .then(() => resolve(this))
+        .catch((error) => reject(error));
+    });
+  }
+
+  delete(): Promise<void> {
+    return new Promise((resolve, reject) => {
+      this.ref.delete()
+        .then(() => resolve())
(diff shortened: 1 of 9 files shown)


src/models/FirestoreModel.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  get(): Promise<FirestoreModel<T>> {
+    return new Promise((resolve, reject) => {
+      this.ref.get()
+        .then((doc: firestore.DocumentSnapshot<T>) => {
+          if (!doc.exists) {
+            reject(new Error(`${this.ref.path} not found`));
+            return;
+          }
+          resolve(new FirestoreModel<T>(doc.ref, doc.data() as T));
```
edwingl10, line 19: `doc.data() as T`
Would this interpret the data as our class models? e.g. User

src/models/FirestoreModel.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  delete(): Promise<void> {
+    return new Promise((resolve, reject) => {
+      this.ref.delete()
+        .then(() => resolve())
+        .catch((error) => reject(error));
+    });
+  }
+
+  subscribe(onChange: (model: Model<T>) => void, onError = console.error): () => void {
```
edwingl10, line 41: I'm having trouble understanding what this does, what happens when there is a change?

src/models/FirestoreModel.ts
josueetcom: Yes that's right

src/models/FirestoreModel.ts
josueetcom: This is a wrapper around [`DocumentReference.onSnapshot`](https://firebase.google.com/docs/reference/js/firebase.firestore.DocumentReference#onsnapshot), which is used to [listen for realtime updates](https://firebase.google.com/docs/firestore/query-data/listen).

Basically what happens is whenever an update to the document happens (e.g. either on the initial call or due to other updates), if the document exists `onChange` gets invoked with a `FirestoreModel` object that contains the data.

josueetcom: [Here](https://github.com/beyondhb1079/s4us/blob/95bd0c1a7d51a05d40ec66ebceae1226edd4d697/src/pages/ScholarshipDetails.jsx#L12-L17)'s how it's used in the follow-up PR

edwingl10: Ohh I see, that makes sense!

edwingl10 (approved):

src/models/FirestoreModel.ts
josueetcom: Yep! One more note. The returned function is a function to detach the listener, in this case it's just the function returned by `DocumentReference.onSnapshot`. That detach function should be invoked whenever a component unmounts and/or needs to be updated with a `onChange` listener.

edwingl10 (approved):

PR #90, opened 2020-10-17
Update LoginButton.jsx to use React effect hooks

josueetcom:
There were technically two bugs here:
- on every render we were registering a listener on auth state changed
- if/when this component unmounts each listener would _still_ listen for auth changes and trying to set state of the unmounted component

Using [`useEffect`](https://en.reactjs.org/docs/hooks-effect.html) here allows us to register a single listener and return a cleanup function that unregisters the listener when the component unmounts.


gonzalo1330 (approved):

gonzalo1330:
Oh okay nice, good catch

Issue #59, opened 2020-09-28
Add a banner somewhere on the page linking to the PR when running on Render


josueetcom:
This'll allow easy navigation back to GitHub

josueetcom:
Closed via #112

PR #112, opened 2020-10-22
Add an alert banner in the header when rendering from Render

josueetcom:
You can see it at the top here:
https://s4us-pr-112.onrender.com/


edwingl10 (approved):

josueetcom:
Close #59

PR #113, opened 2020-10-22
Stylize App + header with basic theme

### LINKED ISSUE #17: Site Header

### LINKED ISSUE #111: Create a basic theme
Material UI supports [theming](https://material-ui.com/customization/theming/) so that we don't have to copy/paste styling everywhere. Material UI components will use that theme. We need to define a custom theme with our own colors, fonts, etc. For now we can just wrap a [ThemeProvider](https://material-ui.com/customization/theming/#theme-provider) around the application with the primary and secondary colors from the design

josueetcom:
- Created a basic [theme](https://material-ui.com/customization/theming/). This'll allow us to change various things app-wide as opposed to writing CSS for all of it.
- Stylized the header according to the current design. For spacing/font differences I went with the defaults but if we really want to we can change it via the theme and/or additional styling.
- Changed the PR preview alert ([see page bottom here](https://s4us-pr-113.onrender.com/)) to be a snackbar at the bottom instead so it's less disruptive and dismissable.
- Fixed a slight styling issue with the LoginDialog because we were using a `<div>`. We should use a [`<React.Fragment>`](https://en.reactjs.org/docs/fragments.html) (shorthand `<>`) instead when we need to return a JSX fragment with multiple children.

Fixes #111 and fixes #17


edwingl10 (approved):

gonzalo1330:
It appears this PR won't be able to get squashed and merged since it is out-of-date with the base branch. I recommend you merge the latest changes from main into this branch (which there is a button to `Update branch` but Im not too sure if it will cause any conflicts so I will leave how it is. 
Aside from that I took a look at the styling which there won't be much of .CSS no more correct? It will mainly be styled from the theme imported and Container, the render looks good btw.

josueetcom:
Yeah styling should be done using `makeStyles` and colors/spacing/etc. should be determined by the theme

PR #129, opened 2020-10-24
TextField functional component

edwingl10:


src/components/TextFieldComp.jsx
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <TextField
+      id={id}
+      label={label}
+      type="text"
+      value={value}
+      onChange={handleFieldChange}
+      fullWidth
+    />
+  );
```
josueetcom, line 20: This could all be simplified to:
```suggestion
  return (
    <TextField {...props} fullWidth/>
  );
```
At which point I don't believe a separate function component is worth it.

src/pages/ScholarshipForm.jsx
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <Container maxWidth="md">
+      <form>
+        <h1>Submit a Scholarship</h1>
+        <TextFieldComp
+          id="name"
+          label="Scholarship Name *"
+          value={formFieldStates.name}
+          onChange={handleFieldChange}
+        />
```
josueetcom, line 32: These could be written using `<TextField>` instead:
```jsx
<TextField id="name" label="Scholarship Name *" value={state.name} onChange={updateFn('name')}
```
Where `updateFn` could be
```jsx
function updateFn(id) {
  return (e) => setState({...state, id: e.target.value});
}
```

src/pages/ScholarshipForm.jsx
```diff
@@ -0,0 +1,52 @@
...
+    minAmount: 0,
+    maxAmount: 0,
+    website: '',
+    scholarshipType: '',
+    active: false,
+  });
+
+  function handleFieldChange(name, value) {
+    setFormFieldStates((prev) => ({ ...prev, [name]: value }));
+  }
```
josueetcom, line 21: Instead of having a `TextFieldComp` component that accepts this, this could return a function that plain `TextField` accepts. E.g.

```suggestion
  function onChangeFn(name) {
    return (e) => setFormFieldStates({ ...state, name: e.target.value});
  }
```

I gave a similar example in the comment below

josueetcom (changes requested):

src/pages/ScholarshipForm.jsx
```diff
@@ -0,0 +1,52 @@
...
+      setFormFieldStates({ ...formFieldStates, [id]: e.target.value });
+  }
+
+  return (
+    <Container maxWidth="md">
+      <form>
+        <h1>Submit a Scholarship</h1>
+        <TextField
+          id="name"
+          label="Scholarship Name *"
```
josueetcom, line 29: Instead of `*` use the `required` attribute

src/pages/ScholarshipForm.jsx
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from 'react';
+import { Container, Button, TextField } from '@material-ui/core';
+
+function ScholarshipForm() {
+  const [formFieldStates, setFormFieldStates] = useState({
```
josueetcom, line 5: Could this state could be initialized to an empty object or just an object with the deadline?

src/pages/ScholarshipForm.jsx
```diff
@@ -0,0 +1,52 @@
...
+          label="Description *"
+          value={formFieldStates.description}
+          onChange={updateFn('description')}
+        />
+        <TextField
+          id="website"
+          label="Website *"
+          value={formFieldStates.website}
+          onChange={updateFn('website')}
+        />
```
josueetcom, line 44: optional: one slight improvement might be to have a helper function that returns a TextField with the `value` and `onChange` set based on the `id` so you can write this as:

```
{requiredTextField('website', 'Website')}
```

This would be different from a TextField function component class as it would be scoped within this function so that it can access state.

josueetcom (approved):
Looks good but there might be slight improvements to make

src/pages/ScholarshipForm.jsx
edwingl10: It gives me the following warning: `Warning: A component is changing an uncontrolled input of type text to be controlled.`

src/pages/ScholarshipForm.jsx
edwingl10: Oo yeah that looks cleaner

josueetcom (approved):

PR #130, opened 2020-10-26
deadline field added to scholarship form

edwingl10:


josueetcom (changes requested):
Just a minor nit to make the date component might be reusable. Maybe it can be called `DateField`?

src/pages/ScholarshipForm.jsx
```diff
@@ -35,9 +42,21 @@ function ScholarshipForm() {
     <Container maxWidth="md">
       <form>
         <h1>Submit a Scholarship</h1>
-        {requiredTextField('name', 'Scholarship Name')}
-        {requiredTextField('description', 'Description')}
-        {requiredTextField('website', 'Website')}
+        <MuiPickersUtilsProvider utils={DateFnsUtils} key="date">
```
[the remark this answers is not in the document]
edwingl10: lol sounds good

src/pages/ScholarshipForm.jsx
```diff
@@ -38,25 +38,31 @@ function ScholarshipForm() {
     );
   }
 
+  function requiredDeadlineField(id, label) {
```
josueetcom, line 41: This works but I was thinking more of a separate file so the component could be reused in another form.

src/pages/ScholarshipForm.jsx
```diff
@@ -38,25 +38,31 @@ function ScholarshipForm() {
...
+    return (
+      <MuiPickersUtilsProvider utils={DateFnsUtils}>
+        <KeyboardDatePicker
```
josueetcom, line 44: I think there's a `required` property you can set here

josueetcom (approved):

src/components/DateField.jsx
```diff
@@ -0,0 +1,33 @@
...
+import DateFnsUtils from '@date-io/date-fns';
+import 'date-fns';
+import {
+  MuiPickersUtilsProvider,
+  KeyboardDatePicker,
+} from '@material-ui/pickers';
+import PropTypes from 'prop-types';
+
+function DateField(props) {
+  const { id, label, value, handler } = props;
```
josueetcom, line 11: I suggest taking in the onChange function directly. The caller already knows what the id is because they passed that in

src/components/DateField.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+  return (
+    <MuiPickersUtilsProvider utils={DateFnsUtils}>
+      <KeyboardDatePicker
+        id={id}
```
josueetcom, line 16: Use `{...props}` to pass in fields of the same name, i.e. id, label, value, onChange

josueetcom (changes requested):

package.json
```diff
@@ -83,6 +83,7 @@
         }
       ],
       "react/jsx-one-expression-per-line": "off",
+      "react/jsx-props-no-spreading": "off",
```
josueetcom: Rather than this I recommend disabling for _just_ the `DateField` file and explaining why it's disabled

src/components/DateField.jsx
```diff
@@ -5,29 +5,13 @@ import {
   MuiPickersUtilsProvider,
   KeyboardDatePicker,
 } from '@material-ui/pickers';
-import PropTypes from 'prop-types';
 
 function DateField(props) {
```
josueetcom, line 11: Maybe `DatePicker` instead of `DateField` to help show the connection with the pickers themselves.

src/pages/ScholarshipForm.jsx
```diff
@@ -42,8 +42,8 @@ function ScholarshipForm() {
           id="deadline"
           label="Deadline"
           value={formFieldStates.deadline}
-          handler={(id, date) =>
-            setFormFieldStates({ ...formFieldStates, [id]: date })
+          onChange={(date) =>
+            setFormFieldStates({ ...formFieldStates, deadline: date })
```
josueetcom, line 46: perfect!

josueetcom (changes requested):

package.json
edwingl10: This was the old version. What I ended up doing was  setting ` "explicitSpread": "ignore"`

package.json
```diff
@@ -80,6 +84,12 @@
         }
       ],
       "react/jsx-one-expression-per-line": "off",
+      "react/jsx-props-no-spreading": [
+        2,
```
josueetcom, line 88: I think you can say "error" as well. Makes it a bit more readable

src/components/DatePicker.jsx
```diff
@@ -0,0 +1,30 @@
...
+import DateFnsUtils from '@date-io/date-fns';
+import 'date-fns';
+import {
+  MuiPickersUtilsProvider,
+  KeyboardDatePicker,
+} from '@material-ui/pickers';
+import PropTypes from 'prop-types';
+import exact from 'prop-types-exact';
+
+function DateField(props) {
```
josueetcom, line 11: *DatePicker

josueetcom (commented):
Just some minor nits.

josueetcom (dismissed):

src/components/DatePicker.jsx
edwingl10: oops don't know how that slipped. I fixed it in the latest push (:

josueetcom (approved):

PR #135, opened 2020-10-27
scholarship type field added

edwingl10:


src/pages/ScholarshipForm.jsx
```diff
@@ -25,11 +33,28 @@ function ScholarshipForm() {
...
+            return (
+              <MenuItem key={option} value={option}>
+                {option}
+              </MenuItem>
+            );
+          })}
+        </Select>
+      </FormControl>
+    );
+  }
```
josueetcom, line 57: This would be a good component to refactor out. Maybe something like `EnumSelectField.jsx`

src/pages/ScholarshipForm.jsx
```diff
@@ -38,9 +63,18 @@ function ScholarshipForm() {
...
-        <Button variant="contained" color="primary">
-          Submit
-        </Button>
+        {scholarshipTypeField('scholarshipType', 'Scholarship Type', [
+          'unknown',
+          'academic',
+          'sports',
+          'community',
+          'organization',
+        ])}
```
josueetcom, line 72: Two things:
1. These values should come from an enum type. 
2. I'm not sure if these types make the most sense.

edwingl10: I'm going to make a PR adding the enums to a file so that I can use them here.
What should the types be? There are the types from the model.

josueetcom: The enum change doesn't necessarily need to be a separate PR if the changes are small enough and no other changes depend on it

src/pages/ScholarshipForm.jsx
```diff
@@ -38,9 +41,21 @@ function ScholarshipForm() {
...
+          label="Scholarship Type"
+          value={formFieldStates.scholarshipType}
+          onChange={updateFn('scholarshipType')}
+          enums={ScholarshipType}
+        />
+        <div>
+          <Button
+            variant="contained"
+            color="primary"
+            onClick={() => console.log(formFieldStates)}>
```
josueetcom, line 55: Hmm, I think you can write the actual submission logic using `Scholarship.new()`. Though since we haven't figured out roles yet maybe we can add the `verificationStatus` enum + field and have it set it to Unverified.

josueetcom: Or if you want to leave the functionality for a different PR/issue, maybe a [Snackbar](https://material-ui.com/components/snackbars/#customized-snackbars) might be a better way to indicate work is still in progress.

src/components/EnumSelectField.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { FormControl, InputLabel, Select, MenuItem } from '@material-ui/core';
+import PropTypes from 'prop-types';
+
+function EnumSelectField(props) {
+  const { label, value, onChange, enums } = props;
+  return (
+    <FormControl>
```
josueetcom, line 8: I suggest adding a `minWidth` style attribute here

src/pages/ScholarshipForm.jsx
```diff
@@ -38,9 +41,21 @@ function ScholarshipForm() {
...
-          Submit
-        </Button>
+        <EnumSelectField
+          id="scholarshipType"
+          label="Scholarship Type"
+          value={formFieldStates.scholarshipType}
+          onChange={updateFn('scholarshipType')}
+          enums={ScholarshipType}
+        />
+        <div>
```
josueetcom, line 51: Instead of a div I suggest looking into surrounding the form elements in a [Grid](https://material-ui.com/components/grid/) and setting the `spacing`

josueetcom (changes requested):

src/pages/ScholarshipForm.jsx
josueetcom: Or alternatively, remove the `<div>` here and set `display="block"` for the button

josueetcom: Or rather for the element above the button

josueetcom: Or I guess a div works too 🤷

josueetcom:
Whoops, I accidentally merged main. You can delete my commit by force pushing your branch `git push -f`.

src/pages/ScholarshipForm.jsx
edwingl10: I'm going to leave it for another PR, so that I can have all the required fields finished first. For now I'll put the SnackBar.

PR #153, opened 2020-10-31
Scholarhip amount fields

edwingl10:


src/types/AmountType.ts
```diff
@@ -0,0 +1,8 @@
+enum AmountType {
+  Unknown = 'UNKNOWN',
+  Fixed = 'FIXED',
+  Range = 'RANGE',
+  Fullride = 'FULLRIDE',
```
josueetcom, line 5:
```suggestion
  FullRide = 'FULL_RIDE',
```

src/pages/ScholarshipForm.jsx
```diff
@@ -20,6 +20,20 @@ function ScholarshipForm() {
...
+      ...formFieldStates,
+      minAmount: value,
+      maxAmount: value,
+    });
+  }
+  function updateAmount(id) {
+    return (e) =>
+      setFormFieldStates({
+        ...formFieldStates,
+        [id]: parseInt(e.target.value, 10) || '',
```
josueetcom, line 34: Is e.target.value a string for number TextField?

src/pages/ScholarshipForm.jsx
```diff
@@ -48,9 +62,20 @@ function ScholarshipForm() {
...
-          Submit
-        </Button>
+        <ScholarshipAmountField
+          amountType={formFieldStates.amountType}
+          minAmount={formFieldStates.minAmount}
+          maxAmount={formFieldStates.maxAmount}
+          onTypeChange={updateFn('amountType')}
+          onMinChange={updateAmount('minAmount')}
+          onMaxChange={updateAmount('maxAmount')}
+          updateMinMax={updateMinMax}
```
josueetcom, line 72: You could instead provide a single `updateMinMax` function that accepts a min and max as parameters. The other functions could be derived from it

src/components/ScholarshipAmountField.jsx
```diff
@@ -0,0 +1,87 @@
...
+    minAmount,
+    maxAmount,
+    onTypeChange,
+    onMinChange,
+    onMaxChange,
+    updateMinMax,
+  } = props;
+
+  function displayAmountFields() {
+    if (amountType === AmountType.Fixed) {
```
josueetcom, line 32: I suggest using a switch statement instead of if/else

src/components/ScholarshipAmountField.jsx
```diff
@@ -0,0 +1,87 @@
...
+      return (
+        <TextField
+          label="Amount"
+          type="number"
+          InputLabelProps={{ shrink: true }}
+          fullWidth
+          InputProps={{ inputProps: { min: 0, step: 50 } }}
+          value={minAmount}
+          onChange={(e) => updateMinMax(parseInt(e.target.value, 10) || '')}
```
josueetcom, line 41: Would it perhaps be useful to refactor a separate `AmountTextField` component that has most of these properties set? It could do the integer parsing as well and set e.target.value to a number or `''`

src/components/ScholarshipAmountField.jsx
```diff
@@ -0,0 +1,87 @@
...
+      ));
+    }
+    return <div />;
+  }
+
+  return (
+    <FormControl className={classes.formControl}>
+      <InputLabel shrink>Amount Type</InputLabel>
+      <Select value={amountType} onChange={onTypeChange}>
+        {Object.keys(AmountType).map((option) => {
```
josueetcom, line 66: The EnumTextField component from the other PR would be useful here

src/components/ScholarshipAmountField.jsx
```diff
@@ -0,0 +1,87 @@
...
+  } = props;
+
+  function displayAmountFields() {
+    if (amountType === AmountType.Fixed) {
+      return (
+        <TextField
+          label="Amount"
+          type="number"
+          InputLabelProps={{ shrink: true }}
+          fullWidth
```
josueetcom, line 38: I suggest not making these full width for visual consistency. Imagine you have lots of form fields below and then you toggle this. All of the sudden everything below shifts up/down. Take a look at https://images.app.goo.gl/c4vCztKPDjp8sz5u5 to see what I mean

src/components/ScholarshipAmountField.jsx
josueetcom: Alternatively instead of a drop-down you could also use four radio buttons with any related form elements next to them. The form elements can be disabled if their radio option isn't set.

josueetcom (changes requested):

src/components/ScholarshipAmountField.jsx
```diff
@@ -23,65 +24,80 @@ function ScholarshipAmountField(props) {
...
-    onMinChange,
-    onMaxChange,
-    updateMinMax,
+    updateAmount,
   } = props;
 
+  function amountTextField(label, value, onChange) {
+    return (
+      <TextField
+        className={classes.formControl}
```
josueetcom, line 33: I recommend decoupling style between different element types.

src/pages/ScholarshipForm.jsx
```diff
@@ -20,20 +20,15 @@ function ScholarshipForm() {
...
-        ...formFieldStates,
-        [id]: parseInt(e.target.value, 10) || '',
-      });
+
+  function updateAmount(id, value, id2 = '') {
+    if (id2) {
+      setFormFieldStates({ ...formFieldStates, [id]: value, [id2]: value });
+    } else {
+      setFormFieldStates({ ...formFieldStates, [id]: value });
+    }
```
josueetcom, line 29: I was thinking more like 
```js
function updateAmount(minAmount, maxAmount) {
  setFormFieldStates({...formFieldStates, minAmount, maxAmount});
}
```
And not worrying about which text field is being set.

src/pages/ScholarshipForm.jsx
```diff
@@ -20,20 +20,15 @@ function ScholarshipForm() {
     return (e) =>
       setFormFieldStates({ ...formFieldStates, [id]: e.target.value });
```
josueetcom, line 21: btw I think `[id]` here can simply be `id`

src/components/ScholarshipAmountField.jsx
```diff
@@ -23,65 +24,80 @@ function ScholarshipAmountField(props) {
...
+        return amountTextField('Amount', minAmount, (e) =>
+          updateAmount(
+            'minAmount',
+            parseInt(e.target.value, 10) || '',
+            'maxAmount'
+          )
+        );
     }
-    return <div />;
   }
```
josueetcom: Instead of this you could just have them show up next to the relevant radio button options. E.g. something like
```
( ) Unknown  
( ) Fixed Amount: $[Amount]
( ) Range: $[Min Amount] to $ [Max Amount]
( ) Full Tuition
```

And just disable the ones for options that aren't selected based on `amountType`

src/components/ScholarshipAmountField.jsx
```diff
@@ -1,20 +1,21 @@
...
 } from '@material-ui/core';
 import { makeStyles } from '@material-ui/core/styles';
 import PropTypes from 'prop-types';
 import AmountType from '../types/AmountType';
 
-const useStyles = makeStyles({
+const useStyles = makeStyles((theme) => ({
   formControl: {
-    minWidth: 200,
+    margin: theme.spacing(1),
```
josueetcom, line 16: This causes the margin on the left to be off. I recommend setting the `spacing` property in the `Grid container` below instead.

src/components/ScholarshipAmountField.jsx
```diff
@@ -23,65 +24,80 @@ function ScholarshipAmountField(props) {
     minAmount,
     maxAmount,
     onTypeChange,
-    onMinChange,
-    onMaxChange,
-    updateMinMax,
+    updateAmount,
   } = props;
 
+  function amountTextField(label, value, onChange) {
```
josueetcom, line 30: This could be a separate component and you could use `<AmountTextField>` below

src/pages/ScholarshipForm.jsx
edwingl10: I tried it and what ends up happening is that the value doesn't update.

josueetcom: Ah I think I know why, it's because it would set the `id` property instead of whatever the value of `id` is. That's why `[id]` is needed in this case.

src/pages/ScholarshipForm.jsx
josueetcom: Notice I named the parameters `minAmount` and `maxAmount` explicitly for the sake of using the [object property value shorthand](https://javascript.info/object#property-value-shorthand)

edwingl10: Ahh I think I got it. It looks much better for sure.

src/components/ScholarshipAmountField.jsx
```diff
@@ -0,0 +1,117 @@
...
+  } = props;
+
+  function displayAmountFields(option) {
+    switch (option) {
+      case AmountType.Range:
+        return (
+          <>
+            <AmountTextField
+              label="Minimum Amount"
+              value={minAmount || ''}
```
edwingl10, line 40: This is to that the `placeholder` shows, without it the default value `0` would show. 
What I could also do is let the value be a string and then when verifying the input I can convert it to a number.

src/components/ScholarshipAmountField.jsx
```diff
@@ -27,74 +30,85 @@ function ScholarshipAmountField(props) {
...
+          );
+        })}
+      </RadioGroup>
+    </FormControl>
   );
 }
 
 ScholarshipAmountField.propTypes = {
-  amountType: PropTypes.oneOfType(PropTypes.string).isRequired,
+  amountType: PropTypes.string.isRequired,
```
josueetcom, line 111:
```suggestion
  amountType: PropTypes.oneOf(Object.keys(AmountType)).isRequired,
```

src/components/AmountTextField.jsx
```diff
@@ -0,0 +1,37 @@
...
+
+  return (
+    <TextField
+      className={classes.amountField}
+      helperText={label}
+      placeholder="Unknown"
+      type="number"
+      InputProps={{
+        inputProps: { min: 0, step: 50 },
+        startAdornment: <InputAdornment position="start">$</InputAdornment>,
```
josueetcom, line 25: Nice!

src/components/AmountTextField.jsx
```diff
@@ -0,0 +1,37 @@
...
+function AmountTextField(props) {
+  const classes = useStyles();
+  const { label, value, onChange, disabled } = props;
+
+  return (
+    <TextField
+      className={classes.amountField}
+      helperText={label}
+      placeholder="Unknown"
+      type="number"
```
josueetcom, line 22: I imagine if you add `defaultValue=""` here you might be able to get rid of the `|| ''` hack.

edwingl10: I thought so too but it still displays the default value `0`

josueetcom:
I'm encountering #161 when viewing https://s4us-pr-153.onrender.com/scholarships/new. I wonder if the issue is the `date-fns` dependency added in https://github.com/beyondhb1079/s4us/commit/d6adbfd18769cf280ba6793ca20b474d0919e755. You might want to try poking around the DatePicker to see if that's the issue and how to fix it

src/components/AmountTextField.jsx
josueetcom: Ah that's unfortunate

josueetcom:
Per #161, to prevent a bug like that from getting merged I recommend writing a test that checks that the form elements like the DatePicker show up. Check out the [React Testing Library docs](https://testing-library.com/docs/react-testing-library). Write a test that fails without the fix and passes with the fix

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@date-io/date-fns": "2.x",
+    "@date-io/date-fns": "1.x",
```
josueetcom, line 6: Btw, if this wasn't the issue I suggest reverting it back to the newer version.

josueetcom: If it is the issue, then I would file a separate issue to investigate for later

edwingl10: It seems to have fixed it. I made a separate issue for later use.

josueetcom (approved):
Sorry, I thought I had sent out the review on Friday. 

Just some minor nits. Feel free to submit as is and address the comments in a follow up PR or address them now and you can Slack me to approve it again.

src/components/DatePicker.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+      onChange={() => {}}
+    />,
+    { wrapper: MemoryRouter }
+  );
+
+  const deadline = getByLabelText('Deadline *');
+  fireEvent.change(deadline, {
+    target: { value: new Date('December 17, 2005') },
+  });
+  expect(deadline.value).toBe('17/20/0500');
```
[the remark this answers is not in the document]
edwingl10: I just chose a random date. I'll probably change that in the next PR.

src/components/ScholarshipAmountField.jsx
```diff
@@ -0,0 +1,104 @@
...
+            />
+          );
+        })}
+      </RadioGroup>
+    </FormControl>
+  );
+}
+
+ScholarshipAmountField.propTypes = {
+  amountType: PropTypes.oneOf(Object.keys(AmountType)).isRequired, // supports default value of empty string
```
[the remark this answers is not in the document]
edwingl10: Yes, but it throws an error when the type is an empty string in the beginning. I'll find a way to fix that in the next PR.

src/components/DatePicker.test.jsx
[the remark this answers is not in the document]
josueetcom: Oh I meant it doesn't seem to match the date. Where is december in the string?

PR #155, opened 2020-11-02
Fetch data and style scholarships search page

josueetcom:


src/components/ScholarshipList.jsx
```diff
@@ -1,43 +1,113 @@
...
-        <ListItemText primary={item.name} secondary={item.school} />
-        <ListItemText primary={item.amount} />
-        <ListItemText primary={item.deadline} />
-      </ListItem>
-    </Link>
-  ));
-
-  return <List className={classes.root}>{scholarshipList}</List>;
+  const shareFn = (id, data) => () => {
+    if (navigator.share) {
```
edwingl10, line 42: where does `navigator` come from? I don't see it imported?

josueetcom: [navigator](https://developer.mozilla.org/en-US/docs/Web/API/Window/navigator) is something present in the Web APIs that represents the state and identity of the user agent.

edwingl10: ooo I see!

edwingl10 (approved):

PR #163, opened 2020-11-08
Home learn more content

gonzalo1330:


gonzalo1330:
One question I have is: is it possible to align text within the grid to be centered but from top-down not centered in the sides? If there is can you point me towards the right direction or an API that shows those attributes I looked on the material UI but couldn't find it but I might have missed it because of the wording possible. Aside from that then the render looks similar to the mock design. This portion will go below PR #89

src/components/LearnMore.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+export default function LearnMore() {
+  return (
+    <Container style={{ textAlign: 'left', paddingTop: '50px' }}>
+      <Grid container spacing={2}>
+        <Grid item>
+          <img width="500" src={logopic} className="profilePic" alt="" />
+        </Grid>
+        <Grid item xs={12} sm={6} container>
+          <Grid item xs container direction="column" spacing={2}>
```
edwingl10, line 17: For vertical centering try `style={{ alignItems:'center' }}`
I believe it should work for the first container but if not maybe try it on the the one below.

src/App.css
```diff
@@ -1,5 +1,6 @@
 .App {
   text-align: center;
```
josueetcom, line 2: Please remove this so that you can left align below

src/App.css
```diff
@@ -1,5 +1,6 @@
...
   width: 100vw;
-  min-height: 100vh;
+  min-height: 90vh;
+  background-color: white;
```
josueetcom, line 5: Please revert these changes. The last line conflicts with the theme's background color

josueetcom (commented):
> One question I have is: is it possible to align text within the grid to be centered but from top-down not centered in the sides? If there is can you point me towards the right direction or an API that shows those attributes I looked on the material UI but couldn't find it but I might have missed it because of the wording possible. Aside from that then the render looks similar to the mock design. This portion will go below PR #89

I'm not sure exactly what you mean but you can always set `textAlign` on parent elements and that'll affect their children. Additionaly you can try exploring the [interactive Grid example](https://material-ui.com/components/grid/#interactive) to see if there's a combination of `direction`, `justify` and `alignItems` you can use.

src/components/LearnMore.jsx
```diff
@@ -0,0 +1,38 @@
...
+import Grid from '@material-ui/core/Grid';
+import Typography from '@material-ui/core/Typography';
+import { Container } from '@material-ui/core';
+import { Link } from 'react-router-dom';
+import Button from '@material-ui/core/Button';
+import logopic from '../img/blank.png';
+
+export default function LearnMore() {
+  return (
+    <Container style={{ textAlign: 'left', paddingTop: '50px' }}>
```
josueetcom, line 11: `textAlign` shouldn't need to be set. Try inspecting this element in the browser and figuring what's changing the text alignment and fix that instead.

`paddingTop` is not necessary here either

josueetcom:
> One question I have is: is it possible to align text within the grid to be centered but from top-down not centered in the sides? If there is can you point me towards the right direction or an API that shows those attributes I looked on the material UI but couldn't find it but I might have missed it because of the wording possible. Aside from that then the render looks similar to the mock design. This portion will go below PR #89

Ah, I think I know what you mean now. This is the reason you added the `paddingTop` field huh? Well, based on the [interactive Grid example](https://material-ui.com/components/grid/#interactive) it looks like what you want is to set `justify` to `flex-start` and `alignItems` to `center`. But that assumes that you add each element as a Grid item in a container. So your code would look something like

```jsx
<Grid container>
  <Grid item container xs={12} sm={6}>  // You'll need to style justify and alignItems
    <Grid item>
       <Typography>...</Typography>
    </Grid>
    // repeat for each element
  </Grid>
  <Grid item xs={12} sm={6}>
  </Grid>
</Grid>
```

src/components/LoginDialog.jsx
```diff
@@ -10,14 +10,18 @@ import PropTypes from 'prop-types';
 
 const uiConfig = {
   signInFlow: 'popup',
+  signInSuccessWithAuthResult() {
+    return true;
+  },
+  signInSuccessUrl: '/get-started',
```
josueetcom, line 16: This file's changes do not seem related to this PR

src/components/LearnMore.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import Grid from '@material-ui/core/Grid';
+import Typography from '@material-ui/core/Typography';
+import { Container } from '@material-ui/core';
+import { Link } from 'react-router-dom';
+import Button from '@material-ui/core/Button';
+import logopic from '../img/blank.png';
+
+export default function LearnMore() {
```
josueetcom, line 9: The structure of this component is very similar to the other ones that are going to be on the home page. Something to think about for a later PR is how you might combine the two into a more generic `HomeSection` component that take in things like the title, image, body, and direction.

src/App.css
gonzalo1330: Ok removed them from there

src/App.css
gonzalo1330: ok removed so the only thing on the CSS file should be : 
```css
.App {
  width: 100vw;
}
```

src/components/LearnMore.jsx
gonzalo1330: Ok got it figured out how to resolve this
& also paddingTop is removed but there might need to be space above once the component `ScholarshipMadeSimple` is above

src/components/LoginDialog.jsx
gonzalo1330: what should I do? remove them or keep them

src/components/LearnMore.jsx
gonzalo1330: Right that does work for horizontal but I was messing around with the web inspector and discovered that align-self: 'center' does the centering for vertical position. Thanks for the help btw

gonzalo1330:
I believe the changed should allow this PR to merge with main its looking exactly as the mock design and the img issue was fixed along with the setting of styles inline. If any changes are needed ill get them done once they are requested

src/components/LearnMore.jsx
gonzalo1330: ok ill consider how to get this done

src/components/LoginDialog.jsx
josueetcom: Remove them from this PR as they're not related.

PR #177, opened 2020-11-14
Form validation 

edwingl10:


edwingl10:
Quick note, I ended up changing the State type for the min/max to string. This is so I could check the case in which the user chose to input a range and did the following: 
O $ hello to $10
Previously I couldn't check for this because if the user inputted an invalid field, it stored '' in the state. This caused problems because even though the state value was '', the input field showed whatever string the user typed, therefore if they did the above, the test would pass as a valid input.

src/pages/ScholarshipForm.jsx
```diff
@@ -1,21 +1,112 @@
...
+import AmountType from '../types/AmountType';
+
+const useStyles = makeStyles({
+  textFieldStyle: {
+    minWidth: 400,
+  },
+});
+
+// regular expression used for valid website urls
+const websiteReg = /^(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[a-z0-9]+([-.]{1}[a-z0-9]+)*\.[a-z]{2,5}(:[0-9]{1,5})?(\/.*)?$/;
```
josueetcom, line 15: I recommend trying these out on a regex testing website like https://regexr.com/ and testing to see if there's a simpler form that's still relatively simple to read. E.g. `(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?` can be simplified to `(https?:\/\/(www\.)?)?`. It captures the optional `s` and the optional `www.`. If you want to show the optional `s` a little more clearly you could do `(http(s)?:\/\/(www\.)?)?`.

```suggestion
const websiteReg = /^(http(s)?:\/\/(www\.)?)?[\w-]+(\.[\w-]+)*\.[a-z]{2,}((\/|#|\?)(\S)*)?$/;
```

You can play around with it [here](https://regexr.com/5g91f)

src/pages/ScholarshipForm.jsx
```diff
@@ -42,26 +133,54 @@ function ScholarshipForm() {
     <Container maxWidth="md">
       <form>
         <h1>Submit a Scholarship</h1>
-        {requiredTextField('name', 'Scholarship Name')}
+        <div>
+          {requiredTextField(
+            'name',
+            'Scholarship Name',
+            formFieldErrors.name,
+            'Please enter a name'
```
josueetcom, line 141: It seems like the `requiredTextField` component could do basic checking that there's a value by default since it's required.

src/pages/ScholarshipForm.jsx
```diff
@@ -42,26 +133,54 @@ function ScholarshipForm() {
...
-        {requiredTextField('description', 'Description')}
-        {requiredTextField('website', 'Website')}
+        <div>
+          {requiredTextField(
+            'description',
+            'Description',
+            formFieldErrors.description,
+            'Please enter a description',
+            true
+          )}
```
josueetcom, line 161: With so many parameters it gets hard to read this and understand values like `true` by themselves without referencing the function component. I suggest capitalizing the component name (and maybe putting in its own file) and accepting properties so you can use it like this:
```jsx
<RequiredTextField id="description" label="Description" ... multiline/>
```

Also, it seems like `id` or `label` could be calculated from each other (at least in these cases). That could potentially be an optimization to make but I can also see how it could make things difficult if that's ever not the case.

src/pages/ScholarshipForm.jsx
```diff
@@ -42,26 +133,54 @@ function ScholarshipForm() {
...
+          helperText={amountHelperText}
           amountType={formFieldStates.amountType}
           minAmount={formFieldStates.minAmount}
           maxAmount={formFieldStates.maxAmount}
           onTypeChange={updateFn('amountType')}
           updateAmount={updateAmount}
         />
         <div>
-          <Button variant="contained" color="primary">
+          <Button variant="contained" color="primary" onClick={validateFields}>
```
josueetcom, line 183: Have you tried setting the form's `onSubmit` handler and making this an `<input type="submit" value="Submit" component={Button} color="primary"/>` element?  I'm looking at the [React forms docs](https://reactjs.org/docs/forms.html)

josueetcom: I'm curious if that would help highlight the required fields when the user tries to submit instead of having to check every single required field.

josueetcom: Since the `onSubmit` handler would receive all form values, I also wonder if there's a way to avoid passing in `onChange` handlers and instead just pass in validation functions and have the component itself worry about when to surface an error.

src/pages/ScholarshipForm.jsx
```diff
@@ -42,26 +133,54 @@ function ScholarshipForm() {
...
+            true
+          )}
+        </div>
+        <div>
+          {requiredTextField(
+            'website',
+            'Website',
+            formFieldErrors.website,
+            'Please enter a valid website'
+          )}
```
josueetcom, line 169: Another field you could accept is `type`, e.g. so that [`type="url"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/url) can be passed instead of having our own URL validation. `Input` is one of the components `TextField` [inherits from](https://material-ui.com/api/text-field/#advanced-configuration)

src/pages/ScholarshipForm.jsx
```diff
@@ -25,13 +116,13 @@ function ScholarshipForm() {
...
   }
 
-  function requiredTextField(id, label) {
+  function requiredTextField(id, label, error, helperText, multiline = false) {
     return (
       <TextField
-        id={id}
-        label={label}
+        className={classes.textFieldStyle}
+        {...{ id, label, error, multiline }}
```
josueetcom, line 123: I would place these either before or after all the other properties

src/pages/ScholarshipForm.jsx
```diff
@@ -1,21 +1,112 @@
...
+    deadline: false,
+    description: false,
+    amountType: false,
+    website: false,
+    minAmount: false,
+    maxAmount: false,
+  });
+
+  // returns true if amount is invalid
+  function invalidAmountFields() {
```
josueetcom, line 42: I feel like this validation doesn't belong here as it's specific to the `ScholarshipAmountField` component. You'd have to copy this logic over to another form if the `ScholarshipAmountField` component ever gets used elsewhere.

src/components/ScholarshipAmountField.jsx
```diff
@@ -90,15 +92,29 @@ function ScholarshipAmountField(props) {
...
   );
 }
 
 ScholarshipAmountField.propTypes = {
-  amountType: PropTypes.oneOf(Object.keys(AmountType)).isRequired, // supports default value of empty string
-  minAmount: PropTypes.number.isRequired,
-  maxAmount: PropTypes.number.isRequired,
+  amountType: PropTypes.oneOf(Object.values(AmountType)),
+  minAmount: PropTypes.string.isRequired,
+  maxAmount: PropTypes.string.isRequired,
```
josueetcom, line 105: You could also leave these as numbers and just assume that `0` is not a valid min or max amount value (and hence show an empty text field).

josueetcom (changes requested):
Initial pass. I think there are ways to simplify this and take advantage of existing components

josueetcom:
Btw, this looks more like a PR for form validation. I couldn't find any submission business logic (which should probably be in a separate PR from form validation anyways)

src/pages/ScholarshipForm.jsx
edwingl10: Much simpler!

src/pages/ScholarshipForm.jsx
edwingl10: I like this idea, I'll try to implement it.

src/components/ScholarshipAmountField.jsx
edwingl10: To show an empty text field I would have to do the `amount || ''` hack like before.

josueetcom: Yeah that's probably right

src/pages/ScholarshipForm.jsx
edwingl10: Alright I put it into another file and import the function in the `ScholarshipForm` page. That grants us re-usability.

src/pages/ScholarshipForm.jsx
edwingl10: ok so I tried this and it does highlight the required fields, and also checks depending on the type. Should I do it this way or keep and manage the states. One thing I'm not sure about yet is how to make it work for the amount fields, specifically the range

src/components/ScholarshipAmountField.jsx
```diff
@@ -5,13 +5,17 @@ import {
...
+  FormHelperText,
 } from '@material-ui/core';
 import { makeStyles } from '@material-ui/core/styles';
 import PropTypes from 'prop-types';
 import AmountType from '../types/AmountType';
 import AmountTextField from './AmountTextField';
 
 const useStyles = makeStyles({
+  formControlStyle: {
+    paddingTop: 20,
```
josueetcom, line 17: Use the theme's spacing:
```suggestion
const useStyles = makeStyles((theme) => {
  formControlStyle: {
    paddingTop: theme.spacing(2),
```
I believe the default spacing unit is 8px so this'll give a padding of 16px. Either 1 or 2 units should be fine.

src/components/ScholarshipAmountField.jsx
```diff
@@ -21,6 +25,10 @@ const useStyles = makeStyles({
 function ScholarshipAmountField(props) {
   const classes = useStyles();
   const {
+    typeError,
+    minAmountError,
+    maxAmountError,
```
josueetcom, line 30: Simplify by simply passing `error` and calculate individual fields' `error` (and `helperText` too if you like) based on `amountType`
```suggestion
    error
```
Right beneath this statement you can do:
```jsx
const isRange = amountType === AmountType.Range;
const isFixed = amountType === AmountType.Fixed;
```

And then in your `switch` statement below render text fields like this:
```jsx
<AmountTextField error={error && isRange}
              value={minAmount || ''}
              onChange={...}
              disabled={!isRange}
              helperText={helperText && isRange}/>
```

src/components/ScholarshipAmountField.jsx
```diff
@@ -35,18 +43,22 @@ function ScholarshipAmountField(props) {
...
             <AmountTextField
-              value={minAmount || ''} // hack so that the placeholer 'Unknown' shows up instead of the default value of 0
-              onChange={(e) =>
-                updateAmount(parseInt(e.target.value, 10) || '', maxAmount)
-              }
+              error={amountType === AmountType.Range && minAmountError}
+              value={minAmount || ''}
+              onChange={(e) => {
+                const val = parseInt(e.target.value, 10);
+                updateAmount(val % 1 === 0 ? val : 0, maxAmount);
```
josueetcom, line 50: `val % 1` always returns `0` so this can be written:
```suggestion
                updateAmount(val, maxAmount);
```
If you're checking for `NaN`:
```suggestion
                updateAmount(val === NaN ? val : 0, maxAmount);
```
Or more succinctly I think you can do:
```suggestion
                updateAmount(val || 0, maxAmount);
```

josueetcom: Basically what you had was fine, just needed to change `''` to `0`

src/components/ScholarshipAmountField.jsx
```diff
@@ -56,13 +68,12 @@ function ScholarshipAmountField(props) {
...
-                updateAmount(
-                  parseInt(e.target.value, 10) || '',
-                  parseInt(e.target.value, 10) || ''
-                )
-              }
+              error={amountType === AmountType.Fixed && minAmountError}
+              value={minAmount || ''}
+              onChange={(e) => {
+                const val = parseInt(e.target.value, 10);
+                updateAmount(val % 1 === 0 ? val : 0, val % 1 === 0 ? val : 0);
```
josueetcom, line 75:
```suggestion
                const val = parseInt(e.target.value, 10) || 0;
                updateAmount(val, val);
```

src/components/ScholarshipAmountField.jsx
```diff
@@ -90,15 +101,29 @@ function ScholarshipAmountField(props) {
           );
         })}
       </RadioGroup>
+      <FormHelperText error>
+        {typeError || minAmountError || maxAmountError ? helperText : ' '}
```
josueetcom, line 105: Using just `error` and `amountType`:
```suggestion
        {error && !amountType && 'Please choose an option above'}
```

src/validation/amountValidation.js
```diff
@@ -0,0 +1,44 @@
...
+// returns true if amount is invalid
+export function invalidAmountFields(amountType, min, max, setHelperText) {
+  // 0 is not a valid input
+  const err1 = min === 0;
+  const err2 = max === 0;
+
+  if (amountType === AmountType.Fixed) {
+    setHelperText(err1 ? 'Please input a valid number' : '');
+    return { minAmountError: err1, maxAmountError: false };
+  }
```
josueetcom, line 12: You can skip this entirely by setting `min=1` in the `TextField` component.

josueetcom (commented):
I imagine work is still being done on it. Namely to use form submission as it'll do many of the validations for you. But I've left some additional suggestions

src/pages/ScholarshipForm.jsx
josueetcom: Adding to this, I think the `<form>` component on submission will actually validate a valid URL is passed (might want to double check). But for that you need to set the submit button's `type` to `"submit"`

src/pages/ScholarshipForm.jsx
josueetcom: Form validation is supposed to help avoid doing common validations you'll find across forms. I suggest using the **Submit** button and utilizing [built-in form validation](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#Using_built-in_form_validation). This can cover these use cases of ours:
- Min/Max amount values
- Required fields set
- TextField `type`'s like `url`. Or alternatively `pattern`

For styling, the elements will match the `:invalid` CSS pseudoclass.

Btw, if you want to prevent the redirect/refresh on form submission you'll need to add `event.preventDefault()` in the `onSubmit` handler

josueetcom: I do think you might need to still keep track of state though.

josueetcom: It's up to you ultimately but I would opt for whatever is easiest.

src/components/ScholarshipAmountField.jsx
edwingl10: I initially had this to make sure the inputted value wasn't a decimal but the new implementation doesn't allow characters such as `.` so the first option works.

src/validation/amountValidation.js
```diff
@@ -0,0 +1,44 @@
+import AmountType from '../types/AmountType';
+/* eslint-disable import/prefer-default-export */
+// returns true if amount is invalid
+export function invalidAmountFields(amountType, min, max, setHelperText) {
```
josueetcom, line 4: Instead of passing `setHelperText` I suggest returning the string itself so you can write tests against it.

src/pages/ScholarshipForm.jsx
```diff
@@ -1,21 +1,64 @@
...
+    minWidth: 400,
+  },
+});
+
+// regular expression used for valid website urls
+const websiteReg = /^(http(s)?:\/\/(www\.)?)?[\w-]+(\.[\w-]+)*\.[a-z]{2,}((\/|#|\?)(\S)*)?$/;
 
 function ScholarshipForm() {
+  const classes = useStyles();
   const [formFieldStates, setFormFieldStates] = useState({
```
josueetcom, line 19: (inspired by reading Formik docs) you could shorten these names too:
```suggestion
  const [values, setValues] = useState({
```

And could also, if you like, replace lines like these:
```jsx
onChange={updateFn('name')}
```
with
```jsx
onChange={(e) => setValues({values..., name: e.target.value})}
```
If that makes things easier to read

src/pages/ScholarshipForm.jsx
```diff
@@ -1,21 +1,64 @@
...
 
+  const [formFieldErrors, setFormFieldErrors] = useState({
+    name: false,
+    deadline: false,
+    description: false,
+    website: false,
+    amountHelperText: '',
+  });
+
+  function validateFields() {
```
josueetcom, line 38:
```suggestion
  function validateForm() {
```

src/pages/ScholarshipForm.jsx
```diff
@@ -1,21 +1,64 @@
...
+      !formFieldStates.deadline || isNaN(formFieldStates.deadline); // checks if deadline is invalid date
+    const descriptionError = !formFieldStates.description;
+    const websiteError = !websiteReg.test(formFieldStates.website);
+
+    const amountHelper = invalidAmountFields(
+      formFieldStates.amountType,
+      formFieldStates.minAmount,
+      formFieldStates.maxAmount
+    );
+
```
josueetcom, line 51: You could inline most if not all of them in the setter:
```jsx
setErrors({
  name: !values.name,
  deadline: !values.deadline || Number.isNaN(values.deadline),
  description: !values.description,
  website: !websiteReg.test(values.website),
  amount: invalidAmountField(...),
});
```

src/pages/ScholarshipForm.jsx
josueetcom: I'm going to punt this for #209 to take care of. There's other potential solutions that might be better

src/pages/ScholarshipForm.jsx
```diff
@@ -1,21 +1,64 @@
...
+    setFormFieldErrors({
+      ...formFieldErrors,
+      name: nameError,
+      deadline: deadlineError,
+      description: descriptionError,
+      website: websiteError,
+      amountHelperText: amountHelper,
+    });
+  }
```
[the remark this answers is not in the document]
josueetcom: Feel free to close this and leave #209 to figure out something cleaner

src/pages/ScholarshipForm.jsx
```diff
@@ -1,21 +1,64 @@
...
+  const classes = useStyles();
   const [formFieldStates, setFormFieldStates] = useState({
     name: '',
-    deadline: new Date(),
+    deadline: null,
     description: '',
-    amountType: '',
+    amountType: null,
     minAmount: 0,
     maxAmount: 0,
```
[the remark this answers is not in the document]
josueetcom: you could also leave this change for the submission logic PR as it's probably most helpful there when you're creating the Scholarship object.

src/validation/amountValidation.js
```diff
@@ -0,0 +1,25 @@
+import AmountType from '../types/AmountType';
+/* eslint-disable import/prefer-default-export */
+// returns string for helper text if amount is invalid
+export function invalidAmountFields(amountType, min, max) {
```
josueetcom, line 4: I think you can fix the lint error like this:
```suggestion
// returns string for helper text if amount is invalid
export default function invalidAmountFields(amountType, min, max) {
```

src/pages/ScholarshipForm.jsx
```diff
@@ -25,43 +68,70 @@ function ScholarshipForm() {
...
+            label="Description"
+            error={formFieldErrors.description}
+            helperText={
+              formFieldErrors.description ? 'Please enter a description' : ' '
+            }
+            className={classes.textFieldStyle}
+            required
+            value={formFieldStates.description}
+            onChange={updateFn('description')}
+          />
```
josueetcom, line 109: Make this [Multiline](https://material-ui.com/components/text-fields/#multiline)

src/components/DatePicker.jsx
```diff
@@ -7,24 +7,40 @@ import {
...
   return (
     <MuiPickersUtilsProvider utils={DateFnsUtils}>
       <KeyboardDatePicker
+        className={classes.datePickerStyle}
         format="MM/dd/yyyy"
-        fullWidth
         required
-        {...{ id, label, value, onChange }}
+        helperText={error ? helperText : ' '}
+        {...{ id, label, error, value, onChange }}
```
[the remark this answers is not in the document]
edwingl10: Ok I did the suggested commit but I had to change how I passed in the `helperText` so that it wouldn't appear by default.

src/components/ScholarshipAmountField.jsx
```diff
@@ -90,15 +104,21 @@ function ScholarshipAmountField(props) {
           );
         })}
       </RadioGroup>
+      <FormHelperText error>{error ? helperText : ' '}</FormHelperText>
```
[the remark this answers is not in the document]
edwingl10: I ended up doing `helperText || '  ' ` so that the helperText wouldn't shift the submit button down when it appeared.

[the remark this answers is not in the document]
josueetcom: Works for me!

src/pages/ScholarshipForm.jsx
```diff
@@ -1,21 +1,64 @@
...
-    amountType: '',
+    amountType: null,
     minAmount: 0,
     maxAmount: 0,
     website: '',
     scholarshipType: '',
-    active: false,
   });
 
+  const [formFieldErrors, setFormFieldErrors] = useState({
```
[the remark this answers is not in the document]
edwingl10: ahh yes you're right!

src/pages/ScholarshipForm.jsx
```diff
@@ -1,21 +1,64 @@
...
+    description: false,
+    website: false,
+    amountHelperText: '',
+  });
+
+  function validateFields() {
+    const nameError = !formFieldStates.name;
+    /* eslint-disable no-restricted-globals */
+    const deadlineError =
+      !formFieldStates.deadline || isNaN(formFieldStates.deadline); // checks if deadline is invalid date
```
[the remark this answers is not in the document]
edwingl10: When I do this it allows the user to input an invalid date. 
For example, `13/21/9001` doesn't throw an error like before.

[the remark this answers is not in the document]
josueetcom: Interesting! Hmm, we probably need a different way to validate it's a date then (this is again something where built-in HTML validation could've helped)

src/pages/ScholarshipForm.jsx
```diff
@@ -25,43 +68,70 @@ function ScholarshipForm() {
...
+            required
+            value={formFieldStates.name}
+            onChange={updateFn('name')}
+          />
+        </div>
         <DatePicker
           id="deadline"
           label="Deadline"
+          error={formFieldErrors.deadline}
+          helperText="Please enter a valid date"
```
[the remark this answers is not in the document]
edwingl10: It's because before I had `helperText={error ?  HelperText : ' ']`
So since the error is false before clicking the submit button, the `helperText` doesn't show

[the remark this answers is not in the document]
josueetcom: Ahh alright yeah that's a little confusing considering what `helperText` means in Material UI Form components

src/pages/ScholarshipForm.jsx
josueetcom: #209 can explore this

src/pages/ScholarshipForm.jsx
```diff
@@ -6,8 +6,8 @@ import DatePicker from '../components/DatePicker';
 import { invalidAmountFields } from '../validation/amountValidation';
 
 const useStyles = makeStyles({
-  textFieldStyle: {
-    minWidth: 400,
+  containerStyle: {
```
josueetcom, line 9: *fieldStyle?

josueetcom (approved):
Looks good to me!

PR #179, opened 2020-11-15
About Us: Josue

### LINKED ISSUE #9: About: Josue Rios
Come up with 2-3 sentences about yourself so we can fill the [About page](https://github.com/beyondhb1079/s4us/blob/main/src/pages/About.jsx#L9)

josueetcom:
Also made some structural improvements to the About page.  

There's still some things left to be done though (e.g. using `<Typography>` elements instead of raw `<h*>` elements and cleaning up the page to match current design).

Fixes #9

### DIFF
--- a/src/components/AboutCard.jsx
+++ b/src/components/AboutCard.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { makeStyles, StylesProvider } from '@material-ui/core/styles';
+import { makeStyles } from '@material-ui/core/styles';
 import Card from '@material-ui/core/Card';
 import CardContent from '@material-ui/core/CardContent';
 import CardMedia from '@material-ui/core/CardMedia';
@@ -7,12 +7,9 @@ import Typography from '@material-ui/core/Typography';
 import PropTypes from 'prop-types';
 
 const useStyles = makeStyles({
-  card: {
-    height: 300,
-    width: '80%',
-  },
+  card: {},
   media: {
-    height: 150,
+    height: 200,
   },
 });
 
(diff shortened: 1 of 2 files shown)


src/components/AboutCard.jsx
```diff
@@ -1,18 +1,15 @@
...
 
 const useStyles = makeStyles({
-  card: {
-    height: 300,
-    width: '80%',
-  },
+  card: {},
   media: {
-    height: 150,
+    height: 200,
```
josueetcom, line 12: The Figma design has a height of 198px. I rounded it up here.

src/components/AboutCard.jsx
```diff
@@ -21,26 +18,24 @@ function AboutCard(props) {
...
+      </CardContent>
+    </Card>
   );
 }
 
 AboutCard.propTypes = {
   img: PropTypes.string.isRequired,
   name: PropTypes.string.isRequired,
-  description: PropTypes.string.isRequired,
+  description: PropTypes.node.isRequired,
```
josueetcom, line 38: This allows more customized descriptions if we like via rendered types (such as my about description that has a link)

src/components/AboutCard.jsx
```diff
@@ -21,26 +18,24 @@ function AboutCard(props) {
   const { img, name, description } = props;
 
   return (
-    <StylesProvider injectFirst>
```
josueetcom, line 24: I removed this as I don't believe it's necessary. It should probably actually be injected last (the default) since other stylesheets could potentially override certain settings

src/pages/About.jsx
```diff
@@ -2,41 +2,63 @@ import React from 'react';
 import { Link } from 'react-router-dom';
 import Grid from '@material-ui/core/Grid';
 import Button from '@material-ui/core/Button';
+import MuiLink from '@material-ui/core/Link';
 import Container from '@material-ui/core/Container';
 import AboutCard from '../components/AboutCard';
 import testPic from '../logo.svg';
 import { BRAND_NAME } from '../config/constants';
 
-/* eslint-disable max-len */
```
josueetcom, line 10: Surrounding plain text with React fragments we no longer have to worry about line length as Prettier will auto-format whatever we write.

edwingl10 (approved):

PR #180, opened 2020-11-16
ScholarshipAmount type

### LINKED ISSUE #174: ScholarshipAmount type
See [this comment](https://github.com/beyondhb1079/s4us/issues/56#issuecomment-707469752) in #56 for details.

For rendering the various possible states into human readable text we'll probably want a `toString()` function that depending on the amount type (via a switch) and values (for ranges) returns different strings.

Please add `toString()` tests for all possible cases:
- Unknown
- Fixed Amount
- Full Tuition
- Range with lower bound only
- Range with upper bound only
- Range with both lower and upper bounds

josueetcom:
Fixes #174

### DIFF
--- a/src/types/ScholarshipAmount.ts
+++ b/src/types/ScholarshipAmount.ts
@@ -1,0 +1,51 @@
+import AmountType from './AmountType';
+
+export default class ScholarshipAmount {
+  readonly type: AmountType;
+  readonly min: number;
+  readonly max: number;
+
+  constructor(data?: { type?: AmountType; min?: number; max?: number }) {
+    this.type = data?.type ?? AmountType.Unknown;
+    this.min = data?.min ?? 0;
+    this.max = data?.max ?? 0;
+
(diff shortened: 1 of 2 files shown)


src/types/ScholarshipAmount.ts
```diff
@@ -0,0 +1,51 @@
+import AmountType from './AmountType';
+
+export default class ScholarshipAmount {
+  readonly type: AmountType;
+  readonly min: number;
+  readonly max: number;
+
+  constructor(data?: { type?: AmountType; min?: number; max?: number }) {
+    this.type = data?.type ?? AmountType.Unknown;
```
edwingl10, line 9: Woah I didn't know about the `??`, pretty cool!

josueetcom: Yeah it's nice!

edwingl10 (approved):

Issue #209, opened 2020-11-20
Explore cleaner form validation solutions


josueetcom:
Forms are messy but they're necessary and because humans make mistakes so are validations necessary! Our current scholarship form implementation is a bit messy though 😞 . I suggest whoever takes this up to _explore_ alternatives:

- HTML has [built-in validation](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#Using_built-in_form_validation) we're not using which could help us on some of the basic stuff (e.g. required/min/max) but it might be hard to get it work with React, especially to get the styling right when there's an error.
- [Formik](https://formik.org/) is another one to look into. It seems simple enough but I wonder if it'll really make our code easier to read/maintain 🤔 
- Others

I suggest exploring [this article](https://blog.logrocket.com/the-ultimate-roundup-of-react-form-validation-solutions/) and identifying ones that might best fit our needs:
- Works with Material UI (the easier the better)
- Reads easily
- Is easier to maintain (not necessarily less _lines_ but less complexity!)
- Works well with our current scholarship form

PR #236, opened 2020-11-26
Scholarships form validation using Formik

edwingl10:
## Description
Uses [Formik](https://formik.org/docs/overview) and [Yup](https://github.com/jquense/yup) for a cleaner approach to the form validation. 

## Motivation and Context
Attempts to fix [#209](https://github.com/beyondhb1079/s4us/issues/209). See that issue to get more context.

## How Has This Been Tested?

I've checked the solution on Firefox on my laptop.

## Types of changes

## Previewing Changes

2. Click on Add found in the header
3. Fill the form or leave empty and click on the submit button
4. Notice the validation errors
5. When all fields are valid, you should see a pop out with the values inputted

## Checklist:


src/components/ScholarshipAmountField.jsx
```diff
@@ -14,7 +14,7 @@ import AmountTextField from './AmountTextField';
 
 const useStyles = makeStyles((theme) => ({
   formControlStyle: {
-    paddingTop: theme.spacing(2),
+    paddingBottom: theme.spacing(1),
   },
```
josueetcom, line 18: Rather than hardcode style the AmountTextField I suggest receiving the classname with the styles in the parent. Note that this is for the overall formcontrol and not the inner amount textfields

src/validation/ValidationSchema.js
```diff
@@ -0,0 +1,55 @@
+import * as yup from 'yup';
+import AmountType from '../types/AmountType';
+
+/* eslint-disable import/prefer-default-export */
+export const validationSchema = yup.object({
```
josueetcom, line 5: To get rid of the lint error I think you can do
```jsx
export default const validationSchema = yup.object({
```

Or do
```jsx
const validationSchema = ...;

export default validationSchema
```

src/pages/ScholarshipForm.jsx
```diff
@@ -1,115 +1,134 @@
...
-const useStyles = makeStyles({
+const useStyles = makeStyles((theme) => ({
   containerStyle: {
     maxWidth: 400,
   },
-});
-
-// regular expression used for valid website urls
-const websiteReg = /^(http(s)?:\/\/(www\.)?)?[\w-]+(\.[\w-]+)*\.[a-z]{2,}((\/|#|\?)(\S)*)?$/;
+  fieldStye: {
```
josueetcom, line 13: *fieldStyle

src/pages/ScholarshipForm.jsx
```diff
@@ -1,115 +1,134 @@
...
   function updateAmount(minAmount, maxAmount) {
-    setValues({ ...values, minAmount, maxAmount });
+    formik.setFieldValue('amount.minAmount', minAmount, false);
+    formik.setFieldValue('amount.maxAmount', maxAmount, false);
   }
 
   return (
     <Container maxWidth="md">
-      <form className={classes.containerStyle}>
+      <form className={classes.containerStyle} onSubmit={formik.handleSubmit}>
```
josueetcom, line 49: Should this be Formik's [`<Form>`](https://formik.org/docs/api/form)?

src/pages/ScholarshipForm.jsx
```diff
@@ -1,115 +1,134 @@
...
-          helperText={(errors.deadline && 'Please enter a valid date') || ' '}
-          value={values.deadline}
-          onChange={(date) => setValues({ ...values, deadline: date })}
+          label="Deadline *"
+          error={formik.touched.deadline && Boolean(formik.errors.deadline)}
+          helperText={
+            (formik.touched.deadline && formik.errors.deadline) || ' '
+          }
+          value={formik.values.deadline}
+          onChange={(date) => formik.setFieldValue('deadline', date, false)}
```
josueetcom, line 73: The third parameter is to validate on change, why set it to false?

src/pages/ScholarshipForm.jsx
```diff
@@ -1,115 +1,134 @@
...
           <TextField
+            className={classes.fieldStye}
             id="name"
-            label="Scholarship Name"
-            error={errors.name}
-            helperText={errors.name ? 'Please enter a name' : ' '}
-            required
-            value={values.name}
-            onChange={(e) => setValues({ ...values, name: e.target.value })}
+            label="Scholarship Name *"
```
josueetcom, line 55: Is the `*` necessary if you still pass in the `required` field?

src/pages/ScholarshipForm.jsx
```diff
@@ -1,115 +1,134 @@
...
         </div>
+
         <ScholarshipAmountField
-          helperText={errors.amountHelperText}
-          amountType={values.amountType}
-          minAmount={values.minAmount}
-          maxAmount={values.maxAmount}
+          helperText={
+            (getIn(formik.touched, 'amount.type') &&
+              getIn(formik.errors, 'amount.type')) ||
```
josueetcom, line 112: I'm not familiar with formik but maybe `formik.touched.amount.type`, etc. works?

josueetcom: Or if not, I imagine `formik.touched['amount.type']` probably does.

src/validation/ValidationSchema.js
```diff
@@ -0,0 +1,55 @@
...
+    .string()
+    .required('Please enter the scholarship description'),
+  website: yup
+    .string()
+    .url('Website must be a valid URL')
+    .required('Please enter the scholarship website'),
+  amount: yup.object().shape({
+    type: yup
+      .mixed()
+      .oneOf(Object.values(AmountType), 'Please choose an option above'),
```
josueetcom, line 21: Does `.required('Please choose an option above')` work as well?

src/validation/ValidationSchema.js
```diff
@@ -0,0 +1,55 @@
...
+      .number()
+      .when('type', {
+        is: AmountType.Fixed,
+        then: yup
+          .number()
+          .required()
+          .moreThan(0, 'Please enter the scholarship amount'),
+      })
+      /* eslint-disable func-names */
+      .test('test', 'Minimum must be less than the maximum', function (value) {
```
josueetcom, line 32: I think you can fix the lint issue (here and elsewhere) by using arrow functions:
```jsx
      .test('test', 'Minimum must be less than the maximum', (value) => {
```

src/validation/ValidationSchema.js
```diff
@@ -0,0 +1,55 @@
...
+          value > 0 &&
+          maxAmount > 0 &&
+          value >= maxAmount
+        )
+          return false;
+        return true;
+      })
+      .test(
+        'test',
+        'Amount range must have a minimum and/or a maximum',
```
josueetcom, line 45:
```suggestion
        'Range amount must have a minimum and/or a maximum',
```

src/pages/ScholarshipForm.jsx
edwingl10: If I use `<Form>` I would have to wrap my form in a context. Something like the example [here](https://formik.org/docs/overview#reducing-boilerplate)

src/pages/ScholarshipForm.jsx
edwingl10: I took out the required field because It was triggering the built in validation on submit, instead of using Formik.

josueetcom: (I'm ok either way)

src/pages/ScholarshipForm.jsx
edwingl10: I had thought It would be better to validate when the user tried submitting the form again but since the other fields validate `onChange`, setting the validation to true makes more sense.

src/pages/ScholarshipForm.jsx
josueetcom: Ah alright

src/pages/ScholarshipForm.jsx
edwingl10: The first suggestion throws an error while the 2nd suggestion doesn't access the values.

src/validation/ValidationSchema.js
edwingl10: Yes! I changed it

src/validation/ValidationSchema.js
edwingl10: But then `this` returns undefined for `{type, maxAmount}`

src/pages/ScholarshipForm.jsx
josueetcom: Hmm alright

src/pages/ScholarshipForm.jsx
josueetcom: Ah alright

src/validation/ValidationSchema.js
josueetcom: Ah, maybe this then?
```jsx
      .test('test', 'Minimum must be less than the maximum', function test(value) {
```

Or you can leave it as is too

src/pages/ScholarshipForm.jsx
```diff
@@ -40,8 +40,8 @@ function ScholarshipForm() {
   });
 
   function updateAmount(minAmount, maxAmount) {
-    formik.setFieldValue('amount.minAmount', minAmount, false);
-    formik.setFieldValue('amount.maxAmount', maxAmount, false);
+    formik.setFieldValue('amount.minAmount', minAmount, true);
```
josueetcom, line 43: I think `true` is the default behavior is [`validateOnChange`](https://formik.org/docs/api/formik#validateonchange-boolean) is `true` (which it is by default)

josueetcom (approved):

PR #237, opened 2020-11-26
Pass scholarship state between list and detail pages

### LINKED ISSUE #204: Optimize Firestore calls from scholarship list -> detail page
If the search/list page already has the object, why don't we just use that in the detail page? I don't think it's too important that we listen for changes as we don't expect info to change once it's stable. We can trigger a manual refresh by refreshing the page if necessary.

josueetcom:
## Description

- Pass scholarship data from results page to detail page via [location.state](https://reactrouter.com/web/api/location)
- In detail page, only fetch data if needed
- In detail page, clear location state so that refreshing the page works (we might at some point want to remove this but for now it's unintuitive if the data doesn't update on refresh).

Additionally:
- Persist Firestore emulator data across `yarn start` sessions. This'll prevent us from having to create scholarships each time.
- Fix warnings `yarn test` was complaining about.
- Silence lint warnings.

## Motivation and Context

Fixes #204. This reduces having to make another Firestore call in the detail view when we already have that data in the list view. There shouldn't be a need to re-fetch data we already have.

## How Has This Been Tested?

I added a new test that makes sure the passed in scholarship data is rendered immediately if it's passed in. Existing tests assume no location state has been passed in.

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-237.onrender.com/scholarships.
2. Click on a result.
3. Notice how it doesn't say **Loading...** anymore. This is because the scholarship data was immediately passed in.
4. Refresh the page
5. Notice how it says **Loading...** for a split second, reflecting the refresh request.

## Checklist:

### DIFF
--- a/src/pages/ScholarshipDetails.jsx
+++ b/src/pages/ScholarshipDetails.jsx
@@ -34,10 +34,9 @@ export default function ScholarshipDetails({ history, location, match }) {
     }
   }, [history, location]);
 
+  // Fetch the scholarship if we need to load it
   useEffect(() => {
-    // Fetch the scholarship if it wasn't passed
     if (loading) {
-      // Should we maybe set the scholarhip instead of loading every time?
       Scholarships.id(id).get().then(setScholarship).catch(setError);
     }
   }, [id, loading]);
--- a/src/pages/ScholarshipDetails.jsx
+++ b/src/pages/ScholarshipDetails.jsx
@@ -17,6 +17,7 @@ const useStyles = makeStyles(() => ({
 }));
 
 export default function ScholarshipDetails({ history, location, match }) {
+  console.log('history.location', history.location);
   const classes = useStyles();
   const { id } = match.params;
   const [scholarship, setScholarship] = useState(location.state?.scholarship);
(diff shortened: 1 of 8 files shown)


src/pages/ScholarshipDetails.jsx
```diff
@@ -16,27 +16,30 @@ const useStyles = makeStyles(() => ({
...
+  if (scholarship) {
+    document.title = `${BRAND_NAME} | ${scholarship.data.name}`;
+  }
+
+  // Clear location state in case of refresh/navigation to other pages.
+  useEffect(() => {
+    if (location.state.scholarship) {
+      history.replace({ state: {} });
+    }
+  }, [history, location]);
```
edwingl10, line 35: This runs the first time the ScholarshipsDetails page is loaded, and then each time the `history` and `location` change?

edwingl10 (approved):

src/pages/ScholarshipDetails.jsx
josueetcom: That's right. We don't have a good way to detect a page refresh so the best we can do at the moment is replace the state in history so that the next time this page is loaded it's no longer using it. Unfortunately this also means if you navigate away and come back then it refreshes as well. There might be a better solution using something else besides location state though (I think I saw React Redux state mentioned as a potential solution).

PR #239, opened 2020-11-28
Scholarship submission 

edwingl10:
## Description

Implemented submission logic for new scholarships. Submits when all fields contain a valid input and the User clicks on the submit button. Deactivates submit button when submitting form, clears form if submission successful, shows an alert notifying a User if their submission was successful or not.

## Motivation and Context

Fixes [173](https://github.com/beyondhb1079/s4us/issues/173)

## How Has This Been Tested?

I've tested this solution on my laptop on Firefox. 

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-239.onrender.com/.
2. Click on Add in the header
3. Fill the form with valid data
4. Click the submit button
5. If successful, you should see an alert at the bottom of the screen indicating your submission was successful
6. You can view your new scholarship in the Scholarships page

## Checklist:


josueetcom:
> * [x]  Bug fix (non-breaking change which fixes an issue)
> * [ ]  New feature (non-breaking change which adds functionality)
> * [ ]  Breaking change (fix or feature that would cause existing functionality to change)
> * [ ]  User visible change (users will notice UI or functional changes)

This is not a bug fix, this adds new functionality. Also, this is a user visible change

src/validation/ValidationSchema.js
```diff
@@ -27,27 +27,22 @@ const validationSchema = yup.object({
       })
       /* eslint-disable func-names */
       .test('test', 'Minimum must be less than the maximum', function (value) {
```
josueetcom, line 29: `min` is probably a better parameter name

src/validation/ValidationSchema.js
```diff
@@ -27,27 +27,22 @@ const validationSchema = yup.object({
...
-          const { type, maxAmount } = this.parent;
-          if (type === AmountType.Range && value === 0 && maxAmount === 0)
+          const { type, max } = this.parent;
+          if (type === AmountType.Range && value === 0 && max === 0)
             return false;
           return true;
         }
       ),
-    maxAmount: yup.number().notRequired(),
+    max: yup.number().notRequired(),
```
josueetcom, line 45: This needs to be at least 0 btw. It doesn't seem to give an error if I put in an negative number

src/validation/ValidationSchema.js
```diff
@@ -27,27 +27,22 @@ const validationSchema = yup.object({
...
         'Range amount must have a minimum and/or a maximum',
         function (value) {
-          const { type, maxAmount } = this.parent;
-          if (type === AmountType.Range && value === 0 && maxAmount === 0)
+          const { type, max } = this.parent;
+          if (type === AmountType.Range && value === 0 && max === 0)
             return false;
           return true;
         }
       ),
```
josueetcom, line 44: Does this work?
```jsx
      .when('type', {
        is: AmountType.Range,
        then: yup
          .test('test', 'Minimum must be less than the maximum',
            (min, {parent}) => min === 0 || parent.max > 0 && min < max
          )
          .test(
            'test',
            'Range amount must have a minimum and/or a maximum',
            (min, {parent}) =>  min > 0 || parent.max > 0
          )
      }),
```

I think it's better to leave the tests only for the Range type. Also, I just learned that the function you pass to [test()](https://github.com/jquense/yup#mixedtestname-string-message-string--function-test-function-schema) actually accepts the **this** context as the second parameter so you can write them more succinctly and get rid of the lint error.

src/validation/ValidationSchema.js
```diff
@@ -16,7 +16,7 @@ const validationSchema = yup.object({
     .required('Please enter the scholarship website'),
   amount: yup.object().shape({
     type: yup.mixed().required('Please choose an option above'),
-    minAmount: yup
+    min: yup
       .number()
       .when('type', {
         is: AmountType.Fixed,
```
josueetcom, line 22: Btw, I think the `.number()` in line 24 might be redundant with the one from line 20

josueetcom (changes requested):
Just a few comments, nothing major, I think we can get this merged soon.

src/pages/ScholarshipForm.jsx
```diff
@@ -25,27 +29,52 @@ function ScholarshipForm() {
...
     },
     validationSchema,
     onSubmit: (values, { setSubmitting, resetForm }) => {
       setSubmitting(true);
-      alert(JSON.stringify(values, null, 2));
-      setSubmitting(false);
-      resetForm();
+      Scholarships.new(values)
+        .save()
+        .then(() => {
```
josueetcom, line 42: The result here btw is a Scholarship object so you could do something with that. E.g. 

```jsx
        .then((scholarship) => {
          const link = /* generate from scholarship.id like the scholarships page does for its results */;
          setSubmissionAlert(<SnackBar ...> /* Somewhere use the link/title/etc. */</SnackBar>);
        });
```

src/validation/ValidationSchema.js
edwingl10: I need it in order to chain the `.required().moreThan()`. Without it, it throws an error :/

src/validation/ValidationSchema.js
edwingl10: woah yeah it worked! I just had to add `|| parent.max === 0` for the first test so that It could handle the min being filled while the max stayed empty.

src/validation/ValidationSchema.js
edwingl10: Did the field allow you to put negative numbers? For me it doesn't let me input the `-` symbol, therefore I can't input a negative number. Regardless I put the test to check for positive numbers.

josueetcom: Yeah the field had allowed me to put a negative number. Looks like it's not the case anymore though

src/validation/ValidationSchema.js
josueetcom: ohhh I see 👍

src/validation/ValidationSchema.js
josueetcom: Ahh it probably should be `(min, {parent}) min === 0 || parent.max === 0 || min < max` then. Either one of them is 0 or min is less than max.

src/pages/ScholarshipForm.jsx
```diff
@@ -19,8 +20,19 @@ const useStyles = makeStyles((theme) => ({
 
 function ScholarshipForm() {
   const classes = useStyles();
-  const [submissionAlert, setSubmissionAlert] = useState(false);
-  const [error, setError] = useState(false);
+  const [showAlert, setShowAlert] = useState(false);
```
josueetcom, line 23: You can derive `showAlert` from just whether `submissionAlert` is set and set `submissionAlert` to null when there's nothing to show.

src/pages/ScholarshipForm.jsx
```diff
@@ -19,8 +20,19 @@ const useStyles = makeStyles((theme) => ({
...
+  function showSubmissionAlert(severity, message) {
+    return (
+      <Alert
+        variant="filled"
+        onClose={() => setShowAlert(false)}
+        severity={severity}>
+        {message}
+      </Alert>
+    );
+  }
```
josueetcom, line 35: A few things:
1. It might be nice to use an `AlertTitle` here ([demo](https://github.com/knative/client/blob/3da89001a2a1d514554a8fa20b35764cc5d2f5c3/pkg/kn/commands/service/configuration_edit_flags.go#L209)) along with a description. The title can be a single word or a short message while the description contain additional details (e.g. the exact error returned)
1. An [action](https://material-ui.com/components/alert/#actions) to view the Scholarship would be nice. You might need to add the `close` icon action as well if `onClose` doesn't play well with `action` 
1. I think the default `variant` is probably fine.
1. I'm not sure how much this function really deduplicates between the two alerts

src/pages/ScholarshipForm.jsx
```diff
@@ -60,21 +82,7 @@ function ScholarshipForm() {
...
-          variant="filled"
-          onClose={() => setSubmissionAlert(false)}
-          severity={error ? 'error' : 'success'}>
-          {error
-            ? 'Something went wrong. Please try again later.'
-            : 'Submission successful!'}
-        </MuiAlert>
-      </Snackbar>
-
+      <Fade in={showAlert}>{submissionAlert || <Alert>NA</Alert>}</Fade>
```
josueetcom, line 85: I'm curious, what does this look like when you just don't render the dummy **Alert**? I'm curious if **Fade** or another transition would gracefully handle the sudden state change from object to null/empty string well.

src/pages/ScholarshipForm.jsx
edwingl10: regarding this, the `scholarship` parameter only returns the object that was submitted.

src/pages/ScholarshipForm.jsx
edwingl10: I can probably just set the state directly without the function. 
Regarding the action to show the scholarship, how would I get the id? The `scholarship` object returned in the `.then()` is the scholarship that was submitted.

src/pages/ScholarshipForm.jsx
josueetcom: Yes, where `scholarship.data` should match the field values and `scholarship.id` should be the ID of the scholarhip.

src/pages/ScholarshipForm.jsx
edwingl10: Without the Fade it shifts down everything below it. What I could do is make room for it with a dummy Alert showing nothing and then replace it when there is an actual Alert. I think that should prevent it shifting the elements.

src/pages/ScholarshipForm.jsx
josueetcom: `scholarship.id` should work as `save()` resolves to [`this`](https://github.com/beyondhb1079/s4us/blob/dd57b640aeb6bbff61f1c9ba196816503e8f9b36/src/models/base/FirestoreModel.ts#L33), which is of type Model<T>.

josueetcom: Alternatively if you do `const scholarship = Scholarship.new(data)`, `scholarship.id` will also hold the ID.

src/pages/ScholarshipForm.jsx
josueetcom: Hmm, what you could also do if you want to use other transitions to avoid shifting of the form is show the alert at the very bottom beneath the Submit button. That way it doesn't affect the elements above it

src/pages/ScholarshipForm.jsx
```diff
@@ -53,23 +41,22 @@ function ScholarshipForm() {
...
-        .catch(() =>
+        .catch((error) =>
           setSubmissionAlert(
-            showSubmissionAlert(
-              'error',
-              'Something went wrong. Please try again later.'
-            )
+            <Alert onClose={() => setSubmissionAlert(null)} severity="error">
+              <AlertTitle>Error</AlertTitle>
+              {error}
```
josueetcom, line 55: Have you tested this? You might need to do `error.toString()`

edwingl10: yes, my bad.

src/pages/ScholarshipForm.jsx
edwingl10: Ooo that makes sense! Ok I'll add the action to view the scholarship. Thanks!

src/pages/ScholarshipForm.jsx
```diff
@@ -25,27 +29,47 @@ function ScholarshipForm() {
...
+          setSubmissionAlert(
+            <Alert onClose={() => setSubmissionAlert(null)} severity="error">
+              <AlertTitle>Error</AlertTitle>
+              {error}
+            </Alert>
+          )
+        )
+        .finally(() => {
+          setSubmitting(false);
+        });
```
josueetcom, line 61:
```suggestion
        .finally(() => setSubmitting(false));
```

josueetcom (approved):
I'm not sure how I feel about the alert showing up below the submit button yet. Maybe it could be a SnackBar at the top that's dismissable and has the VIEW button (but maybe that would still look weird 🤔)? In any case, we don't have a design or feedback from others yet so I'm good merging this as is.

PR #292, opened 2020-12-17
Gate scholarship form page

### LINKED ISSUE #245: Gate scholarship form page behind authenticated users only
**Is your feature request related to a problem? Please describe.**
Currently anyone on the internet can access our scholarship form at https://dreamerscholars.web.app/scholarships/new. 

**Describe the solution you'd like**
We should gate the form page so that only authenticated users can access/see it. We'll want to either redirect users or bring up the login dialog first. Alternatively we can show some sort of message that they need to log in first. We could also make the **Add** link in the header disabled with mouse hover text saying you need to be logged in to add scholarships.

**Additional context**
React Router has an [example auth workflow](https://reactrouter.com/web/example/auth-workflow) on how to redirect going to protected pages.

#178 is a related issue as Firestore security rules can deny write access to unauthenticated users.

edwingl10:
## Description

Gate the scholarship form page from users who are not authenticated. 

## Motivation and Context

Fixes #245 

## How Has This Been Tested?

On laptop using firefox. 

## Types of changes

## Previewing Changes

2. Make sure you are signed out
3. Try to access the scholarship form page by clicking on 'Add' in the header

## Checklist:

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -28,6 +28,10 @@ RouteWithTitle.propTypes = {
   component: PropTypes.elementType.isRequired,
   path: PropTypes.string.isRequired,
   title: PropTypes.string.isRequired,
+  guard: PropTypes.bool,
+};
+RouteWithTitle.defaultProps = {
+  guard: false,
 };
 
 const routes = [
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -15,6 +15,7 @@ import theme from './theme';
 import { BRAND_NAME } from './config/constants';
 import FirebaseProvider from './lib/FirebaseProvider';
 import ProtectedRoute from './components/ProtectedRoute';
+import LoginDialog from './components/LoginDialog';
 
 function RouteWithTitle({ path, component, title, guard }) {
   useEffect(() => {
@@ -82,6 +83,7 @@ function App() {
               />
             ))}
           </Switch>
+          <Route path="/" component={LoginDialog} />
           <Footer />
         </Router>
       </ThemeProvider>
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -3,7 +3,6 @@ import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
 import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { CssBaseline, ThemeProvider } from '@material-ui/core';
-import { LastLocationProvider } from 'react-router-last-location';
 import Footer from './components/Footer';
 import Header from './components/Header';
 import Home from './pages/Home';
@@ -79,19 +78,17 @@ function App() {
       <ThemeProvider theme={theme}>
         <CssBaseline />
         <Router>
-          <LastLocationProvider>
-            <Header />
-            <Switch>
-              {routes.map(({ path, component, title, guard }) => (
-                <RouteWithTitle
-                  key={path}
-                  {...{ path, component, title, guard }}
-                />
-              ))}
-            </Switch>
-            <LoginDialog />
-            <Footer />
-          </LastLocationProvider>
+          <Header />
+          <Switch>
+            {routes.map(({ path, component, title, guard }) => (
+              <RouteWithTitle
+                key={path}
+                {...{ path, component, title, guard }}
+              />
+            ))}
+          </Switch>
+          <LoginDialog />
+          <Footer />
         </Router>
       </ThemeProvider>
     </FirebaseProvider>
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -21,8 +21,11 @@ function RouteWithTitle({ path, component, title, guard }) {
   useEffect(() => {
     document.title = `${BRAND_NAME} | ${title}`;
   }, [title]);
-  if (guard) return <ProtectedRoute {...{ path, component }} />;
-  return <Route {...{ path, component }} />;
+  return guard ? (
+    <ProtectedRoute {...{ path, component }} />
+  ) : (
+    <Route {...{ path, component }} />
+  );
 }
 RouteWithTitle.propTypes = {
   component: PropTypes.elementType.isRequired,
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -24,7 +24,7 @@ function RouteWithTitle({ path, component, title, guard }) {
   return <Route {...{ path, component }} />;
 }
 RouteWithTitle.propTypes = {
-  component: PropTypes.node.isRequired,
+  component: PropTypes.elementType.isRequired,
   path: PropTypes.string.isRequired,
   title: PropTypes.string.isRequired,
 };
@@ -76,7 +76,10 @@ function App() {
           <Header />
           <Switch>
             {routes.map(({ path, component, title, guard }) => (
-              <RouteWithTitle {...{ path, component, title, guard }} />
+              <RouteWithTitle
+                key={path}
+                {...{ path, component, title, guard }}
+              />
             ))}
           </Switch>
           <Footer />
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -14,11 +14,13 @@ import ScholarshipForm from './pages/ScholarshipForm';
 import theme from './theme';
 import { BRAND_NAME } from './config/constants';
 import FirebaseProvider from './lib/FirebaseProvider';
+import ProtectedRoute from './components/ProtectedRoute';
 
-function RouteWithTitle({ path, component, title }) {
+function RouteWithTitle({ path, component, title, guard }) {
   useEffect(() => {
     document.title = `${BRAND_NAME} | ${title}`;
   }, [title]);
+  if (guard) return <ProtectedRoute {...{ path, component }} />;
   return <Route {...{ path, component }} />;
 }
 RouteWithTitle.propTypes = {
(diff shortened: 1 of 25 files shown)


src/components/LoginButton.jsx
```diff
@@ -16,26 +17,29 @@ export default function LoginButton() {
...
-        Sign out
+  return isSignedIn ? (
+    <ProfileMenu
+      signOut={signUserOut}
+      name={user.displayName}
+      email={user.email}
+      photo={user.photoURL}
+    />
+  ) : (
+    <>
```
josueetcom, line 31: These aren't needed when there's a single component btw

src/components/ProtectedRoute.jsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <Route
+      path
+      render={() => {
+        return isSignedIn ? (
+          <Component />
+        ) : (
+          <Redirect
+            to={{
+              pathname: `/`,
```
josueetcom, line 29: This should be whatever path the user's currently on or `/` if there's no history.

src/components/LoginDialog.jsx
```diff
@@ -30,7 +40,7 @@ export default function LoginDialog() {
 
   return (
     <Dialog
-      open={params.get('login') === 'true' || false}
+      open={(params.get('login') === 'true' && !isSignedIn) || false}
```
josueetcom, line 43:
```suggestion
      open={params.get('login') === 'true' && !isSignedIn}
```

src/components/LoginDialog.jsx
```diff
@@ -11,6 +11,16 @@ export default function LoginDialog() {
...
   const history = useHistory();
   const params = new URLSearchParams(location.search);
+  const [isSignedIn, setIsSignedIn] = useState(!!firebase.auth().currentUser);
+
+  useEffect(() => {
+    const unregisterAuthObserver = firebase
+      .auth()
+      .onAuthStateChanged((user) => {
+        setIsSignedIn(!!user);
+      });
```
josueetcom, line 21:
```suggestion
      .onAuthStateChanged(user => setIsSignedIn(!!user));
```

src/App.jsx
```diff
@@ -70,13 +79,19 @@ function App() {
...
+            <Header />
+            <Switch>
+              {routes.map(({ path, component, title, guard }) => (
+                <RouteWithTitle
+                  key={path}
+                  {...{ path, component, title, guard }}
+                />
+              ))}
+            </Switch>
+            <Route path="/" component={LoginDialog} />
```
josueetcom, line 92: What does this do? We already have a route for this path on line 66.

src/components/LoginButton.jsx
```diff
@@ -16,10 +15,9 @@ export default function LoginButton() {
         setIsSignedIn(!!user);
       });
     return unregisterAuthObserver;
-  }, []); // [] skips cleanup of this effect until the component is unmounted
+  }, []);
 
-  const signUserOut = () => firebase.auth().signOut() && handleClose(); // sign out + closes dialog
-  const showLoginDialog = () => setLoginDialogOpen(true);
+  const signUserOut = () => firebase.auth().signOut(); // sign out + closes dialog
```
josueetcom, line 20:
```suggestion
  const signUserOut = () => firebase.auth().signOut();
```

josueetcom (changes requested):
Some UI behavior I noticed:
- If visit the page and log in, then I have to press back 3 times to get to the previous site. This is probably too much considering that the page itself doesn't change.
- If I'm already logged in and then go to a page with `&login=true` in the URL, then the Login dialog shows up even though I'm logged in.
- If I'm elsewhere besides the home page and then click on Login then it redirects to the home page where the login dialog shows up.

These issues need to be addressed. Also since it's tricky it'd be best to write test cases for these redirects. The React Router library has ways to [start at specific routes](https://reactrouter.com/web/guides/testing/starting-at-specific-routes) and [check location in tests](https://reactrouter.com/web/guides/testing/checking-location-in-tests).

Instead of `login=true` an easier to do this might be to have a `/login` route with the login dialog body that redirects to `/` or whatever path you were trying to access. I think often times I see something like `referrer=` or `nextPage=` query parameters in the URLs.

src/App.jsx
edwingl10: This is so that the login dialog appears, I couldn't get it to show without it making it a route.

src/components/ProtectedRoute.jsx
```diff
@@ -0,0 +1,44 @@
...
+        ) : (
+          <Redirect
+            to={{
+              pathname: lastLocation?.pathname || '/',
+              search: '?login=true',
+              state: { from: path },
+            }}
+          />
+        );
+      }}
```
josueetcom, line 32:
```suggestion
        return isSignedIn === false ? (
          <Redirect
            to={{
              pathname: lastLocation?.pathname || '/',
              search: '?login=true',
              state: { from: path },
            }}
          />
        ) : (
          <Component />
        );
      }}
```

src/App.jsx
```diff
@@ -14,24 +14,32 @@ import ScholarshipForm from './pages/ScholarshipForm';
...
+import ProtectedRoute from './components/ProtectedRoute';
+import LoginDialog from './components/LoginDialog';
 
-function RouteWithTitle({ path, component, title }) {
+function RouteWithTitle({ path, component, title, guard }) {
   useEffect(() => {
     document.title = `${BRAND_NAME} | ${title}`;
   }, [title]);
+  if (guard) return <ProtectedRoute {...{ path, component }} />;
   return <Route {...{ path, component }} />;
```
josueetcom, line 25:
```suggestion
  return guard ? <ProtectedRoute {...{ path, component }} /> : <Route {...{ path, component }} />;
```

src/components/LoginDialog.jsx
```diff
@@ -1,32 +1,36 @@
...
-    signInSuccessWithAuthResult: () => false,
-  },
-};
+export default function LoginDialog() {
+  const location = useLocation();
+  const { showLoginDialog } = location.state || { showLoginDialog: false };
+  const history = useHistory();
+
+  const closeDialog = () =>
+    history.replace({ state: { showLoginDialog: false } });
```
josueetcom, line 16: I suggest ordering/grouping these so that related statements are together, e.g.
```js
  const { showLoginDialog } = location.state || { showLoginDialog: false };
  
  const history = useHistory();
  const closeDialog = () =>
    history.replace({ state: { showLoginDialog: false } });

```

The extra blank line at the bottom is so that `uiConfig` is separated from relatively unrelated lines.

src/components/LoginButton.jsx
josueetcom: Still relevant.

src/components/LoginButton.jsx
```diff
@@ -30,11 +23,15 @@ export default function LoginButton() {
...
+    <Button
+      variant="contained"
+      color="primary"
+      component={Link}
+      to={{
+        state: { showLoginDialog: true },
+      }}
+      replace>
+      Login
+    </Button>
```
josueetcom, line 35: This is not a comment to address in this PR, but it's interesting how simple this `Button` has become. It makes me think we should define this as `LoginButton` instead and have `HeaderNavMenu` handle the question _should I show a dropdown or a Login button?_

josueetcom (changes requested):
Just some minor changes

src/components/LoginDialog.jsx
```diff
@@ -1,32 +1,36 @@
...
-    firebase.auth.FacebookAuthProvider.PROVIDER_ID,
-  ],
-  credentialHelper: 'none', // hacky way to disable redirect on email login
-  callbacks: {
-    signInSuccessWithAuthResult: () => false,
-  },
-};
+export default function LoginDialog() {
+  const location = useLocation();
+  const { showLoginDialog } = location.state || { showLoginDialog: false };
```
josueetcom, line 12: For consistency maybe the default here should be `undefined` as well. It's treated as a falsy value too so I think the login dialog won't show up by default.

josueetcom:
@gonzalo1330 I think it'd be helpful for you to see the changes in this PR and understand them.

Some points of interest:
- Notice how to show/hide the dialog we're now depending on a [`location`](https://reactrouter.com/web/api/location)'s `state.showLoginDialog`. Note how it can be `true`, `false` or `undefined`.
- Notice how [`history.replace`](https://reactrouter.com/web/api/history), [`Link.replace`](https://reactrouter.com/web/api/Link/replace-bool) and [`Redirect.push`](https://reactrouter.com/web/api/Redirect/push-bool) are used to _replace_ the current page (so that a new browser history entry isn't created).
- Notice how `ProtectedRoute` distinguishes between the three User states: _definitely signed in_, _definitely not signed in_, and _unknown_. It shows a progress while the state is unknown instead of treating unknown as false and doing a redirect immediately.

src/components/LoginButton.jsx
edwingl10: Yeah I see how simple it is. We can make another PR where we move the login to the `HeaderNavMenu`

src/components/LoginDialog.jsx
edwingl10: yup works the same.

josueetcom (approved):

PR #306, opened 2021-01-11
track user id and email

### LINKED ISSUE #290: Scholarship form: track author id and email (auth user id)
We should add author id and email to a scholarship. This'll allow:

- Author to edit the scholarship (via ID)
- Author's email to be BCC'd on a scholarship issue report

edwingl10:
## Description

Adds author id and email to scholarship when they submit form. 

## Motivation and Context

Fixes #290

## How Has This Been Tested?

On laptop using firefox and firebase emulator. 

## Types of changes

## Previewing Changes

2. submit a scholarship
3. check scholarship added in firestore

## Checklist:

### DIFF
--- a/src/pages/ScholarshipForm.jsx
+++ b/src/pages/ScholarshipForm.jsx
@@ -1,6 +1,7 @@
 import React, { useState } from 'react';
 import { Link } from 'react-router-dom';
 import { useFormik, getIn } from 'formik';
+import firebase from 'firebase';
 import { Container, Button, TextField, IconButton } from '@material-ui/core';
 import CloseIcon from '@material-ui/icons/Close';
 import { Alert, AlertTitle } from '@material-ui/lab';
@@ -25,6 +26,7 @@ const useStyles = makeStyles((theme) => ({
 function ScholarshipForm() {
   const classes = useStyles();
   const [submissionAlert, setSubmissionAlert] = useState(null);
+  const user = firebase.auth().currentUser;
 
   const formik = useFormik({
     initialValues: {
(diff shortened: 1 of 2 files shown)


src/pages/ScholarshipForm.jsx
```diff
@@ -37,6 +39,8 @@ function ScholarshipForm() {
         max: 0,
       },
       website: '',
+      authId: user?.uid || null,
+      authEmail: user?.email || null,
```
josueetcom, line 43: `auth` could be a little confusing. I'd prefer `author` or `owner`.

josueetcom: Additionally, this violates the [scholarships data schema](https://github.com/beyondhb1079/s4us/blob/main/src/models/Scholarships.ts#L5) expected. I'm sure if you wrote a test that tries to save this, you won't be able to read `authId` and `authEmail` back using the `Scholarships` class.

The Scholarship model needs to be updated to include this. Also, considering that `firebase.auth.currentUser` may be null initially, these values would probably be better set in `onSubmit` instead and ensured that they're not empty.

src/pages/ScholarshipForm.jsx
```diff
@@ -41,7 +43,11 @@ function ScholarshipForm() {
     validationSchema,
     onSubmit: (values, { setSubmitting, resetForm }) => {
       setSubmitting(true);
-      Scholarships.new(values)
+      Scholarships.new({
+        ...values,
+        authorId: user?.uid,
+        authorEmail: user?.email,
```
josueetcom, line 49: I wonder if this logic should be part of `Scholarships.new` or `Collections.new`. We'll probably end up moving it at some point when a similar pattern occurs.

josueetcom (approved):
Looks good enough for now

PR #342, opened 2021-01-19
Sort option functionality

### LINKED ISSUE #276: Sorting options: functionality
- [x] Ensure a sorting option component exists, or create a basic one if needed with all the sorting options
- [ ] Add support to the scholarship list page to apply the sorting option

gonzalo1330:
## Motivation and Context

Fixes #276 

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-342.onrender.com/scholarships.
2. Try sorting
3. See it work

## Checklist:


src/components/SortDropdown.jsx
```diff
@@ -0,0 +1,49 @@
...
+  const classes = useStyles();
+
+  function selectChoice(event) {
+    console.log(`switch Value changed to: ${event.target.value}`);
+    setChoice(event.target.value);
+    updateSorting(event.target.value);
+  }
+
+  return (
+    <>
```
josueetcom, line 29: These aren't needed when there's a single child

src/pages/Scholarships.jsx
```diff
@@ -20,23 +20,26 @@ function ScholarshipsPage() {
...
       .catch(setError)
       .finally(() => setLoading(false));
-  }, []);
+  }, [sortBy, sortFormat]);
 
   const classes = useStyles();
 
   return (
     <Container>
+      {console.log(`Currently sorting by: ${sortBy} + ${sortFormat}`)}
```
josueetcom, line 38: Please delete before merging.

src/pages/Scholarships.jsx
```diff
@@ -20,23 +20,26 @@ function ScholarshipsPage() {
   const [scholarships, setScholarships] = useState([]);
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState();
+  const [sortBy, setSortBy] = useState('deadline');
+  const [sortFormat, setSortFormat] = useState('asc');
 
   useEffect(() => {
     // TODO: Create cancellable promises
-    Scholarships.list({ sortField: 'deadline' })
+    Scholarships.list({ sortField: `${sortBy}`, sortDir: `${sortFormat}` })
```
josueetcom, line 28: `stringVar` is the same as `${stringVar}`.

Also, I suggest renaming the new state variables to `sortField` and `sortDir` so you can simply do this instead:
```jsx
    Scholarships.list({ sortField, sortDir })
```

src/components/SortDropdown.jsx
```diff
@@ -0,0 +1,49 @@
...
+      <FormControl variant="outlined" className={classes.formControl}>
+        <InputLabel>{label}</InputLabel>
+        <Select value={choice} onChange={selectChoice} label={label}>
+          {Object.keys(items).map((key) => (
+            <MenuItem key={key} value={key}>
+              {items[key]}
+            </MenuItem>
+          ))}
+        </Select>
+      </FormControl>
```
josueetcom, line 39: This looks pretty much like `FilterDropdown`. When you see this sort of pattern you need to ask yourself if you can reuse the existing component. If the answer is no (which is why I assume you created this component), then you need to ask yourself if the functionality you need is specific to the sorting dropdown or if they're generally applicable to filter dropdowns in general. In this case, the questions to ask are:
- Is passing a function to invoke on selection change relevant for filter dropdowns in general?
- Is passing an initial selection/choice relevant for filter dropdowns in general?
- Is passing a list of items/labels apart from list a list of values relevant for filter dropdowns in general?

The answer to all those questions is yes. Since the functionality you need is relevant for filter dropdowns in general, it therefore makes sense to reuse `FilterDropdown` and add `onChange`, `initialChoice`, and `values` props to instead. What you'll notice then is that this class basically becomes a thin wrapper around `FilterDropdown` that's probably unnecessary. At least for now. Much later on if we have similar sort dropdowns in other places then it'll make sense to refactor a reusable sort dropdown component.

josueetcom (changes requested):
Initial pass

src/pages/Scholarships.jsx
gonzalo1330: Ok, must of forgot to remove this last console.log

src/components/SortDropdown.jsx
gonzalo1330: Ok must of not removed when I used them for including a `Button` onClick() when testing the change of state variables. I'll remove it

src/components/SortDropdown.jsx
gonzalo1330: Essentially you want the FilterDropdown to contain the logic for sorting and then also you want the FilterDropdown to handle the change in values passed to the Scholarships.list not how it currently is in the FilterBar. BTW I had trouble passing down function that was passed via props, do I have to make a new function within that class that calls the props function and then pass the created function that make the call to the prop function to the FilterDropdown or something?

src/pages/Scholarships.jsx
gonzalo1330: ok

src/components/SortDropdown.jsx
josueetcom: No, it wouldn't have _sorting_ specific logic, it'd have _filter dropdown_ specific logic. That is, logic that one would expect all filter dropdowns can do.

 Basically you'd replace this code with something like this:
```jsx
<FilterDropDown
          label="Sorting"
          items={sortingOptions} // "Deadline (earliest first)", etc.
          values={sortingValues}  // "deadlineSoon", etc.
          defaultValue="deadlineSoon"
          onChange={updateSorting}  // function that updates sortField/sortDir based on the value selected
        />
```
Notice how the _sorting_ specific data/logic is passed in as general filter dropdown properties. You're not coding _sorting_ specific logic in `FilterDropDown`. You're just telling `FilterDropDown` what values to use to style itself and how it should react on a selection change.

josueetcom: To pass a function that was passed via props just pass it. E.g. in `FilterBar` you can do this

```
<SortDropdown setSortDir={setSortDir} setSortField={setSortField}>
```
where `setSortDir` and `setSortField` are properties in both the `FilterBar` and `SortDropdown` components. Then inside `SortDropdown` your `updateSorting` function does the sorting-specific logic you currently have in `FilterBar`.

gonzalo1330: Sounds good changes are pushed

src/components/FilterDropdown.jsx
```diff
@@ -43,13 +44,15 @@ function FilterDropdown(props) {
...
+  defaultValue: PropTypes.string,
   items: PropTypes.objectOf(PropTypes.string).isRequired,
   removeNone: PropTypes.bool,
+  onChange: PropTypes.func,
 };
 
 FilterDropdown.defaultProps = {
-  defaultSelect: '',
+  defaultValue: '',
+  onChange: null,
```
josueetcom, line 55: Rather than defaulting to `null` I think it should be defaulted to a no-op function like `(e) => {}`. Then you don't have to check if it's null.

src/components/FilterDropdown.jsx
```diff
@@ -14,12 +14,13 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 function FilterDropdown(props) {
-  const { label, items, defaultSelect, removeNone } = props;
-  const [choice, setChoice] = useState(`${defaultSelect}`);
+  const { label, items, defaultValue, removeNone, onChange } = props;
+  const [choice, setChoice] = useState(`${defaultValue}`);
```
josueetcom, line 18: LIke I meant to infer in the other comment, a string template with a string value and no other formatting isn't doing. You can simply do this:

```js
  const [choice, setChoice] = useState(defaultValue);
```

josueetcom (dismissed):
Just two minor nits.

src/components/FilterDropdown.jsx
gonzalo1330: ok got it

src/components/FilterDropdown.jsx
gonzalo1330: I think the e in not needed right? I pushed my changes to this and removed the checking of `null`

gonzalo1330:
This PR should be ready to be merged with main

josueetcom (approved):

src/components/FilterDropdown.jsx
josueetcom: I don't think it is. If the linter complains just remove it.

PR #343, opened 2021-01-21
Amount filter button

### LINKED ISSUE #269: Amount filter: styling

edwingl10:
## Description
Amount filter styling to match the Figma design. 

## Motivation and Context

Fixes #269 

## How Has This Been Tested?

On laptop using Firefox. 

## Types of changes

## Previewing Changes

2. Go to the scholarships page
3. Click on the amount filter 

## Checklist:


edwingl10:
Quick Note: In the design I noticed the drop down was below the filter button, I let it cover the filter since the other filters do the same. I wonder if we should let the drop down cover the filter button or if should be displayed below, like in the design.

src/components/AmountFilter.jsx
```diff
@@ -0,0 +1,72 @@
...
+        endIcon={<ArrowDropDownIcon />}>
+        Amount
+      </Button>
+      <Popover
+        id="amount-filter"
+        open={Boolean(anchorEl)}
+        anchorEl={anchorEl}
+        onClose={() => setAnchorEl(null)}
+        anchorOrigin={{ vertical: 'top', horizontal: 'left' }}
+        transformOrigin={{ vertical: -5, horizontal: 'left' }}>
```
edwingl10, line 57: the `vertical: -5` is so that the drop down doesn't fully cover the filter. I wanted it to look and behave somewhat to the other filters we have.

src/components/AmountFilter.jsx
```diff
@@ -0,0 +1,72 @@
...
+import { makeStyles } from '@material-ui/core/styles';
+import ArrowDropDownIcon from '@material-ui/icons/ArrowDropDown';
+import RemoveIcon from '@material-ui/icons/Remove';
+import AmountTextField from './AmountTextField';
+
+const useStyles = makeStyles((theme) => ({
+  buttonStyle: {
+    margin: theme.spacing(1),
+    maxWidth: 120,
+    height: 57,
```
josueetcom, line 12: What is this?

src/components/AmountFilter.jsx
```diff
@@ -0,0 +1,72 @@
...
+        onClick={handleClick}
+        endIcon={<ArrowDropDownIcon />}>
+        Amount
+      </Button>
+      <Popover
+        id="amount-filter"
+        open={Boolean(anchorEl)}
+        anchorEl={anchorEl}
+        onClose={() => setAnchorEl(null)}
+        anchorOrigin={{ vertical: 'top', horizontal: 'left' }}
```
josueetcom, line 56: This seems to align more with the design:
```suggestion
        anchorOrigin={{ vertical: 'bottom', horizontal: 'left' }}
```

It also gives a place to click to close the popover

src/components/AmountFilter.jsx
```diff
@@ -0,0 +1,72 @@
...
+    height: 57,
+  },
+  labelStyle: { paddingLeft: theme.spacing(1) },
+  dashStyle: { margin: '45px auto auto' },
+  filterStyle: { margin: '10px 0' },
+}));
+
+export default function AmountFilter() {
+  const classes = useStyles();
+  const [anchorEl, setAnchorEl] = useState(null);
```
josueetcom, line 21: What is anchorEl and how is it used?

src/components/AmountFilter.jsx
josueetcom: I think with the suggested `anchorOrigin` change `vertical: 'top'` would be more inline with the design here.

src/components/AmountFilter.jsx
```diff
@@ -0,0 +1,72 @@
...
+        <Grid container>
+          {displayFilter('Min Amount', values.min, (e) =>
+            handleChange('min', e.target.value)
+          )}
+          <Grid item className={classes.dashStyle}>
+            <RemoveIcon />
+          </Grid>
+          {displayFilter('Max Amount', values.max, (e) =>
+            handleChange('max', e.target.value)
+          )}
```
josueetcom, line 67: I thought for these you would reuse the [`AmountTextField`](https://github.com/beyondhb1079/s4us/blob/main/src/components/AmountTextField.jsx) component. If it's not good enough yet then I'd adjust that instead so it can be reused. Otherwise if we change how we display/manage amounts in one component we'll have to remember to change it in the `displayFilter` component here too.

src/components/AmountFilter.jsx
```diff
@@ -0,0 +1,72 @@
...
+import { Button, Popover, InputLabel, Grid } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
+import ArrowDropDownIcon from '@material-ui/icons/ArrowDropDown';
+import RemoveIcon from '@material-ui/icons/Remove';
+import AmountTextField from './AmountTextField';
+
+const useStyles = makeStyles((theme) => ({
+  buttonStyle: {
+    margin: theme.spacing(1),
+    maxWidth: 120,
```
josueetcom, line 11: Why is this necessary?

src/components/AmountFilter.jsx
```diff
@@ -0,0 +1,72 @@
...
+    height: 57,
+  },
+  labelStyle: { paddingLeft: theme.spacing(1) },
+  dashStyle: { margin: '45px auto auto' },
+  filterStyle: { margin: '10px 0' },
```
josueetcom, line 16: Margins should be based on `theme.spacing(n)`

josueetcom (changes requested):

src/components/AmountFilter.jsx
edwingl10: It sets the height of the Filter button. It looks weird without it. 
Screen Shot 2021-01-23 at 11 08 45 PM

src/components/AmountFilter.jsx
edwingl10: I was trying to make the width somewhat similar to the other filters but it looks consistent without it so I took it off.

src/components/AmountFilter.jsx
edwingl10: `anchorEl` is the HTML element that the popover uses to set its position. In the `handleClick` function I set the `anchorEl` to the `event.currentTarget` (which is the Amount button). Maybe a better name would be `anchorElement` to make it less confusing?

src/components/AmountFilter.jsx
edwingl10: I did in the `displayFilter` function above. I just extracted it into another function to make it easier to read, since it takes about 8 lines of code to display the `AmountTextField` with its label.

src/components/AmountFilter.jsx
josueetcom: Hmm, 57 is a rather magical number (also from the inspector it looks like the actual size is _56_) but it looks like this is due to the other selects having the [outlined](https://material-ui.com/components/selects/#simple-select) variant as the non-variant ones have a height of 32px.

Instead of introducing a magical number I suggest removing the `variant="outlined"` attribute from the other elements so that there isn't such a height discrepancy. You may need to additionally vertically center the elements on the left as it looks like [buttons](https://material-ui.com/components/buttons/) are usually 36px (you can look at how it's done on the right side of the FilterBar for reference).

src/components/AmountFilter.jsx
josueetcom: I think the current name is fine (el -> element is a logical implication), I was just wondering how it's used.

src/components/AmountFilter.jsx
```diff
@@ -0,0 +1,70 @@
...
+  const handleClick = (event) => setAnchorEl(event.currentTarget);
+  const handleChange = (index, val) =>
+    setValues({ ...values, [index]: parseInt(val, 10) || '' });
+
+  function displayFilter(label, value, onChange) {
+    return (
+      <Grid item className={classes.filterStyle}>
+        <InputLabel className={classes.labelStyle}>{label}</InputLabel>
+        <AmountTextField
+          disabled={false}
```
josueetcom, line 32: Maybe `disabled={false}` should be the default? If it isn't I'd make it the default and change where it _is_ disabled to have the property `disabled`/`disabled={true}`.

src/components/AmountFilter.jsx
```diff
@@ -0,0 +1,70 @@
...
+          variation="outlined"
```
josueetcom, line 33: If the variants/thick heights are removed from the filter bar this can probably be removed

src/components/AmountFilter.jsx
josueetcom: Ah I see. Hmm, I think this may be a case of over-optimization. Helper function components are nice when you can logically separate JSX elements that make up component. In this case though, it's a parent `<Grid item>` element with two child elements. The `<Grid item>` element should probably be wherever its `<Grid item>` siblings are or refactored such that some mapping is done if the children are similar.

josueetcom (changes requested):

src/components/FilterDropdown.jsx
```diff
@@ -23,7 +23,7 @@ function FilterDropdown(props) {
   }
 
   return (
-    <FormControl variant="outlined" className={classes.formControl}>
+    <FormControl variant={variation} className={classes.formControl}>
```
edwingl10, line 26: I added this to not mess with the styling of the sort filter.

josueetcom: I think the sorting filter should actually use the default variant too. Though maybe styled on top of it

src/components/AmountFilter.jsx
```diff
@@ -0,0 +1,68 @@
...
+  const [values, setValues] = useState({});
+
+  const handleClick = (event) => setAnchorEl(event.currentTarget);
+  const handleChange = (index, val) =>
+    setValues({ ...values, [index]: parseInt(val, 10) || '' });
+
+  return (
+    <>
+      <Button
+        variant="contained"
```
josueetcom, line 33: I think an [outlined button](https://material-ui.com/components/buttons/#outlined-buttons) would actually look more like the Figma design

src/components/AmountFilter.jsx
```diff
@@ -0,0 +1,68 @@
...
+        className={classes.buttonStyle}
+        onClick={handleClick}
+        endIcon={<ArrowDropDownIcon />}>
```
josueetcom, line 36: you should add the `color="primary"` property to this icon so it's consistent with the Figma design. The greenish blue found on Figma is eventually going to be our theme's primary color when we change it.

src/components/AmountFilter.jsx
```diff
@@ -54,15 +44,23 @@ export default function AmountFilter() {
...
+            />
+          </Grid>
           <Grid item className={classes.dashStyle}>
             <RemoveIcon />
           </Grid>
-          {displayFilter('Max Amount', values.max, (e) =>
-            handleChange('max', e.target.value)
-          )}
+          <Grid item className={classes.filterStyle}>
+            <InputLabel className={classes.labelStyle}>Min Amount</InputLabel>
```
josueetcom, line 58: *Max Amount

src/components/AmountFilter.jsx
```diff
@@ -7,12 +7,15 @@ import AmountTextField from './AmountTextField';
 
 const useStyles = makeStyles((theme) => ({
   buttonStyle: {
-    margin: theme.spacing(1),
-    height: 57,
+    margin: `${theme.spacing(2)}px ${theme.spacing(1)}px 0`,
```
josueetcom, line 10: This looks pretty hacky to get this to line up with the other dropdowns. This is also very specific to the FilterBar case and may not be applicable on mobile when a filters dialog or something is shown. On desktop if you select a value from the other dropdowns the alignment looks off again. I'd leave it at a margin of `theme.spacing(1)` for now to be less hacky and align a little better. Eventually once the other dropdowns get replaced this'll look better aligned with them if they also have the same margin.

src/components/FilterDropdown.jsx
```diff
@@ -23,12 +25,15 @@ function FilterDropdown(props) {
...
-      <InputLabel>{label}</InputLabel>
-      <Select value={choice} onChange={selectChoice} label={label}>
-        {!removeNone && (
+    <FormControl variant="outlined" className={classes.formControl}>
+      <Select
+        value={choice}
+        onChange={selectChoice}
+        displayEmpty
+        className={classes.selectStyle}>
+        {removeNone || (
```
josueetcom, line 34: Revert this back
```suggestion
        {!removeNone && (
```

josueetcom (approved):

Issue #307, opened 2021-01-12
Major filter: functionality


josueetcom:
- [x] Ensure major filter component exists, or create a basic one if needed
- [x] Add any necessary support to [`Scholarships.list()`](https://github.com/beyondhb1079/s4us/blob/main/src/models/base/FirestoreCollection.ts#L29)
- [x] Add support to the scholarship list page to apply the filter.

Issue #274, opened 2020-12-14
Grade level filter: functionality


josueetcom:
- [ ] Ensure grade level filter component exists, or create a basic one if needed
- [ ] Add any necessary support to [`Scholarships.list()`](https://github.com/beyondhb1079/s4us/blob/main/src/models/base/FirestoreCollection.ts#L29)
- [ ] Add support to the scholarship list page to apply the filter.

Issue #94, opened 2020-10-18
Pagination: functionality


josueetcom:
- [ ] Ensure [pagination](https://material-ui.com/components/pagination/) component exists, or create a basic one if needed
- [ ] Add any necessary support to [`Scholarships.list()`](https://github.com/beyondhb1079/s4us/blob/main/src/models/base/FirestoreCollection.ts#L29)
- [ ] Ensure clicking load more loads more results.

This might be helpful for figuring out the Firestore part of it: https://youtu.be/poqTHxtDXwU

josueetcom:
Relevant: https://youtu.be/poqTHxtDXwU

Jobhdez:
I will start working on the issue as soon as I can.

PR #346, opened 2021-01-26
Make Scholarships.list() customizable

josueetcom:
## Description

Made `Scholarships.list` customizable so that future options can be easily provided and implemented. I also fixed some `console.error` issues that kept showing up when running the tests and set `console.error` up so that it fails a test.

## Motivation and Context

Functionality for scholarship-specific filters will require a scholarship-specific `list()` function as those filters may not be applicable for other collections. This should simplify the work required for #94, #271, and #274, and #307 as all that'll be needed is to [build ](https://firebase.google.com/docs/firestore/query-data/queries) on the Firestore query. 

## How Has This Been Tested?

I've moved the sorting functionality to the Scholarships collection class and added test cases against the expected sorting order for different fields/directions.

## Types of changes

## Checklist:


src/components/FilterDropdown.jsx
```diff
@@ -26,7 +26,7 @@ function FilterDropdown(props) {
     <FormControl variant="outlined" className={classes.formControl}>
       <InputLabel>{label}</InputLabel>
       <Select value={choice} onChange={selectChoice} label={label}>
-        {removeNone || (
```
josueetcom, line 29: @gonzalo1330 I just learned from the test warning statements that using `||` is not really an option as it'll render `true`

src/models/Scholarships.test.ts
```diff
@@ -2,13 +2,48 @@ import { firestore } from 'firebase';
...
+  return Scholarships.new({
+    name: amount.toString(),
+    amount,
+    deadline: new Date(new Date().getTime() + index),
+    website: 'foo.com',
+    description: 'some description',
+  });
+}
+
+// All scholarships, sorted by deadline
```
josueetcom, line 21: The ordering is guaranteed by the last parameter. Each scholarship will be created at the same time or right after the previous scholarship but the index increases so in the end this ends up being the sorted by deadline order from earliest to latest.

src/components/LoginDialog.jsx
```diff
@@ -9,7 +9,7 @@ import StyleFirebaseAuth from 'react-firebaseui/StyledFirebaseAuth';
 
 export default function LoginDialog() {
   const location = useLocation();
-  const { showLoginDialog } = location.state || { showLoginDialog: undefined };
```
josueetcom, line 12: Otherwise there's a warning about passing `undefined` when `open` expects a boolean.

src/components/FilterDropdown.jsx
```diff
@@ -45,10 +45,11 @@ FilterDropdown.propTypes = {
   label: PropTypes.string.isRequired,
   defaultSelect: PropTypes.string,
   items: PropTypes.objectOf(PropTypes.string).isRequired,
-  removeNone: PropTypes.bool.isRequired,
```
josueetcom, line 48: This was marked required but not actually passed in all `FilterDropdown` places. Probably shouldn't be required.

src/components/LoginDialog.jsx
edwingl10: Oh yeah I noticed that.

edwingl10 (approved):

src/components/FilterDropdown.jsx
gonzalo1330: Ok yea I saw the change from main and include the updated change to this with my PR for sorting functionality

src/components/FilterDropdown.jsx
gonzalo1330: Yeah its no longer required with my PR being merged to main it will fix this since it failed a test when I attempted to push today

PR #350, opened 2021-01-27
Basic pagination component

Jobhdez:
## Description
added simple pagination component; I had the understanding that adding functionality for the pagination component was a separate issue so this all I did. I may be wrong though.

### DIFF
--- a/src/pages/About.jsx
+++ b/src/pages/About.jsx
@@ -48,17 +48,17 @@ const team = [
       </>
     ),
   },
-    {
-        name: 'Job Hernandez',
-        img: testPic,
-        description: (
-            <>
-            I am interested and driven to solve problems with software; software
-            has tremendous power to make an impact in the world and transform lives.
-               </>
-        ),
-    },
-        
+  {
+    name: 'Job Hernandez',
+    img: testPic,
+    description: (
+      <>
+        I am interested and driven to solve problems with software; software has
+        tremendous power to make an impact in the world and transform lives.
+      </>
+    ),
+  },
+
   {
     name: 'Sergio Mejia',
     img: testPic,
--- a/src/pages/About.jsx
+++ b/src/pages/About.jsx
@@ -48,6 +48,17 @@ const team = [
       </>
     ),
   },
+    {
+        name: 'Job Hernandez',
+        img: testPic,
+        description: (
+            <>
+            I am interested and driven to solve problems with software; software
+            has tremendous power to make an impact in the world and transform lives.
+               </>
+        ),
+    },
+        
   {
     name: 'Sergio Mejia',
     img: testPic,
--- a/src/pages/Scholarships.jsx
+++ b/src/pages/Scholarships.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, Fragment } from 'react';
 import {
   CircularProgress,
   Container,
@@ -45,7 +45,7 @@ function ScholarshipsPage() {
       <FilterBar changeSortBy={setSortField} changeSortFormat={setSortDir} />
       {error?.toString() ||
         (loading && <CircularProgress className={classes.progress} />) || (
-          <React.Fragment>
+          <Fragment>
             <ScholarshipList scholarships={scholarships} />
             <Button
               className={classes.loadMoreButton}
@@ -55,7 +55,7 @@ function ScholarshipsPage() {
               }}>
               Load More
             </Button>
-          </React.Fragment>
+          </Fragment>
         )}
     </Container>
   );
--- a/src/pages/Scholarships.jsx
+++ b/src/pages/Scholarships.jsx
@@ -6,10 +6,10 @@ import {
   makeStyles,
   Typography,
 } from '@material-ui/core';
+import Pagination from '@material-ui/lab/Pagination';
 import Scholarships from '../models/Scholarships';
 import ScholarshipList from '../components/ScholarshipList';
 import FilterBar from '../components/FilterBar';
-import Pagination from '@material-ui/lab/Pagination';
 
 const useStyles = makeStyles(() => ({
   progress: {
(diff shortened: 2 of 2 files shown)


src/pages/Scholarships.jsx
```diff
@@ -41,6 +43,9 @@ function ScholarshipsPage() {
         (loading && <CircularProgress className={classes.progress} />) || (
           <ScholarshipList scholarships={scholarships} />
         )}
+      <Divider />
+
+      <Pagination count={10} />
```
josueetcom, line 48: The [Figma design](https://www.figma.com/file/nwrW3UebcvSWyQZGb5ixNr/DreamScholar?node-id=0%3A1) has theming on it. You can achieve using the theme's primary color adding `color="primary"`. Also, I think the size should be [Large](https://material-ui.com/components/pagination/#pagination-size).

src/pages/Scholarships.jsx
```diff
@@ -41,6 +43,9 @@ function ScholarshipsPage() {
         (loading && <CircularProgress className={classes.progress} />) || (
           <ScholarshipList scholarships={scholarships} />
         )}
+      <Divider />
```
josueetcom, line 46: Instead of a divider how about adding margins above and below the Pagination component so that it has space above/below it? 

This'll look more like the Figma design.

Relevant info: [styles](https://material-ui.com/styles/basics/). I suggest a bottom/top margin of 24px

josueetcom (dismissed):

src/pages/About.jsx
```diff
@@ -48,6 +48,17 @@ const team = [
...
+    name: 'Job Hernandez',
+    img: testPic,
+    description: (
+      <>
+        I am interested and driven to solve problems with software; software has
+        tremendous power to make an impact in the world and transform lives.
+      </>
+    ),
+  },
+
```
josueetcom, line 61: Gotta update your branch to make these changes disappear. GitHub has a button for this on the main PR tab but otherwise you can do `git pull origin main` to merge changes from the `main` branch on GitHub.

src/pages/Scholarships.jsx
josueetcom: I think the spacing issue here is still relevant. The only thing that's changing is the Pagination component.

src/pages/Scholarships.jsx
josueetcom: Instead of this Pagination component, since we can't know the number of pages beforehand, let's just have a simple **Load More Results** button that loads more results when clicked on it. You don't have to get it working as part of this PR, you can get it to work as part of addressing #94. Eventually we may replace the button with auto-loading when the user scrolls to the bottom but to keep things simple let's just add a simple button.

Jobhdez:
Added a simple load more button.

src/pages/Scholarships.jsx
```diff
@@ -43,6 +45,14 @@ function ScholarshipsPage() {
         (loading && <CircularProgress className={classes.progress} />) || (
           <ScholarshipList scholarships={scholarships} />
         )}
+      <Box mt={24} mb={24} />
```
josueetcom, line 48: This creates an awkward gap between the Load More button and the results of 192px, not 24px. Those [spacing properties](https://material-ui.com/system/spacing/#api) are multipliers on the theme's spacing. In this case, the theme's spacing is 8px as 8*24 = 192. You can double check this with the inspector.

Instead of a box I suggest adding top and bottom margins of 24px (3 * theme's spacing) to the Button.

josueetcom (changes requested):

src/pages/Scholarships.jsx
```diff
@@ -33,6 +35,10 @@ function ScholarshipsPage() {
 
   const classes = useStyles();
 
+  const theme = createMuiTheme({
+    spacing: 8,
+  });
```
josueetcom, line 40: This is not how themes work. This creates a new theme instead of using the existing theme. You can modify line 14 to take in the theme and use the default theme instead:
```js
```

src/pages/Scholarships.jsx
```diff
@@ -45,8 +49,9 @@ function ScholarshipsPage() {
         (loading && <CircularProgress className={classes.progress} />) || (
           <ScholarshipList scholarships={scholarships} />
         )}
-      <Box mt={24} mb={24} />
       <Button
+        mt={theme.spacing(3)}
+        mb={theme.spacing(3)}
```
josueetcom, line 54: This doesn't actually set the margin. You can see what I mean by looking at this element via the inspector:
image

There's no margin. 

If you read [spacing docs](https://material-ui.com/system/spacing/) more carefully, you'll notice how the [input](https://material-ui.com/system/spacing/#transformation) is just a number. E.g. `3`, not `theme.spacing(3)`. `theme.spacing(3)` is used when applying styles using `makeStyles((theme) => {...})` like you technically could on line 14 in this file.

josueetcom (changes requested):

src/pages/Scholarships.jsx
```diff
@@ -5,17 +5,19 @@ import {
...
 import ScholarshipList from '../components/ScholarshipList';
 import FilterBar from '../components/FilterBar';
 
-const useStyles = makeStyles(() => ({
+const useStyles = makeStyles((theme) => ({
   progress: {
     display: 'block',
     margin: 'auto',
   },
+  marginTB: {
```
josueetcom, line 18: Ideally these class names would describe the element they style. Maybe `loadMoreButton` instead?

src/pages/Scholarships.jsx
```diff
@@ -43,6 +47,14 @@ function ScholarshipsPage() {
         (loading && <CircularProgress className={classes.progress} />) || (
           <ScholarshipList scholarships={scholarships} />
         )}
```
josueetcom, line 49: Move this line below the `<Button>` you added. This'll make the button only show up when there are results

src/pages/Scholarships.jsx
```diff
@@ -1,19 +1,23 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, Fragment } from 'react';
```
josueetcom, line 1: This can probably be reverted now that the shorthand is used

josueetcom (approved):
Ship it! 🚀

PR #352, opened 2021-01-28
fixed login button small second glitch

### LINKED ISSUE #42: Smooth login button/avatar transition from Unknown auth
Steps to reproduce:
1. Go to the [website](https://dreamerscholars.web.app/)
2. Log in
3. Refresh the page

Expectation: **Login** button shows Profile Avatar

Actuality: It says **Login** for a split second

I'm not sure if this is a bug in our end, in React's end, or if it's just a browser caching issue.

gonzalo1330:
## Description

Fixed the split second of rendering login button when a user is logged in and reloads page.

## Motivation and Context

Avoid login confusion

Fixes #42 

## How Has This Been Tested?

I have logged in and refreshed and it behaves as expected.

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)
- [ x ] User visible change (users will notice UI or functional changes)

## Previewing Changes

1. Click the `https://s4us-pr-352.onrender.com` URL **render[bot]** left in the comment below.
2. Click `Login In` and sign on
3. Refresh the page and see the circular progress prior to display profile avatar on heading

## Checklist:


src/components/LoginButton.jsx
```diff
@@ -1,37 +1,59 @@
...
+          <Button
+            variant="contained"
+            color="primary"
+            component={Link}
+            to={{
+              state: { showLoginDialog: true },
+            }}
+            replace>
+            Login
+          </Button>
```
gonzalo1330, line 55: When it re directs the first time user logs in there is a movement of the links due to size of avatar should this be concerning? As I fixed the issue with the button/avatar falling under the links

edwingl10: I think that's what the [first](https://github.com/beyondhb1079/s4us/issues/42#issuecomment-766661364) issue is referring to.

src/components/LoginButton.jsx
```diff
@@ -1,37 +1,59 @@
...
-      color="primary"
-      component={Link}
-      to={{
-        state: { showLoginDialog: true },
-      }}
-      replace>
-      Login
-    </Button>
+  return (
+    <>
```
edwingl10, line 35: I think if you put a width here, it will prevent the other links to shift when the button changes to an Avatar. I'm not sure what width that would be set to though.

gonzalo1330: Hm the width would be applied where? I included `width` and `marginLeft` to the avatar element and got it to align as close as possible to the link location when the login button is rendered

edwingl10: in the `<>` although you would have to change it to `<div className={loginSection}>` and in your useStyles make the class and set the width. I checked in the inspector and I think the Login button is 64px in width. So that might work.

src/components/ProfileDropdown.jsx
```diff
@@ -84,7 +81,7 @@ export default function ProfileDropdown(props) {
 
   return (
     <>
-      <Avatar src={photo} onClick={handleClick} />
+      <Avatar src={photo} onClick={handleClick} className={classes.medium} />
```
edwingl10, line 84: The `className={classes.medium}` might not be needed if a common width is set in the `LoginButton`.

src/components/LoginButton.jsx
```diff
@@ -1,37 +1,59 @@
...
+      {error?.toString() ||
+        (loading && <CircularProgress className={classes.progress} />) ||
+        (isSignedIn ? (
```
josueetcom, line 38: Btw, this logic probably makes more sense outside of LoginButton. When I think of what LoginButton is supposed to be, I think it's a button for logging in and maybe logging out. But in reality this component is more than just the button so I suggest moving this logic to HeaderNavMenu and getting rid of the `LoginButton` component as it's no longer just a button.

src/components/LoginButton.jsx
```diff
@@ -1,37 +1,59 @@
...
-      to={{
-        state: { showLoginDialog: true },
-      }}
-      replace>
-      Login
-    </Button>
+  return (
+    <>
+      {error?.toString() ||
+        (loading && <CircularProgress className={classes.progress} />) ||
```
josueetcom, line 37: This isn't a very smooth transition. I think rather than this we can show nothing when the auth state is unknown and then use a transition like [Zoom](https://material-ui.com/components/transitions/#zoom) to display whatever element is appropriate

src/components/LoginButton.jsx
josueetcom: You can fix the shift issue by setting a big enough `minWidth` to the `<Grid item>` holding the button/profile menu in `HeaderNavMenu.jsx` and making the items in the item right-aligned.

src/components/ProfileDropdown.jsx
```diff
@@ -53,12 +53,9 @@ const useStyles = makeStyles((theme) => ({
...
   medium: {
-    width: '48px',
-    height: '48px',
-  },
-  large: {
-    width: '60px',
-    height: '60px',
+    width: '50px',
+    height: '50px',
+    marginLeft: theme.spacing(2),
```
josueetcom, line 58: Is this fixing a spacing issue within the `ProfileDropdown` component? If it isn't then don't add this here. Add the fix in the component where that spacing matters, i.e. in this case `HeaderNavMenu` is where needs to be enough width/spacing between the LoginButton/ProfileDropdown and the other buttons.

src/components/ProfileDropdown.jsx
josueetcom: +1. If `minWidth` is set on the element that contains both the login button and the profile dropdown then the size of the Avatar does not need to change.

Also, this new size is making the header bar slightly thicker when loaded so it introduces a sizing glitch when it goes from loading -> medium avatar shown.

josueetcom (changes requested):

src/components/LoginButton.jsx
gonzalo1330: Wait so the entire LoginButton functionality now be contained within the HeaderNavMenu and no longer have a LoginButton component?

src/components/Header.jsx
```diff
@@ -44,12 +44,12 @@ function Header() {
     <Container>
       {alert}
       <Grid container className={classes.header} spacing={3}>
-        <Grid item>
+        <Grid item sm={8}>
```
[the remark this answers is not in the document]
gonzalo1330: Ok changes reverted

src/components/ProfileDropdown.jsx
gonzalo1330: Ok yea that did work

src/components/ProfileDropdown.jsx
gonzalo1330: This is actually for the dropdown avatar but I removed the marginLeft and the sizing is like that because in the previous PR you mentioned the default size of the avatar within the dropdown was too small. These changes no longer correspond to the avatar within the HeaderNavMenu

src/components/LoginButton.jsx
gonzalo1330: Yeah that does make a lot of sense. I just changed the CircularProgress component with a null component (showing nothing) now I will work on getting a Zoom display for the transition of the avatar

src/components/ProfileDropdown.jsx
gonzalo1330: Its not making a difference since now the Login button and the avatar are always below the links

gonzalo1330: That was what the changes that messed up the iPad view was suppose to fix but since you wanted them undone it no longer fixes that sizing between the links and button/avatar

src/components/LoginButton.jsx
gonzalo1330: I removed it but the login button component is used in another component the ScholarshipsMadeSimple component which I did git rm to the LoginButton and now the code broke and the component is not in my trash bin. Can you just undo the latest commit and go back to the previous, thanks

josueetcom: It comes down to naming and purpose. `LoginButton` as a name suggests it's a button for logging in. But the logic of "should I show a ProfileMenu or a LoginButton?" does not really belong in a component named `LoginButton`. That logic belongs in `HeaderNavMenu` when it's deciding which child element to show. 

Regarding the issue with ScholarshipsMadeSimple, I think you just need to replace `LoginButton` with the same `Button` you now have in `HeaderNavMenu`.

src/components/HeaderNavMenu.jsx
```diff
@@ -30,8 +52,27 @@ function HeaderNavMenu() {
...
+            <Button
+              variant="contained"
+              color="primary"
+              component={Link}
+              to={{
+                state: { showLoginDialog: true },
+              }}
+              replace>
+              Login
+            </Button>
```
josueetcom, line 74: You can copy this to the other place you used `LoginButton`.

src/components/HeaderNavMenu.jsx
```diff
@@ -30,8 +52,27 @@ function HeaderNavMenu() {
...
+              photo={user.photoURL}
+            />
+          ) : (
+            <Button
+              variant="contained"
+              color="primary"
+              component={Link}
+              to={{
+                state: { showLoginDialog: true },
+              }}
```
josueetcom, line 71: This could be written as 
```suggestion
              to={{state: { showLoginDialog: true }}}
```

src/components/HeaderNavMenu.jsx
```diff
@@ -1,16 +1,21 @@
...
+import firebase from 'firebase';
 import { makeStyles, Link as MuiLink, Grid } from '@material-ui/core';
-import LoginButton from './LoginButton';
+import Button from '@material-ui/core/Button';
+import ProfileMenu from './ProfileDropdown';
 
 const useStyles = makeStyles((theme) => ({
   menu: {
-    alignItems: 'center',
+    alignItems: 'right',
```
josueetcom, line 10: This is an invalid value so it makes the menu items not aligned anywhere.

src/components/HeaderNavMenu.jsx
```diff
@@ -1,16 +1,21 @@
...
     justifyContent: 'center',
   },
   menuItem: {
     color: theme.palette.text.secondary,
   },
+  avatarMenu: {
```
josueetcom, line 16: Hmm, better name might be `authItem`. `avatarMenu` makes it sound like you're styling the ProfileMenu itself.

src/components/HeaderNavMenu.jsx
```diff
@@ -21,6 +26,23 @@ function HeaderNavMenu() {
...
+  function updateUserStatus(user) {
+    setLoading(false);
+    setIsSignedIn(!!user);
+  }
+
+  useEffect(
+    () => firebase.auth().onAuthStateChanged((user) => updateUserStatus(user)),
+    []
+  );
+
```
josueetcom, line 42: While you're moving this around, a better idea around this is to initialize `isSignedIn` to `undefined` by default like we do in [ProtectedRoute](https://github.com/beyondhb1079/s4us/blob/e85ef2def10365a1fcd91d7d20e0fc7b5b8779e6/src/components/ProtectedRoute.jsx) and set `loading` to true if `isSignedIn` is `undefined`:

```js
  const [isSignedIn, setIsSignedIn] = useState(!!firebase.auth().currentUser || undefined);
  const loading = isSignedIn == undefined;

  useEffect(
    () => firebase.auth().onAuthStateChanged((user) => setIsSignedIn(!!user)),
    []
  );
```

This makes sure we know when we're in an undefined state vs. the yes/no states afterwards.

src/components/HeaderNavMenu.jsx
```diff
@@ -30,8 +52,27 @@ function HeaderNavMenu() {
           </MuiLink>
         </Grid>
       ))}
-      <Grid item>
-        <LoginButton className={classes.menuItem} />
+      <Grid item className={classes.avatarMenu}>
+        {(loading && <null />) ||
```
josueetcom, line 56: `<null />` probably results in console warning/error messages. You can achieve the same by doing this instead:
```suggestion
        {!loading && 
```

src/components/HeaderNavMenu.jsx
```diff
@@ -30,8 +52,27 @@ function HeaderNavMenu() {
...
+          (isSignedIn ? (
+            <ProfileMenu
+              signOut={signUserOut}
+              name={user.displayName}
+              email={user.email}
+              photo={user.photoURL}
```
josueetcom, line 62: Now that I think about it, I wonder why `ProfileMenu` takes in user date it can read itself from `firebase.auth()` 🤔 . Wanna try doing that?

src/components/HeaderNavMenu.jsx
```diff
@@ -30,8 +52,27 @@ function HeaderNavMenu() {
           </MuiLink>
         </Grid>
       ))}
-      <Grid item>
-        <LoginButton className={classes.menuItem} />
+      <Grid item className={classes.avatarMenu}>
+        {(loading && <null />) ||
+          (isSignedIn ? (
+            <ProfileMenu
```
josueetcom, line 58: One styling I might suggest so that the avatar doesn't appear too far left is to add `margin` of value `'auto'`.

src/components/HeaderNavMenu.jsx
```diff
@@ -30,8 +52,27 @@ function HeaderNavMenu() {
           </MuiLink>
         </Grid>
       ))}
-      <Grid item>
-        <LoginButton className={classes.menuItem} />
+      <Grid item className={classes.avatarMenu}>
+        {(loading && <null />) ||
+          (isSignedIn ? (
```
josueetcom, line 57: This whole `isSignedIn ? A : B` thing can surrounded by a [Zoom transition](https://material-ui.com/components/transitions/#zoom):
```jsx
        (<Zoom in>
          {isSignedIn ? A :B }
        </Zoom>)
```

The code is only ever reached if `!loading` so `in` is always `true`.

src/components/ProfileDropdown.jsx
```diff
@@ -53,12 +53,8 @@ const useStyles = makeStyles((theme) => ({
...
   medium: {
-    width: '48px',
-    height: '48px',
-  },
-  large: {
-    width: '60px',
-    height: '60px',
+    width: '50px',
+    height: '50px',
   },
```
josueetcom, line 58: This CSS class isn't used. Remove?

src/components/ProfileDropdown.jsx
```diff
@@ -53,12 +53,8 @@ const useStyles = makeStyles((theme) => ({
...
   medium: {
-    width: '48px',
-    height: '48px',
-  },
-  large: {
-    width: '60px',
-    height: '60px',
+    width: '50px',
+    height: '50px',
```
josueetcom, line 57: Hmm, how about making these `theme.spacing(6)` instead?

src/components/HeaderNavMenu.jsx
gonzalo1330: Ok it worked but now I noticed that when the user is logged in the change to the avatar does not happen. Im assuming the logic for displaying the profile avatar or login button is used there as well to get it to work

src/components/LoginButton.jsx
gonzalo1330: Ok yeah I see exactly what you're referring to, that makes sense. Yeah I replaced it with the button that is in HeaderNavMenu and that did more. Now again as you mentioned above the logic for if a button is displayed or the avatar will happen in the `ScholarshipsMadeSimple`.

src/components/HeaderNavMenu.jsx
gonzalo1330: Ok I changed it

src/components/HeaderNavMenu.jsx
gonzalo1330: Ok

src/components/HeaderNavMenu.jsx
gonzalo1330: Yeah this make a whole lot of sense to do. The changes are made

src/components/HeaderNavMenu.jsx
gonzalo1330: Yeah sure ill do that

src/components/ProfileDropdown.jsx
gonzalo1330: Sure

src/components/HeaderNavMenu.jsx
gonzalo1330: Ah ok nice

src/components/HeaderNavMenu.jsx
gonzalo1330: Ok yep that worked

src/components/HeaderNavMenu.jsx
gonzalo1330: Ok ill do that

src/components/HeaderNavMenu.jsx
gonzalo1330: Ah ok so should I leave it as `center` or remove this line?

src/components/HeaderNavMenu.jsx
gonzalo1330: ok done

src/components/LoginButton.jsx
gonzalo1330: Alright the logic for showing a ProfileMenu or LoginButton on the `ScholarshipsMadeSimple` component is working accordingly.

src/components/HeaderNavMenu.jsx
gonzalo1330: The logic switch between Avatar and LoginButton on the `ScholarshipsMadeSimple` is implemented and fully functioning.

src/components/HeaderNavMenu.jsx
gonzalo1330: Hm, now the avatar is no longer next to the Links. Should this still be applied since the Avatar is now below the links and is always centered.

gonzalo1330:
Wait actually now that I just merged main into this branch I don't think the switch between Avatar and LoginButton should occur in the ScholarshipsMadeSimple component since the PublicHome is never displayed again after the user logs in. Ill make the changes now so that this is resolved now that I caught it

src/components/HeaderNavMenu.jsx
```diff
@@ -53,26 +47,22 @@ function HeaderNavMenu() {
         </Grid>
       ))}
       <Grid item className={classes.avatarMenu}>
```
josueetcom, line 49: Forgot to update this to use the `authItem` CSS class

josueetcom (changes requested):

src/components/ScholarshipsMadeSimple.jsx
```diff
@@ -50,7 +49,19 @@ const tabs = [
...
       'We intend to provide tools and resources for all students in High School, Community College and University. Our goal is to provide you with access to scholarships to allow you to pursue your career path. We understand that finding scholarships that match your attributes is the most difficult and time consuming task to do as a college student. Our purpose is to connect you with our database to ensure the time you spend finding scholarships is minimize and effectively provide you with scholarships that are right for you.',
-    buttons: [<LoginButton />, OutlineButton('students')],
+    buttons: [
+      <Button
+        variant="contained"
+        color="primary"
+        component={Link}
+        to={{
+          state: { showLoginDialog: true },
+        }}
```
josueetcom, line 59:
```suggestion
        to={{state: { showLoginDialog: true }}}
```

josueetcom (approved):
Looks good!

src/components/HeaderNavMenu.jsx
josueetcom: Still need this

josueetcom (changes requested):

src/components/HeaderNavMenu.jsx
gonzalo1330: ok change is made

josueetcom (approved):

PR #354, opened 2021-01-30
Update pull_request_template.md with hopefully clearer language

josueetcom:
## Motivation and Context

I'm not sure if the template instructions are clear enough so I've tried to make them clearer and slightly simpler.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,19 +1,16 @@
-<!--- Provide a general summary of your changes in the Title above -->
-
-## Description
-
-<!--- Describe or list your changes in detail -->
+<!--- SUMMARIZE your changes in the Title above -->
+<!--- Detail any specific changes here -->
 
 ## Motivation and Context
 
-<!--- Why is this change required? What problem does it solve? -->
-<!--- If it fixes an open issue, please link to the issue here. -->
+<!--- EXPLAIN why this change is required. -->
+<!--- Link any relevant issues via "Fixes #" or "Helps with #". -->
 
 Fixes #
 
 ## How Has This Been Tested?
 
-<!--- Please describe in detail how you tested your changes. -->
+<!--- DESCRIBE in detail how you tested your changes. -->
 <!--- Are these changes covered by new tests or existing tests? -->
 <!--- How else did you test these changes? -->
 
@@ -21,7 +18,7 @@ Fixes #
 
 ## Types of changes
 
-<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->
+<!--- CHECK all the types of changes introduced by replacing "[ ]" with "[x]" in the boxes that apply: -->
 
 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] New feature (non-breaking change which adds functionality)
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -30,18 +27,19 @@ Fixes #
...
+<!--- DELETE THIS SECTION IF THERE ARE NO VISIBLE CHANGES. --->
+
+<!--- DETAIL the steps needed to preview your user visible changes. --->
 <!--- Be very specific about what to look for and what things to try. --->
 
-1. Click the `https://s4us-pr-NUMBER.onrender.com` URL **render[bot]** left in the comment below.
-2.
+<!-- TODO: REPLACE "NUMBER" WITH THE PULL REQUEST NUMBER --->
+<!-- The URL should match the URL left by render[bot]'s comment. --->
+1. Go to https://s4us-pr-NUMBER.onrender.com.
```
edwingl10, line 37: I think this is what is making the test fail.

edwingl10 (changes requested):

.github/pull_request_template.md
josueetcom: Dang yeah you're right. Hmm, I need to make a two line change to get this to work. This PR is low priority so I think I'm going to work on that PR that auto-approves/keeps approval for small changes after an approval.

Jobhdez (approved):
Looks good

Jobhdez (dismissed):
uses clear language -- it is clear as it is but maybe the person who is contributing should add comments to the code? It will make the code easier to understand. maybe the person who is contributing should also make changes to the documentation if necessary. the contributor should not assume the maintainer is going to update the README for instance.

josueetcom:
> uses clear language -- it is clear as it is but maybe the person who is contributing should add comments to the code? It will make the code easier to understand. maybe the person who is contributing should also make changes to the documentation if necessary. the contributor should not assume the maintainer is going to update the README for instance.

I believe when it refers to "documentation" it's likely referring to documentation that library packages provide. As we're not a library package besides README.md I don't know what else might need regular updating. Whether or not to update the README can probably be determined by the reviewer.

Jobhdez:
> > uses clear language -- it is clear as it is but maybe the person who is contributing should add comments to the code? It will make the code easier to understand. maybe the person who is contributing should also make changes to the documentation if necessary. the contributor should not assume the maintainer is going to update the README for instance.
> 
> I believe when it refers to "documentation" it's likely referring to documentation that library packages provide. As we're not a library package besides README.md I don't know what else might need regular updating. Whether or not to update the README can probably be determined by the reviewer.
> 
> Regarding code in comments, I generally expect code to be self-documenting (e.g. `name` is some `string` with the value of a name) as it's hard to keep comments up to date with code. Comments become dated really quickly. Though comments that I think are valuable are those that explain why something strange/complex is done. My general expectation is that a reviewer with knowledge of React/JS that doesn't understand why something is done or thinks something's not clear would ask for a comment be added.

ok sounds good! that makes sense comments are not necessarily needed.

gonzalo1330 (approved):

edwingl10 (approved):

PR #355, opened 2021-01-31
Amount filter functionality & URL support

### LINKED ISSUE #271: Amount filter: functionality
- [ ] Ensure amount filter component exists, or create a basic one if needed
- [ ] Add support to the scholarship list page to apply the filter.

edwingl10:
## Description

Filter scholarships by amount. Can set filter values through URL or amount text field. If the values are invalid, we prune them from the URL, if max <= min, we set error instead of showing results. 

## Motivation and Context

Fixes #271 

## How Has This Been Tested?

On laptop using Firefox. 

## Types of changes

## Previewing Changes

2. Go to the Scholarships page
3. Click on the amount filter button
4. Input amount(s) to filter by 
5. Notice how values are added to URL 

## Checklist:


src/components/AmountFilter.jsx
```diff
@@ -48,8 +66,11 @@ export default function AmountFilter() {
           <Grid item className={classes.filterStyle}>
             <InputLabel className={classes.labelStyle}>Min Amount</InputLabel>
             <AmountTextField
-              value={values.min}
-              onChange={(e) => handleChange('min', e.target.value)}
+              error={error}
+              value={amountFilterVals.min || ''}
+              onChange={(e) =>
+                setMinMax('min', parseInt(e.target.value, 10) || 0)
```
edwingl10, line 72: There is bug in the amount filter text field, when you try changing a number in between 2 other, it instead puts the new number at the end. For example, changing the 3 in `5352` to an 8 results to `5528`. I think this might be causing it?

src/pages/Scholarships.jsx
```diff
@@ -26,14 +29,41 @@ function ScholarshipsPage() {
...
   const [sortField, setSortField] = useState('deadline');
   const [sortDir, setSortDir] = useState('asc');
+  const [amountFilterVals, setAmountFilterVals] = useState({ min: 0, max: 0 });
+  const location = useLocation();
+
+  useEffect(() => {
+    const params = queryString.parse(location.search);
+    const minParam = parseInt(params.min ?? 0, 10);
+    const maxParam = parseInt(params.max ?? 0, 10);
+    setAmountFilterVals({
```
edwingl10, line 39: I tried pruning the invalid URL search  parameters after this by doing
 `let params = new URLSearchParams(location.search);
params.delete('min')` but it didn't work and I noticed the URLSearchParams returned a null object.

edwingl10:
For the filtering logic, if a scholarship value overlaps the range of the filter values, it will include it in the results.

src/filter/FilterByAmount.js
```diff
@@ -0,0 +1,51 @@
...
+        filterMax > 0 &&
+        filterMin <= min &&
+        filterMax >= max) ||
+      // min filter only
+      (filterMax === 0 && filterMin <= min) ||
+      // max filter only
+      (filterMin === 0 && filterMax >= max)
+    );
+
+  // if type is Unknown or Full Tuition
```
josueetcom, line 47: I think Full Tuition should be ok to show as long as there's no MAX cap. Unknown should maybe also show as well since we don't know if it falls in the desired range or not.

src/filter/FilterByAmount.js
```diff
@@ -0,0 +1,51 @@
...
+      // both filter values
+      (filterMin > 0 &&
+        filterMax > 0 &&
+        filterMin <= min &&
+        filterMax >= max) ||
+      // min filter only
+      (filterMax === 0 && filterMin <= min) ||
+      // max filter only
+      (filterMin === 0 && filterMax >= max)
+    );
```
josueetcom, line 45: I'm curious why there's a difference between the two types? You're checking if the scholarship's range (which for fixed ones the min/max is the same) overlaps with the amount filtering range

src/pages/Scholarships.jsx
```diff
@@ -26,14 +29,41 @@ function ScholarshipsPage() {
...
+      min: minParam >= 0 ? minParam : 0,
+      max: maxParam >= 0 ? maxParam : 0,
```
josueetcom, line 41: `minAmount` and `maxAmount` might in case there are other min/max filters in the future.

src/pages/Scholarships.jsx
edwingl10: ooo yes it worked!

src/filter/FilterByAmount.js
edwingl10: I didn't realize the same rules apply for both. I'll combine the `Range` and `Fixed` on line 7.

src/filter/FilterByAmount.js
```diff
@@ -0,0 +1,39 @@
...
+      return (
+        filterMax >= min ||
+        // only min filter
+        filterMax === 0 ||
+        // only max filter
+        (filterMin === 0 && filterMax >= min)
+      );
+
+    // only max filled
+    if (max > 0 && min === 0) return filterMin <= max;
```
josueetcom, line 30: I haven't dug too deeply on the logic here, but I don't understand why there's three cases here when we set `min` and `max` to reasonable defaults [`0` and `RANGE_MAX`](https://github.com/beyondhb1079/s4us/blob/afa48290e9f1d16397c3a9f1c5bd995e763ec8a0/src/types/ScholarshipAmount.ts#L53-L54) when they're not set that should allow for easier comparison.

I think this can simply be written as:
```js
      return (
        (filterMin > 0 &&
          filterMax > 0 &&
          min <= filterMax &&
          max >= filterMin) ||
        // only min fitler
        (filterMin > 0 && filterMax === 0 && max >= filterMin) ||
        // only max filter
        (filterMin === 0 && filterMax > 0 && min <= filterMax)
      );
```

Though that looks a bit messy, so I recommend writing it as this instead:

```js
if (filterMin && max < filterMin || filterMax && min > filterMax) {
  return false;
}
return true;
```

Yes, this could be written as `return !(filterMin && max < filterMin || filterMax && min > filterMax)` but I think this is more readable, at least since this is function instead of a lambda.

Also, considering the min/max [constants we use for FullTuition and Unknown amounts](https://github.com/beyondhb1079/s4us/blob/main/src/types/ScholarshipAmount.ts#L56-L63) I think this code works without needing to check the amount type too. If that's the case then this function is like max 3 lines long. I would inline it in a lambda where you're using it instead.

src/components/AmountFilter.jsx
```diff
@@ -19,14 +22,27 @@ const useStyles = makeStyles((theme) => ({
...
+    amountFilterVals.maxAmount <= amountFilterVals.minAmount;
+
+  const setMinMax = (index, val) => {
+    history.push({
+      search: queryString.stringify({
+        ...queryString.parse(location.search),
+        [index]: val,
+      }),
+    });
+  };
```
josueetcom, line 45: It seems a bit strange for query params to be set here, but read in Scholarships.jsx to then later set the props here. You _could_ technically just read/set the query params here instead of accepting properties. But, since the value is parsed, validated and used in Scholarships.jsx, perhaps it makes more sense to pass to AmountFilter `setMax`/`onMaxChange` and `setMin`/`onMinChange` functions instead? It kinda helps keeps query param specific business in one place since everywhere else just needs the parsed value.

src/components/FilterBar.jsx
```diff
@@ -86,4 +86,5 @@ export default function FilterBar(props) {
 FilterBar.propTypes = {
   changeSortBy: PropTypes.func.isRequired,
   changeSortFormat: PropTypes.func.isRequired,
+  amountFilterVals: PropTypes.objectOf(PropTypes.number).isRequired,
```
josueetcom, line 89: Might be good to just receive the entire query params here since all their values will likely be need to initialize various filters.

src/pages/Scholarships.jsx
```diff
@@ -26,14 +29,55 @@ function ScholarshipsPage() {
   const [error, setError] = useState();
   const [sortField, setSortField] = useState('deadline');
   const [sortDir, setSortDir] = useState('asc');
+  const [amountFilterVals, setAmountFilterVals] = useState({
+    minAmount: 0,
+    maxAmount: 0,
+  });
```
josueetcom, line 35: These don't need to be state variables as the URL parameter contains basically the properties of the page.

src/pages/Scholarships.jsx
```diff
@@ -26,14 +29,55 @@ function ScholarshipsPage() {
...
+    const maxParam = params.max ?? 0;
+    setAmountFilterVals({
+      minAmount: minParam > 0 ? minParam : 0,
+      maxAmount: maxParam > 0 ? maxParam : 0,
+    });
+
+    const newParams = { ...params };
+    if (!Number.isInteger(minParam) || minParam <= 0) delete newParams.min;
+    if (!Number.isInteger(maxParam) || maxParam <= 0) delete newParams.max;
+    history.push({ search: queryString.stringify(newParams) });
```
josueetcom, line 51: Is there a way this can be done outside the `useEffect`? Something like
```
params = ...;
if params.max not valid:
  delete params.max
  history.replace with params ("replace" since the param was "unclean")
if params.min not valid:
  delete params.min
  history.replace with params ("replace" since the param was "unclean")
```
That way the other effects and the rest of the component don't load until the params are cleaned.

src/pages/Scholarships.jsx
```diff
@@ -26,14 +29,55 @@ function ScholarshipsPage() {
...
-      .then(setScholarships)
+      .then((results) => {
+        const filterError =
+          amountFilterVals.maxAmount > 0 &&
+          amountFilterVals.maxAmount <= amountFilterVals.minAmount;
+        if (
+          (amountFilterVals.minAmount === 0 &&
+            amountFilterVals.maxAmount === 0) ||
+          filterError
+        )
```
josueetcom, line 65: With param pruning you can get rid of this too since you'll know the filter values will be valid by the time the component gets a chance to run this effect.

josueetcom (changes requested):
Lots of comments after taking a closer look but I think they'll improve the readability of the code and reduce potential areas of error. The PR is currently at **+142-13**. I'm curious to see how that changes too.

src/pages/Scholarships.jsx
```diff
@@ -26,14 +29,55 @@ function ScholarshipsPage() {
...
+          setScholarships(
+            results.filter((result) =>
+              FilterByAmount(
+                result.data.amount,
+                amountFilterVals.minAmount,
+                amountFilterVals.maxAmount
+              )
+            )
+          );
+      })
```
josueetcom, line 77: Considering the [min/max constants we set for special values](https://github.com/beyondhb1079/s4us/blob/afa48290e9f1d16397c3a9f1c5bd995e763ec8a0/src/types/ScholarshipAmount.ts#L3-L15), with beforehand param pruning, this and `FilterByAmount` can probably be replaced by:
```js
setScholarships(
  results
    .filter(({data}) => !params.minAmount || data.amount.max <= params.minAmount)
    .filter(({data}) => !params.maxAmount || data.amount.min >= params.maxAmount)
)
```
Note how each filter statement checks if that filter is set, letting all scholarships pass through if the filter isn't set.

src/components/AmountFilter.jsx
```diff
@@ -58,12 +75,16 @@ export default function AmountFilter() {
...
             />
           </Grid>
         </Grid>
       </Popover>
     </>
   );
 }
+
+AmountFilter.propTypes = {
+  amountFilterVals: PropTypes.objectOf(PropTypes.number).isRequired,
```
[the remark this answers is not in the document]
edwingl10: Ahh ok. I'll change the variables back to `min` and `max` and rename the query parameters.

src/components/AmountFilter.jsx
edwingl10: that makes sense. I moved that logic to Scholarships.jsx and pass it to AmountFitler as one function instead of 2, and like the function above, it uses an index to determine if it should set `minAmount` or `maxAmount`.

src/pages/Scholarships.jsx
edwingl10: Yeah I got it to work outside of useEffect.

src/pages/Scholarships.jsx
edwingl10: The filterError is to check if the filterMin is less than the filterMax. If that's not the case, should we also prune them from the URL? or would it be fine to keep them and show the original set of scholarships

src/pages/Scholarships.jsx
```diff
@@ -27,13 +30,57 @@ function ScholarshipsPage() {
...
     Scholarships.list({ sortField, sortDir })
-      .then(setScholarships)
+      .then((results) => {
+        const filterError = maxParam > 0 && maxParam <= minParam;
+
+        if (filterError) setScholarships(results);
+        else
+          setScholarships(
+            results.filter(({ data }) => {
+              if (data.amount.type === AmountType.Unknown) return true;
```
edwingl10, line 70: Without this, the Scholarships with Unknown amount don't show.

src/components/AmountFilter.jsx
```diff
@@ -58,12 +61,20 @@ export default function AmountFilter() {
...
           </Grid>
         </Grid>
       </Popover>
     </>
   );
 }
+
+AmountFilter.propTypes = {
+  minAmount: PropTypes.number.isRequired,
+  maxAmount: PropTypes.number.isRequired,
```
josueetcom, line 78: I think the shorter names `min` and `max` are straightforward for this component.

src/components/FilterBar.jsx
```diff
@@ -86,5 +94,9 @@ export default function FilterBar(props) {
 FilterBar.propTypes = {
   changeSortBy: PropTypes.func.isRequired,
   changeSortFormat: PropTypes.func.isRequired,
-  amountFilterVals: PropTypes.objectOf(PropTypes.number).isRequired,
+  queryParams: PropTypes.objectOf(PropTypes.number),
+  onAmountFilterChange: PropTypes.func.isRequired,
```
josueetcom, line 98: Hmm, this pattern will require passing in custom functions for each filter. I think a better pattern might be to accept the `clearQueryParam` and `setQueryParam` functions and then have the custom logic live here. `AmountFilter` is a detail of the implementation of `FilterBar` and it shouldn't really be assumed

src/pages/Scholarships.jsx
```diff
@@ -29,55 +29,58 @@ function ScholarshipsPage() {
...
+      search: queryString.stringify({
+        ...queryString.parse(location.search),
+        [index]: val,
+      }),
     });
+  };
+
+  const params = queryString.parse(location.search, { parseNumbers: true });
+  const minParam = params.minAmount;
+  const maxParam = params.maxAmount;
```
josueetcom, line 48: You could do
```js
const {minAmount, maxAmount} = params;
```

src/pages/Scholarships.jsx
```diff
@@ -29,55 +29,58 @@ function ScholarshipsPage() {
...
 
-  useEffect(() => {
-    const params = queryString.parse(location.search, { parseNumbers: true });
-    const minParam = params.min ?? 0;
-    const maxParam = params.max ?? 0;
-    setAmountFilterVals({
-      minAmount: minParam > 0 ? minParam : 0,
-      maxAmount: maxParam > 0 ? maxParam : 0,
+  // sets min and max amount for the amount filter
+  const setMinMax = (index, val) => {
```
josueetcom, line 37: This is basically the generic [`setParam`](https://github.com/beyondhb1079/s4us/wiki/Guide:-Adding-URL-parameter-support#setparam) function. It's not specific to setting min or max.

src/pages/Scholarships.jsx
```diff
@@ -29,55 +29,58 @@ function ScholarshipsPage() {
...
+    history.push({
+      search: queryString.stringify({
+        ...queryString.parse(location.search),
```
josueetcom, line 40: Define `params` before this function and use this:
```suggestion
        ...params,
```

src/pages/Scholarships.jsx
```diff
@@ -29,55 +29,58 @@ function ScholarshipsPage() {
...
+  const params = queryString.parse(location.search, { parseNumbers: true });
+  const minParam = params.minAmount;
+  const maxParam = params.maxAmount;
 
-    const newParams = { ...params };
-    if (!Number.isInteger(minParam) || minParam <= 0) delete newParams.min;
-    if (!Number.isInteger(maxParam) || maxParam <= 0) delete newParams.max;
-    history.push({ search: queryString.stringify(newParams) });
-  }, [location.search, history]);
+  if ((minParam && !Number.isInteger(minParam)) || minParam === 0) {
```
josueetcom, line 50: Hmm, this doesn't check for invalid values of min. Maybe:
```js
  if (minParam !== undefined || !Number.isInteger(minParam) || minParam <= 0) {
```

src/pages/Scholarships.jsx
```diff
@@ -29,55 +29,58 @@ function ScholarshipsPage() {
...
+    delete params.minAmount;
+    history.replace({ search: queryString.stringify(params) });
+  }
+
+  if ((maxParam && !Number.isInteger(maxParam)) || maxParam === 0) {
```
josueetcom, line 55: Similar comment as above.

src/pages/Scholarships.jsx
josueetcom: Why is that? Scholarships with Unknown amount have a max of -1 (so `data.amount.max < minParam` should always be true) and a min of over a billion (so `data.amount.min > maxParam` should always be true).

I wonder if it's just the UNKNOWN scholarship data you have in your local firestore emulator. Try creating a new UNKNOWN scholarship and see if it shows up.

src/pages/Scholarships.jsx
```diff
@@ -29,55 +29,58 @@ function ScholarshipsPage() {
...
-            )
+            results.filter(({ data }) => {
+              if (data.amount.type === AmountType.Unknown) return true;
+              if (
+                (minParam && data.amount.max < minParam) ||
+                (maxParam && data.amount.min > maxParam)
+              ) {
+                return false;
+              }
+              return true;
```
josueetcom, line 77: This can be written
```js
return (!minParam || data.amount.max <= minParam) && (!maxParam || data.amount.min <= maxParam);
```

src/types/ScholarshipAmount.ts
```diff
@@ -51,7 +51,7 @@ export default class ScholarshipAmount {
           throw new Error(`Invalid range ${min}-${max}`);
         }
         this.min = min ?? 0;
-        this.max = max ?? RANGE_MAX;
+        this.max = max || RANGE_MAX;
```
josueetcom, line 54: Ah I see what you meant now. `max ?? RANGE_MAX` doesn't set `RANGE_MAX` if `max` is set to anything (including 0) but `max || RANGE_MAX` sets the `RANGE_MAX` on 0.

src/pages/Scholarships.jsx
```diff
@@ -29,55 +29,58 @@ function ScholarshipsPage() {
...
-        const filterError =
-          amountFilterVals.maxAmount > 0 &&
-          amountFilterVals.maxAmount <= amountFilterVals.minAmount;
-        if (
-          (amountFilterVals.minAmount === 0 &&
-            amountFilterVals.maxAmount === 0) ||
-          filterError
-        )
-          setScholarships(results);
+        const filterError = maxParam > 0 && maxParam <= minParam;
```
josueetcom, line 64: Hmm, I think it'd be okay to proceed with whatever amount filter values we have. There just won't be any results. Though maybe, since this case should only happen if the URL is tampered with, maybe we should prune them before creating this effect.

josueetcom (changes requested):

src/pages/Scholarships.jsx
josueetcom: Another alternative: in the effect check for min > max and if that's the case then don't bother listing scholarships and `setError` instead

src/pages/Scholarships.jsx
edwingl10: The reason was because both conditions are true and therefore returns false.

src/types/ScholarshipAmount.ts
edwingl10: yes, that's why I had 3 cases checking if min,max or both were set lol

src/components/FilterBar.jsx
```diff
@@ -40,13 +40,27 @@ export default function FilterBar(props) {
...
+  ) {
+    clearQueryParam('minAmount');
+  }
+
+  if (
+    maxAmount !== undefined &&
+    !(Number.isInteger(maxAmount) && maxAmount > 0)
+  ) {
+    clearQueryParam('maxAmount');
+  }
```
josueetcom, line 63: I'd keep this check in Scholarships.jsx so that the effect fetching scholarships doesn't run multiple times in these cases.

src/components/AmountFilter.jsx
```diff
@@ -74,7 +70,7 @@ export default function AmountFilter(props) {
 }
 
 AmountFilter.propTypes = {
-  minAmount: PropTypes.number.isRequired,
-  maxAmount: PropTypes.number.isRequired,
-  onAmountFilterChange: PropTypes.func.isRequired,
+  min: PropTypes.number.isRequired,
+  max: PropTypes.number.isRequired,
+  setQueryParam: PropTypes.func.isRequired,
```
josueetcom, line 75: Hmm, I recommend having separate `onMaxChange` and `onMinChange` functions. It's a bit messy to have a specific component like this understand query parameters.

josueetcom: Unless this component has some sort of constant that gets used elsewhere, e.g. `QUERY_PARAM_MIN_AMOUNT='minAmount'`

src/pages/Scholarships.jsx
```diff
@@ -33,54 +34,48 @@ function ScholarshipsPage() {
...
-  }
-
-  if ((maxParam && !Number.isInteger(maxParam)) || maxParam === 0) {
-    delete params.maxAmount;
-    history.replace({ search: queryString.stringify(params) });
-  }
+  const clearQueryParam = (index) => {
+    const newParams = { ...params };
+    delete newParams[index];
+    history.replace({ search: queryString.stringify(newParams) });
```
josueetcom, line 52:
```suggestion
    delete params[index];
    history.replace({ search: queryString.stringify(params) });
```

src/pages/Scholarships.jsx
```diff
@@ -33,54 +34,48 @@ function ScholarshipsPage() {
...
-                return false;
-              }
-              return true;
+              return intersectsRange(
+                data.amount.min,
+                data.amount.max,
+                minAmount,
+                maxAmount
+              );
             })
```
josueetcom: Ah, with `intersectsRange` I meant as a function of the `ScholarshipAmount` object itself as you can check for `UNKNOWN` there and simplify this to:
```js
results.filter((s) => s.data.amount.intersectsRange(minAmount, maxAmount))
```

Then we can write test cases for all sorts of types of ScholarshipAmount and different ranges.

src/components/FilterBar.jsx
```diff
@@ -40,13 +40,27 @@ export default function FilterBar(props) {
...
 
-  const minAmount = queryParams.minAmount ?? 0;
-  const maxAmount = queryParams.maxAmount ?? 0;
+  const { minAmount, maxAmount } = queryParams;
+
+  if (
+    minAmount !== undefined &&
+    !(Number.isInteger(minAmount) && minAmount > 0)
+  ) {
+    clearQueryParam('minAmount');
```
josueetcom, line 55: We should probably refactor these query param magic strings somewhere. Maybe a constant called `QUERY_PARAM_MIN_AMOUNT` or `QueryParams.MIN_AMOUNT` where there's some `src/lib/QueryParams.js` with a `MIN_AMOUNT` constant?

src/pages/Scholarships.jsx
```diff
@@ -33,54 +34,48 @@ function ScholarshipsPage() {
...
   useEffect(() => {
     // TODO: Create cancellable promises
     Scholarships.list({ sortField, sortDir })
       .then((results) => {
-        const filterError = maxParam > 0 && maxParam <= minParam;
-
-        if (filterError) setScholarships(results);
-        else
+        if (maxAmount > 0 && maxAmount <= minAmount)
+          setError('The minimum amount must be less than the Maximum.');
```
josueetcom, line 60: You can do this check before `Scholarships.list()` is called. Also you can chain a `.then()` to this `.then()` that calls `setError(null)` so that this `.then()` can be a single statement.

josueetcom (changes requested):
I think we're almost there! Approval is probably coming in the next review.

josueetcom:
This has been a wild PR to review since we're just figuring out the pattern to set/clear URL query parameters and my head's been all over the place. Thanks for bearing with me!

src/components/AmountFilter.jsx
josueetcom: Just added some notes at the end of the guide:
https://github.com/beyondhb1079/s4us/wiki/Guide:-Adding-URL-parameter-support

It's subjective whether `AmountFilter` should be aware of query params or not the way `FilterBar` is. Personally I think `onMaxChange` and `onMinChange` functions probably keep this component slightly cleaner.

src/pages/Scholarships.jsx
edwingl10: That's interesting, before when I tried this I was getting the no-param-reassign eslint warning.

src/pages/Scholarships.jsx
edwingl10: Oh I see, my bad. That's pretty cool though.

src/components/FilterBar.jsx
edwingl10: For sure, I removed the pruneQueryParam from FilterBar.jsx since it wasn't being used.

src/pages/Scholarships.jsx
```diff
@@ -46,33 +48,38 @@ function ScholarshipsPage() {
...
+  if (
+    maxAmount !== undefined &&
+    !(Number.isInteger(maxAmount) && maxAmount > 0)
+  ) {
+    pruneQueryParam(QUERY_PARAM_MAX_AMOUNT);
+  }
+
   useEffect(() => {
+    if (maxAmount > 0 && maxAmount <= minAmount)
+      setError('The minimum amount must be less than the Maximum.');
```
josueetcom, line 72: Don't load the scholarships in this case. Also, I think it's ok if the min/max are the same.
```suggestion
    if (maxAmount && maxAmount < minAmount) {
      setError('The minimum amount must be less than the Maximum.');
      return;
    }
```

src/pages/Scholarships.jsx
```diff
@@ -46,33 +48,38 @@ function ScholarshipsPage() {
...
-        }
-      })
+      .then((results) =>
+        setScholarships(
+          results.filter((s) =>
+            s.data.amount.intersectsRange(minAmount, maxAmount)
+          )
+        )
+      )
+      .then(() => (maxAmount === 0 || maxAmount > minAmount) && setError(null))
```
josueetcom, line 82: If you short circuit return before loading any scholarships then this can simply just `setError(null)`

src/types/ScholarshipAmount.ts
```diff
@@ -88,4 +79,12 @@ export default class ScholarshipAmount {
         return '(Unknown amount)';
     }
   }
+
+  intersectsRange(filterMin: number, filterMax: number): boolean {
```
josueetcom, line 83: I would use `min` and `max` here instead as this function could potentially be used for non-filtering purposes.

src/components/FilterBar.jsx
```diff
@@ -3,6 +3,10 @@ import PropTypes from 'prop-types';
 import { makeStyles, Grid } from '@material-ui/core';
 import FilterDropDown from './FilterDropdown';
 import AmountFilter from './AmountFilter';
+import {
+  QUERY_PARAM_MIN_AMOUNT,
+  QUERY_PARAM_MAX_AMOUNT,
+} from '../lib/QueryParams';
```
josueetcom, line 9: I was thinking more like
```js
```
And then using `qParams.MIN_AMOUNT` or something like that. This works too though. Also I'm not sure what the right name for the file should be so it's whatever.

josueetcom (changes requested):
A few nits but otherwise it looks good.

josueetcom (approved):

PR #369, opened 2021-02-02
Refactor ScholarshipList component.

at-syot:
## Refactor code

- move ScholarshipListCard out from ScholarShipList
- components tested in ScholarShipListCard.test
- at first i try to use dayjs lib for formatting scholarship.deadline (Date) -> but i found i don't need it so dayjs lib is removed

## Screenshots (if appropriate):

Screen Shot 2564-02-02 at 22 11 39


src/components/ScholarshipListCard.jsx
```diff
@@ -0,0 +1,111 @@
...
+          </Typography>
+          <Typography
+            variant="body1"
+            color="textSecondary"
+            className={classes.description}>
+            {scholarship.description}
+          </Typography>
+          <Box className={classes.tagWrapper}>
+            {scholarship.tags
+              ? scholarship.tags.map((tag) => (
```
josueetcom, line 84: I think you can simplify `X ? X.foo : null` to `X?.foo`

src/components/ScholarshipListCard.jsx
```diff
@@ -0,0 +1,111 @@
...
+          </Typography>
+          <Typography variant="h5">{scholarship.name}</Typography>
+          <Typography variant="subtitle1" className={classes.subtitle}>
+            {scholarship.city}
+          </Typography>
+          <Typography
+            variant="subtitle1"
+            className={classes.subtitle}
+            gutterBottom>
+            $1,000
```
josueetcom, line 74: `{scholarship.amount.toString()}`

src/components/ScholarshipListCard.jsx
```diff
@@ -0,0 +1,111 @@
...
+          <Typography
+            variant="body1"
+            color="textSecondary"
+            className={classes.description}>
+            {scholarship.description}
+          </Typography>
+          <Box className={classes.tagWrapper}>
+            {scholarship.tags
+              ? scholarship.tags.map((tag) => (
+                  <Box component="span" className={classes.tag}>
```
josueetcom, line 85: I think these are supposed to be [outlined chips](https://material-ui.com/components/chips/#outlined-chips)

src/components/ScholarshipList.jsx
```diff
@@ -1,125 +1,53 @@
...
-                aria-label="share"
-                color="primary"
-                onClick={shareFn(id, data)}>
-                <ShareIcon />
-              </IconButton>
-            </CardActions>
-          </Card>
-        </Grid>
-      ))}
+      <Grid item xs={6}>
```
josueetcom, line 46: You can leave this at `xs={12}` for now. The split view when viewing details is being figured out as part of #31

src/components/ScholarshipList.jsx
```diff
@@ -1,125 +1,53 @@
...
-    lineClamp: 5,
-    '-webkit-box-orient': 'vertical',
-    overflow: 'hidden',
-    whiteSpace: 'pre-line',
-  },
-}));
+import ScholarshipListCard from './ScholarshipListCard';
+
+// move out from component
+// because it don't need to recreate every time component render
```
josueetcom, line 10: Maybe leave this how it was? The [build](https://github.com/beyondhb1079/s4us/pull/369/checks?check_run_id=1815635718) is complaining about unused variables. You can add `// eslint-disable-next-line no-unused-vars` above the declaration or comment this out as it's going to be reused eventually somewhere else in the codebase.

src/components/ScholarshipListCard.jsx
```diff
@@ -0,0 +1,111 @@
...
+
+ScholarshipListCard.propTypes = {
+  scholarship: PropTypes.shape({
+    name: PropTypes.string,
+    city: PropTypes.string,
+    description: PropTypes.string,
+    deadline: PropTypes.instanceOf(Date),
+    tags: PropTypes.arrayOf({ title: PropTypes.string }),
+  }),
+  id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
```
josueetcom, line 105: Should be just a string

src/components/ScholarshipListCard.jsx
```diff
@@ -0,0 +1,111 @@
...
+  tagWrapper: {
+    display: 'flex',
+    flexDirection: 'row',
+    flexWrap: 'wrap',
+  },
+  tag: {
+    marginRight: theme.spacing(2),
+    marginBottom: theme.spacing(2),
+    borderRadius: '50px',
+    border: '1px solid #3FB1B5',
```
josueetcom, line 46: We haven't updated the site theme to match the design yet but the outline color is going to be the same as the primary color. [Outlined chips](https://material-ui.com/components/chips/#outlined-chips) have an easy way to use the theme's primary color.

src/components/ScholarshipListCard.jsx
```diff
@@ -0,0 +1,111 @@
...
+    padding: '12px 18px',
```
josueetcom, line 47: I'd prefer spacing numbers be based on theme spacing so that spacing changes as the theme changes.

src/components/ScholarshipListCard.test.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+import ScholarshipListCard from './ScholarshipListCard';
+
+test('renders ScholarshipListCard', () => {
```
josueetcom, line 6: How about a test case for an empty scholarship?

src/components/ScholarshipListCard.jsx
```diff
@@ -0,0 +1,111 @@
...
+        </CardContent>
+      </CardActionArea>
+    </Card>
+  );
+}
+
+ScholarshipListCard.propTypes = {
+  scholarship: PropTypes.shape({
+    name: PropTypes.string,
+    city: PropTypes.string,
```
josueetcom, line 100: *`organization`. It's the name of the organization

josueetcom (changes requested):
Thank you for the pull request! Overall besides the use of [Chips](https://material-ui.com/components/chips/#chip) I think what's left are just minor nits.

src/components/ScholarshipListCard.jsx
at-syot: +1 Ah, i will change this.

src/components/ScholarshipList.jsx
at-syot: OK

src/components/ScholarshipListCard.test.jsx
at-syot: +1

src/components/ScholarshipListCard.jsx
at-syot: I thought X?.foo work only in .ts || .tsx file ?

josueetcom: Hmm, I think it might still work since we're compiling this project with TypeScript/Babel.

src/components/ScholarshipListCard.jsx
at-syot: After check i had found this comment will able to solve by changing ./src/testdata/scholarships.js.

changing list
- id from number to string,
- amount from number to object contains toString function

at-syot: Reason why we have to change this 
- data.id 
    itself generate by firestore so it should be string but now the testing data is number type
- data.amount 
    because in ScholarshipListCard need -> prop.amount.toString() <- for rendering so it much change from plain number to object contains toString function instead.

at-syot: Can i ?

josueetcom: Ah, testdata/scholarships.js is old from before we had much structure in place. You can update that to match the expected [ScholarshipData interface](https://github.com/beyondhb1079/s4us/blob/main/src/models/Scholarships.ts#L6-L18)

src/components/ScholarshipListCard.test.jsx
at-syot: i would like to get more meaning of empty scholarship.
Is scholarship object shape exist with all default value or undefined scholarship ?

josueetcom: Hmm, I guess one without any optional fields set. The only fields we require for a scholarship are name, amount, description, deadline, and website. All other fields are currently optional.

I wouldn't worry too much about it. Feel free to merge without adding it.

josueetcom (approved):
I think this looks good! Thanks!

src/components/ScholarshipListCard.test.jsx
at-syot: Thank you.

at-syot:
Thx for approve my code, 
I have no permission to merge by myself. please.

PR #371, opened 2021-02-06
Optimize style validation job by only installing Prettier

josueetcom:
Our style job only relies on prettier so rather than install everything from package.json let's just install prettier alone.

## Motivation and Context

If you look at the [Validation actions](https://github.com/beyondhb1079/s4us/actions?query=workflow%3AValidation) the `style` validation job is usually the fastest at around `1m10s` or so. But... this can be made faster by ignoring all other dependencies. 

## How Has This Been Tested?

Look at how fast `style` jobs have run in the [checks](371/checks) for this PR for each commit. Blazing fast under 10 seconds!

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -55,20 +55,10 @@ jobs:
     steps:
       - name: Checkout the repo
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


Jobhdez (approved):
looks good to me.

Issue #312, opened 2021-01-12
User home: styling


josueetcom:
According to the [Figma design](https://www.figma.com/file/nwrW3UebcvSWyQZGb5ixNr/DreamScholar?node-id=334%3A1080)

PR #372, opened 2021-02-06
User home styling (part 1)

Jobhdez:
I did 60 percent of the styling in userHome.jsx. (it is 1:30am and wanted to push before going to sleep.) 

I did `yarn start` and everything looked similar to the figma design. 

I also would like to get some feedback so I pushed.

if its ok then I will finish the rest in a different branch. I did most of it. the images in the figma design is what is left to do. so seeing it this way I completed more than 60 percent.


src/pages/UserHome.jsx
```diff
@@ -67,14 +73,35 @@ export default function UserHome() {
...
+      </center>
+      <center>
+        <Typography
+          variant="h10"
+          component="h2"
+          gutterButton
+          color="primary"
+          className={classes.addScholarship}>
+          <p>
+            <a href="more-scholarships">Add A Scholarship</a>
```
gonzalo1330, line 101: I don't think at the moment it redirects to `/scholarships/new`, what its doing is refreshing the page.

josueetcom (changes requested):
Btw there's still some weird padding issues, which are extremely visible on small screens:
image

src/pages/UserHome.jsx
```diff
@@ -35,6 +38,18 @@ const useStyles = makeStyles((theme) => ({
   addScholarship: {
     marginTop: theme.spacing(6),
   },
+  addTextAndLink: {
+    textAlign: 'center',
+  },
+  addIconSpacing: {
```
josueetcom, line 44: `inboxIcon` is a better name. `inboxIcon` is more descriptive of what's being styled than `addIcon`. Also, not having `spacing` in name allows it to be used for styling the icon in non-spacing related ways.

src/pages/UserHome.jsx
```diff
@@ -35,6 +38,18 @@ const useStyles = makeStyles((theme) => ({
...
+    marginTop: theme.spacing(6),
+  },
+  centerTextAndLink: {
```
josueetcom, line 47: This is descriptive of the specific child elements of the element styled but maybe a better name might be `noScholarshipsBox` as it's the `Box` containing elements for when there's no scholarships.

src/pages/UserHome.jsx
```diff
@@ -81,27 +95,27 @@ export default function UserHome() {
           </Button>
         </Grid>
       </Grid>
-      <center>
+      <Box className={classes.centerIcon}>
+        <InboxIcon
+          style={{ fontSize: 200 }}
+          className={classes.addIconSpacing}
+        />
```
josueetcom, line 102: I suggest making this a part of the `Box` below instead of a separate box. They're sort of belong together if you consider how:
1. They're all center aligned
2. They only appear when there are no scholarships
3. They're next to each other visually speaking.

src/pages/UserHome.jsx
```diff
@@ -81,27 +95,27 @@ export default function UserHome() {
...
-          variant="h10"
-          component="h2"
-          gutterButton
-          color="primary"
-          className={classes.addScholarship}>
-          <p>
-            <a href="more-scholarships">Add A Scholarship</a>
-          </p>
+
+        <Typography variant="h10" component="h2" gutterButton color="primary">
```
josueetcom, line 113: *`gutterBottom`, though in this case it might not be needed since there isn't anything underneath.

josueetcom, line 113: `"h10"` is not a valid variant if you look at the type for `variant` in [Typography API](https://material-ui.com/api/typography/). No variant is probably what you want, but if not take a look at the [Typography Component section](https://material-ui.com/components/typography/#component).

src/pages/UserHome.jsx
```diff
@@ -29,6 +32,24 @@ const useStyles = makeStyles((theme) => ({
...
   },
+  noScholarshipsAdded: {
+    marginTop: theme.spacing(12),
+  },
+  addScholarship: {
+    marginTop: theme.spacing(6),
+  },
+  addTextAndLink: {
+    textAlign: 'center',
+  },
```
josueetcom, line 43: Neither of these classes seem to be used anywhere :/

src/pages/UserHome.jsx
```diff
@@ -1,7 +1,10 @@
 import React from 'react';
+import Box from '@material-ui/core/Box';
```
josueetcom, line 2: Tip: instead of this you can also import `Box` in the long list of imports from `@material-ui/core`.

josueetcom (changes requested):
Btw there's **still** weird padding issues, which are extremely visible on small screens:
image

This has to do with some of the padding/margin styles that were present before you started working on this issue.

josueetcom (commented):
The new changes look good. But there's still the padding/spacing issues.

josueetcom:
Helps with #312

josueetcom (approved):
Approved for initial changes

josueetcom (approved):

PR #375, opened 2021-02-07
Avatar transition

gonzalo1330:
Removed the logic from ProfileMenu to HeaderNavMenu for displaying Avatar, allowing the `<Zoom in>` effect to work on the Avatar.

## Motivation and Context

Fixes #

## How Has This Been Tested?

On localhost:3000 the zoom in effect worked when a user is logged in

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-375.onrender.com.
2. Login
3. Refresh page and notice the avatar transition with a zoom in effect

## Checklist:


src/components/ProfileDropdown.jsx
```diff
@@ -82,12 +82,11 @@ export default function ProfileDropdown() {
...
     <>
-      <Avatar src={user.photoURL} onClick={handleClick} />
       <StyledMenu
-        anchorEl={dropMenu}
+        anchorEl={anchorEl}
         keepMounted
-        open={Boolean(dropMenu)}
-        onClose={handleClose}>
+        open={Boolean(anchorEl)}
+        onClose={closeDropMenu}>
```
josueetcom, line 89: Shouldn't this just be `handleClose`?

src/components/ProfileDropdown.jsx
```diff
@@ -82,12 +82,11 @@ export default function ProfileDropdown() {
 
   return (
     <>
```
josueetcom, line 84: This can be removed now that there's a single child.

src/components/ProfileDropdown.jsx
gonzalo1330: Yes indeed it should be

josueetcom (approved):

PR #376, opened 2021-02-07
Make validation checks cache node_modules instead

josueetcom:
If you look at old validation workflows like [this one](https://github.com/beyondhb1079/s4us/actions/runs/542566911) you'll see that the **build**, **lint**, and **test** checks all spend around 1 minute dealing with the yarn cache and installing dependencies.

Since our `yarn.lock` file tells us exactly which dependencies to use, if we instead cache `node_modules` and skip the installation step on cache hits, this'll allow us to:
1. Speed up the caching step because node_modules is more than 50% smaller than the yarn cache directory.
2. Skip installation steps entirely when we get a cache hit.

In total, this speeds up each of these checks by about 50 seconds!

E.g. [before](https://github.com/beyondhb1079/s4us/actions/runs/542566911) vs [after](https://github.com/beyondhb1079/s4us/actions/runs/544905771)

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -19,7 +19,7 @@ jobs:
         id: modules-cache
         with:
           path: node_modules
-          key: ${{ runner.os }}-yarn-${{ hashFiles('yarn.lock') }}
+          key: ${{ runner.os }}-modules-${{ hashFiles('yarn.lock') }}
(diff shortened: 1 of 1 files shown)


gonzalo1330:
Wow reduce speed time by about half its previous time

gonzalo1330 (approved):

PR #379, opened 2021-02-08
sorting URL parameter support

### LINKED ISSUE #277: Sorting options: URL parameter support
URL parameters usually encode information about filters to make pages easier to share. E.g. compare the `q` and `start` parameter values with what's presented in the query and the page number here: https://www.google.com/search?q=url+parameters+filter&start=10

- [ ] Update URL parameters when sorting option is applied
- [ ] Parse URL parameters to determine sorting option to apply or default if unparseable
- [ ] Ensure sorting option component reflects URL parameters
- [ ] Ensure fetched data reflects sorting option set via the URL deadline

gonzalo1330:
Appends to the site url the sorting values to provide a shareable link.

## Motivation and Context
One user can share with anyone the sorting order of scholarships as they update the list while navigating the site.

Fixes #277 

## How Has This Been Tested?

Currently in the localhost:3000 when I update the sorting options in the /scholarship page it does update.

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-379.onrender.com.
2. Select the Scholarships link in the header.
3. Change the sorting option on the upper right corner.
4. After you have selected and the scholarships list is updated click on the website URL and you will now see the updated URL parameters.

## Checklist:

### DIFF
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -31,26 +31,26 @@ const grades = {
 };
 
 const sortItems = {
-  deadlineAsc: 'Deadline (Earliest to Latest)',
-  deadlineDesc: 'Deadline (Latest to Earliest)',
-  amountAsc: 'Amount (Low to High)',
-  amountDesc: 'Amount (High to Low)',
+  'deadline.asc': 'Deadline (Earliest to Latest)',
+  'deadline.desc': 'Deadline (Latest to Earliest)',
+  'amount.asc': 'Amount (Low to High)',
+  'amount.desc': 'Amount (High to Low)',
 };
 
 const sortOptions = {
-  deadlineAsc: {
+  'deadline.asc': {
     field: 'deadline',
     dir: 'asc',
   },
-  deadlineDesc: {
+  'deadline.desc': {
     field: 'deadline',
     dir: 'desc',
   },
-  amountAsc: {
+  'amount.asc': {
     field: 'amount.min',
     dir: 'asc',
   },
-  amountDesc: {
+  'amount.desc': {
     field: 'amount.max',
     dir: 'desc',
   },
@@ -84,7 +84,7 @@ export default function FilterBar(props) {
         <FilterDropDown
           label="Sorting"
           items={sortItems}
-          value={queryParams.sortBy ?? 'deadlineAsc'}
+          value={queryParams.sortBy ?? 'deadline.asc'}
           removeNone
           onChange={updateSortingParams}
         />
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -30,31 +30,32 @@ const grades = {
   16: 'College Senior',
 };
 
-const sortItems = {
-  'deadline.asc': 'Deadline (Earliest to Latest)',
-  'deadline.desc': 'Deadline (Latest to Earliest)',
-  'amount.asc': 'Amount (Low to High)',
-  'amount.desc': 'Amount (High to Low)',
-};
-
 const sortOptions = {
   'deadline.asc': {
+    desc: 'Deadline (Earliest to Latest)',
     field: 'deadline',
     dir: 'asc',
   },
   'deadline.desc': {
+    desc: 'Deadline (Latest to Earliest)',
     field: 'deadline',
     dir: 'desc',
   },
   'amount.asc': {
+    desc: 'Amount (Low to High)',
     field: 'amount.min',
     dir: 'asc',
   },
   'amount.desc': {
+    desc: 'Amount (High to Low)',
     field: 'amount.max',
     dir: 'desc',
   },
 };
+
+const sortItems = Object.fromEntries(
+  Object.entries(sortOptions).map(([k, v]) => [k, v.desc])
+);
 
 export default function FilterBar(props) {
   const { setQueryParam, queryParams } = props;
@@ -62,9 +63,7 @@ export default function FilterBar(props) {
   const { minAmount, maxAmount } = queryParams;
 
   function updateSortingParams(sortOption) {
-    if (sortOption in sortOptions) {
-      setQueryParam('sortBy', sortOption);
-    }
+    setQueryParam('sortBy', sortOption);
   }
 
   return (
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -37,7 +37,7 @@ const sortingOptions = {
 };
 
 export default function FilterBar(props) {
-  const { changeSortBy, changeSortFormat } = props;
+  const { changeSortBy, changeSortFormat, changeURL } = props;
   const classes = useStyles();
 
   function updateSorting(sortingOption) {
@@ -45,18 +45,22 @@ export default function FilterBar(props) {
       case 'deadlineSoon':
         changeSortBy('deadline');
         changeSortFormat('asc');
+        changeURL('sortField=deadline&sortDir=asc');
         break;
       case 'deadlineLatest':
         changeSortBy('deadline');
         changeSortFormat('desc');
+        changeURL('sortField=deadline&sortDir=desc');
         break;
       case 'amountLow':
         changeSortBy('amount.min');
         changeSortFormat('asc');
+        changeURL('sortField=amount.min&sortDir=asc');
         break;
       case 'amountHigh':
         changeSortBy('amount.max');
         changeSortFormat('desc');
+        changeURL('sortField=amount.max&sortDir=desc');
         break;
       default:
     }
@@ -86,4 +90,5 @@ export default function FilterBar(props) {
 FilterBar.propTypes = {
   changeSortBy: PropTypes.func.isRequired,
   changeSortFormat: PropTypes.func.isRequired,
+  changeURL: PropTypes.func.isRequired,
 };
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { makeStyles, Grid } from '@material-ui/core';
 import FilterDropDown from './FilterDropdown';
 import AmountFilter from './AmountFilter';
-import qParams from '../lib/QueryParams';
+import { qParams } from '../lib/QueryParams';
 
 const useStyles = makeStyles(() => ({
   root: {
@@ -37,14 +37,34 @@ const sortingOptions = {
   amountHigh: 'Amount (High to Low)',
 };
 
+const sortParams = {
+  deadlineSoon: {
+    field: 'deadline',
+    dir: 'asc',
+  },
+  deadlineLatest: {
+    field: 'deadline',
+    dir: 'desc',
+  },
+  amountLow: {
+    field: 'amount.min',
+    dir: 'asc',
+  },
+  amountHigh: {
+    field: 'amount.max',
+    dir: 'desc',
+  },
+};
+
 export default function FilterBar(props) {
-  const { setSort, setQueryParam, queryParams } = props;
+  const { setQueryParam, queryParams } = props;
   const classes = useStyles();
   const { minAmount, maxAmount } = queryParams;
 
   function updateSortingParams(sortOption) {
-    if (sortOption in sortingOptions) {
-      setSort(sortOption);
+    if (sortOption in sortParams) {
+      setQueryParam('sortDir', sortParams[sortOption].dir);
+      setQueryParam('sortField', sortParams[sortOption].field);
     }
   }
 
@@ -75,7 +95,6 @@ export default function FilterBar(props) {
 }
 
 FilterBar.propTypes = {
-  setSort: PropTypes.func.isRequired,
   setQueryParam: PropTypes.func.isRequired,
   queryParams: PropTypes.objectOf(PropTypes.number),
 };
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { makeStyles, Grid } from '@material-ui/core';
 import FilterDropDown from './FilterDropdown';
 import AmountFilter from './AmountFilter';
-import { qParams, sortParams } from '../lib/QueryParams';
+import qParams from '../lib/QueryParams';
 
 const useStyles = makeStyles(() => ({
   root: {
@@ -30,11 +30,30 @@ const grades = {
   16: 'College Senior',
 };
 
-const sortingOptions = {
+const sortItems = {
   deadlineAsc: 'Deadline (Earliest to Latest)',
   deadlineDesc: 'Deadline (Latest to Earliest)',
   amountAsc: 'Amount (Low to High)',
   amountDesc: 'Amount (High to Low)',
+};
+
+const sortOptions = {
+  deadlineAsc: {
+    field: 'deadline',
+    dir: 'asc',
+  },
+  deadlineDesc: {
+    field: 'deadline',
+    dir: 'desc',
+  },
+  amountAsc: {
+    field: 'amount.min',
+    dir: 'asc',
+  },
+  amountDesc: {
+    field: 'amount.max',
+    dir: 'desc',
+  },
 };
 
 export default function FilterBar(props) {
@@ -43,7 +62,7 @@ export default function FilterBar(props) {
   const { minAmount, maxAmount } = queryParams;
 
   function updateSortingParams(sortOption) {
-    if (sortOption in sortParams) {
+    if (sortOption in sortOptions) {
       setQueryParam('sortBy', sortOption);
     }
   }
@@ -64,8 +83,8 @@ export default function FilterBar(props) {
         Sort by
         <FilterDropDown
           label="Sorting"
-          items={sortingOptions}
-          value="sorting"
+          items={sortItems}
+          value={queryParams.sortBy ?? 'deadlineAsc'}
           removeNone
           onChange={updateSortingParams}
         />
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -40,7 +40,7 @@ export default function FilterBar(props) {
   const { setSort } = props;
   const classes = useStyles();
 
-  function udpateSortingParams(sortOption) {
+  function updateSortingParams(sortOption) {
     if (sortOption in sortingOptions) {
       setSort(sortOption);
     }
@@ -60,7 +60,7 @@ export default function FilterBar(props) {
           items={sortingOptions}
           defaultValue="deadlineSoon"
           removeNone
-          onChange={udpateSortingParams}
+          onChange={updateSortingParams}
         />
       </Grid>
     </Grid>
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -40,6 +40,12 @@ export default function FilterBar(props) {
   const { setSort } = props;
   const classes = useStyles();
 
+  function udpateSortingParams(sortOption) {
+    if (sortOption in sortingOptions) {
+      setSort(sortOption);
+    }
+  }
+
   return (
     <Grid container spacing={2} className={classes.root}>
       <Grid item className={classes.alignText}>
@@ -54,6 +60,7 @@ export default function FilterBar(props) {
           items={sortingOptions}
           defaultValue="deadlineSoon"
           removeNone
+          onChange={udpateSortingParams}
           setSort={setSort}
         />
       </Grid>
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -37,26 +37,22 @@ const sortingOptions = {
 };
 
 export default function FilterBar(props) {
-  const { setSortField, setSortDir } = props;
+  const { setSort } = props;
   const classes = useStyles();
 
   function updateSorting(sortingOption) {
     switch (sortingOption) {
       case 'deadlineSoon':
-        setSortField('deadline');
-        setSortDir('asc');
+        setSort('deadlineSoon');
         break;
       case 'deadlineLatest':
-        setSortField('deadline');
-        setSortDir('desc');
+        setSort('deadlineLatest');
         break;
       case 'amountLow':
-        setSortField('amount.min');
-        setSortDir('asc');
+        setSort('amountLow');
         break;
       case 'amountHigh':
-        setSortField('amount.max');
-        setSortDir('desc');
+        setSort('amountHigh');
         break;
       default:
     }
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -40,24 +40,6 @@ export default function FilterBar(props) {
   const { setSort } = props;
   const classes = useStyles();
 
-  function updateSorting(sortingOption) {
-    switch (sortingOption) {
-      case 'deadlineSoon':
-        setSort('deadlineSoon');
-        break;
-      case 'deadlineLatest':
-        setSort('deadlineLatest');
-        break;
-      case 'amountLow':
-        setSort('amountLow');
-        break;
-      case 'amountHigh':
-        setSort('amountHigh');
-        break;
-      default:
-    }
-  }
-
   return (
     <Grid container spacing={2} className={classes.root}>
       <Grid item className={classes.alignText}>
@@ -72,7 +54,7 @@ export default function FilterBar(props) {
           items={sortingOptions}
           defaultValue="deadlineSoon"
           removeNone
-          onChange={updateSorting}
+          setSort={setSort}
         />
       </Grid>
     </Grid>
@@ -80,6 +62,5 @@ export default function FilterBar(props) {
 }
 
 FilterBar.propTypes = {
-  setSortField: PropTypes.func.isRequired,
-  setSortDir: PropTypes.func.isRequired,
+  setSort: PropTypes.func.isRequired,
 };
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -37,30 +37,26 @@ const sortingOptions = {
 };
 
 export default function FilterBar(props) {
-  const { changeSortBy, changeSortFormat, changeURL } = props;
+  const { setSortField, setSortDir } = props;
   const classes = useStyles();
 
   function updateSorting(sortingOption) {
     switch (sortingOption) {
       case 'deadlineSoon':
-        changeSortBy('deadline');
-        changeSortFormat('asc');
-        changeURL('sortField=deadline&sortDir=asc');
+        setSortField('deadline');
+        setSortDir('asc');
         break;
       case 'deadlineLatest':
-        changeSortBy('deadline');
-        changeSortFormat('desc');
-        changeURL('sortField=deadline&sortDir=desc');
+        setSortField('deadline');
+        setSortDir('desc');
         break;
       case 'amountLow':
-        changeSortBy('amount.min');
-        changeSortFormat('asc');
-        changeURL('sortField=amount.min&sortDir=asc');
+        setSortField('amount.min');
+        setSortDir('asc');
         break;
       case 'amountHigh':
-        changeSortBy('amount.max');
-        changeSortFormat('desc');
-        changeURL('sortField=amount.max&sortDir=desc');
+        setSortField('amount.max');
+        setSortDir('desc');
         break;
       default:
     }
@@ -88,7 +84,6 @@ export default function FilterBar(props) {
 }
 
 FilterBar.propTypes = {
-  changeSortBy: PropTypes.func.isRequired,
-  changeSortFormat: PropTypes.func.isRequired,
-  changeURL: PropTypes.func.isRequired,
+  setSortField: PropTypes.func.isRequired,
+  setSortDir: PropTypes.func.isRequired,
 };
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { makeStyles, Grid } from '@material-ui/core';
 import FilterDropDown from './FilterDropdown';
 import AmountFilter from './AmountFilter';
-import { qParams } from '../lib/QueryParams';
+import { qParams, sortParams } from '../lib/QueryParams';
 
 const useStyles = makeStyles(() => ({
   root: {
@@ -31,29 +31,10 @@ const grades = {
 };
 
 const sortingOptions = {
-  deadlineSoon: 'Deadline (Earliest to Latest)',
-  deadlineLatest: 'Deadline (Latest to Earliest)',
-  amountLow: 'Amount (Low to High)',
-  amountHigh: 'Amount (High to Low)',
-};
-
-const sortParams = {
-  deadlineSoon: {
-    field: 'deadline',
-    dir: 'asc',
-  },
-  deadlineLatest: {
-    field: 'deadline',
-    dir: 'desc',
-  },
-  amountLow: {
-    field: 'amount.min',
-    dir: 'asc',
-  },
-  amountHigh: {
-    field: 'amount.max',
-    dir: 'desc',
-  },
+  deadlineAsc: 'Deadline (Earliest to Latest)',
+  deadlineDesc: 'Deadline (Latest to Earliest)',
+  amountAsc: 'Amount (Low to High)',
+  amountDesc: 'Amount (High to Low)',
 };
 
 export default function FilterBar(props) {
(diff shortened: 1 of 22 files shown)


src/pages/Scholarships.jsx
```diff
@@ -21,11 +24,23 @@ const useStyles = makeStyles((theme) => ({
...
+  const location = useLocation();
+  const setFoo = (val) => {
+    history.push({
+      search: queryString.stringify({
+        ...queryString.parse(location.search),
+        sorting: val,
+      }),
+    });
+  };
+  const foo = queryString.parse(location.search).foo ?? DEFAULT_SORT_URL_VALUE;
```
josueetcom, line 43:
```suggestion
  const sortDir = queryString.parse(location.search).sortDir ?? DEFAULT_SORT_DIR;
```

josueetcom: TODO: Make sure the sortDir and sortField values make sense. Otherwise use the default values.

src/pages/Scholarships.jsx
```diff
@@ -21,11 +24,23 @@ const useStyles = makeStyles((theme) => ({
...
 function ScholarshipsPage() {
+  const classes = useStyles();
   const [scholarships, setScholarships] = useState([]);
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState();
   const [sortField, setSortField] = useState('deadline');
   const [sortDir, setSortDir] = useState('asc');
+  const history = useHistory();
+  const location = useLocation();
+  const setFoo = (val) => {
```
josueetcom, line 35: You can call this `setSort` and then use this directly in FilterBar.jsx

gonzalo1330:
So URL validation is working correct, if the url `sortDir` or `sortField` are invalid then default of each are used. At the moment the setSortDir function is working properly when the sorting options are changed in the scholarships page but the setSortField seems to not be working properly as the URL value is not updated. However, when the manually changing the URL sortField to `amount.min or amount.max or deadline' the sorting is updated correct. Do you mind inspecting my code and seeing if you can help me find where in the code the issue for changing the `sortField` is occurring?

src/pages/Scholarships.jsx
gonzalo1330: Hm I am not using it directly in the FilterBar.jsx instead in FilterDropdown.jsx, as it was kind of repetitive if you see in the previous commits.

src/pages/Scholarships.jsx
```diff
@@ -21,11 +24,51 @@ const useStyles = makeStyles((theme) => ({
...
+    } else if (val === 'deadlineLatest') {
+      sortField = 'deadline';
+      sortDir = 'desc';
+    } else if (val === 'amountLow') {
+      sortField = 'amount.min';
+      sortDir = 'asc';
+    } else if (val === 'amountHigh') {
+      sortField = 'amount.max';
+      sortDir = 'desc';
+    }
```
gonzalo1330, line 64: I can also do this as a `switch` but it does take more lines to do the same thing and also this logic atleast to me reads more clear than using switch statement.

josueetcom: How about storing these in a dictionary like object so you can skip the `if/else` and `switch` statements and just check if the sorting option key is in the dictionary? E.g.
```js
const sortingOptions = {
  'deadlineSoon': {
    field: 'deadline',
    dir: 'asc',
  },
  // etc.
}
```

src/pages/Scholarships.jsx
```diff
@@ -21,11 +24,51 @@ const useStyles = makeStyles((theme) => ({
...
+    history.push({
+      search: queryString.stringify({
+        sortField,
+        sortDir,
+      }),
```
josueetcom, line 69: This is actually stripping all the other query params besides sortField and sortDir. You can see this by adding a different query param in the URL (e.g. `foo=bar`) and watching it disappear when you select a sorting option.

src/pages/Scholarships.jsx
```diff
@@ -21,11 +24,51 @@ const useStyles = makeStyles((theme) => ({
...
   const [error, setError] = useState();
-  const [sortField, setSortField] = useState('deadline');
-  const [sortDir, setSortDir] = useState('asc');
+  const history = useHistory();
+  const location = useLocation();
+
+  const validField = Boolean(
+    queryString.parse(location.search).sortField === 'deadline' ||
+      queryString.parse(location.search).sortField === 'amount.min' ||
+      queryString.parse(location.search).sortField === 'amount.max'
```
josueetcom, line 37: I wonder if it'd be better to use `amount` instead of `amount.min`/`amount.max` in the `sortField`. We can determine whether to use min/max based on the `sortDir`.

josueetcom (changes requested):

src/components/FilterDropdown.jsx
```diff
@@ -52,12 +52,12 @@ FilterDropdown.propTypes = {
   defaultValue: PropTypes.string,
   items: PropTypes.objectOf(PropTypes.string).isRequired,
   removeNone: PropTypes.bool,
-  onChange: PropTypes.func,
+  setSort: PropTypes.func,
```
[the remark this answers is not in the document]
gonzalo1330: Ok I changed it back

src/pages/Scholarships.jsx
```diff
@@ -21,11 +24,51 @@ const useStyles = makeStyles((theme) => ({
...
+    queryString.parse(location.search).sortDir === 'asc' ||
+      queryString.parse(location.search).sortDir === 'desc'
+  );
+
+  let sortField = validField
+    ? queryString.parse(location.search).sortField
+    : DEFAULT_SORT_FIELD;
+  let sortDir = validDir
+    ? queryString.parse(location.search).sortDir
+    : DEFAULT_SORT_DIR;
```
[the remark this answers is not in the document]
gonzalo1330: Ok I did not think that was possible but I saw on the example that you provided in the URL parameter guide. I made the change

src/config/constants.js
```diff
@@ -3,3 +3,5 @@ export const BRAND_NAME = 'DreamScholars';
 const params = new URLSearchParams(window.location.search);
 export const EXPERIMENT_SHOW_FULL_PROFILE_MENU =
   params.get('experimentShowFullProfileMenu') === '1';
+export const DEFAULT_SORT_FIELD = 'deadline';
+export const DEFAULT_SORT_DIR = 'asc';
```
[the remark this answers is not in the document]
gonzalo1330: Ok in the mean time I guess it can stay the same. Unless, I define it within the Scholarships.jsx as a const instead of importing it?

src/pages/Scholarships.jsx
gonzalo1330: Ok yea nice that is a great idea, right now how I am currently checking if the _key_ exist in the dictionary is by using `.hasOwnPropery(_key_)` is that the correct way to do it? Its working at the moment but im not sure if that's the best way to check.

src/pages/Scholarships.jsx
gonzalo1330: Ah ok so how can I work around not stripping all the other query params and still add sortField and sortDir to the url?

src/pages/Scholarships.jsx
gonzalo1330: Well wouldnt it still display the amount.min and amount.max in the url?

src/pages/Scholarships.jsx
gonzalo1330: I think I was able to fix it with the most current commit, can you take a look and confirm?

src/pages/Scholarships.jsx
gonzalo1330: Actually I got it, the best way is to use the `in` key word so 
`if (sortOption in sortingOptions) {}`

src/pages/Scholarships.jsx
```diff
@@ -30,44 +30,48 @@ function ScholarshipsPage() {
...
+  };
 
   const validField = Boolean(
-    queryString.parse(location.search).sortField === 'deadline' ||
-      queryString.parse(location.search).sortField === 'amount.min' ||
-      queryString.parse(location.search).sortField === 'amount.max'
+    params.sortField === 'deadline' ||
+      params.sortField === 'amount.min' ||
+      params.sortField === 'amount.max'
   );
```
josueetcom: No need to cast this to a boolean. It's already a boolean.

src/config/constants.js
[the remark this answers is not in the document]
josueetcom: Either or is fine. The constants will likely find a new home eventually at some point.

src/pages/Scholarships.jsx
```diff
@@ -30,44 +30,48 @@ function ScholarshipsPage() {
...
   const validDir = Boolean(
-    queryString.parse(location.search).sortDir === 'asc' ||
-      queryString.parse(location.search).sortDir === 'desc'
+    params.sortDir === 'asc' || params.sortDir === 'desc'
   );
```
josueetcom: Same comment as above.

src/pages/Scholarships.jsx
gonzalo1330: Alright I changed the verification for paras.sortField === 'amount' to use the `sortDir` IF it is valid, to set the sorting parameters appropriately, check and let me know if that is what you were thinking or if it should just go back to how it was.

src/pages/Scholarships.jsx
gonzalo1330: ok changed

src/components/FilterBar.jsx
```diff
@@ -76,14 +60,13 @@ export default function FilterBar(props) {
           items={sortingOptions}
           defaultValue="deadlineSoon"
```
josueetcom, line 61: Look at the sort dropdown here:
https://s4us-pr-379.onrender.com/scholarships?sortDir=asc&sortField=amount.min

It's value isn't being reflected. Looks like this needs to be set based on what's currently in the query param.

src/components/FilterBar.jsx
```diff
@@ -37,28 +37,12 @@ const sortingOptions = {
...
-      case 'amountLow':
-        changeSortBy('amount.min');
-        changeSortFormat('asc');
-        break;
-      case 'amountHigh':
-        changeSortBy('amount.max');
-        changeSortFormat('desc');
-        break;
-      default:
+  function udpateSortingParams(sortOption) {
```
josueetcom, line 43: *update

src/components/FilterBar.jsx
```diff
@@ -76,14 +60,13 @@ export default function FilterBar(props) {
...
         />
       </Grid>
     </Grid>
   );
 }
 
 FilterBar.propTypes = {
-  changeSortBy: PropTypes.func.isRequired,
-  changeSortFormat: PropTypes.func.isRequired,
+  setSort: PropTypes.func.isRequired,
```
josueetcom, line 71: So besides a way to update the sorting params, you'll need to pass in the sorting params themselves. Now, this is true for basically every filter param so you can imagine that if there are 10 filters then there we would need properties for at least 10 filter values as well as properties for at least 10 functions for updating them. This doesn't scale well so I suggest instead passing `params` or `queryParams` here and a generic `setParam`/`setQueryParam` function that can be reused by other filter params.

src/pages/Scholarships.jsx
```diff
@@ -21,11 +24,60 @@ const useStyles = makeStyles((theme) => ({
...
+      if (params.sortDir === 'asc') {
+        setSort('amountLow');
+      } else {
+        setSort('amountHigh');
+      }
+    }
+  }
+
+  const sortField = validField ? params.sortField : DEFAULT_SORT_FIELD;
+  const sortDir = validDir ? params.sortDir : DEFAULT_SORT_DIR;
```
josueetcom, line 80: I suggest defining `sortField` and `sortDir` before `validField` and `validDir` and instead [pruning](https://github.com/beyondhb1079/s4us/wiki/Guide:-Adding-URL-parameter-support#pruneparam) each invalid parameter so that by the time the code reaches this spot the sorting direction and field are valid.

josueetcom (changes requested):

src/pages/Scholarships.jsx
josueetcom: This is still relevant btw.

src/components/FilterBar.jsx
gonzalo1330: ok

src/pages/Scholarships.jsx
gonzalo1330: Wait what? Rn its using the validField to set them so how would I check it if its defined first

src/components/FilterBar.jsx
gonzalo1330: How would I do that?

src/pages/Scholarships.jsx
gonzalo1330: Ahhhh what will fix it then

src/config/constants.js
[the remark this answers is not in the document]
gonzalo1330: Ok ill just leave as is, probably should file an issue to know to get that changed at some point

src/components/FilterBar.jsx
josueetcom: Take in the sorting option and set this. Also `value` may be a better name since it's no longer the default value.

josueetcom: I think once you update your branch the value should be available in the `queryParams` prop as `sort` something. You might need to translate it to `deadlineSoon` or whatever the corresponding value is here.

src/pages/Scholarships.jsx
josueetcom: I guess what I mean is this:

1. "define `sortField` and `sortDir` before (where) `validField` and `validDir` (are defined)":
```js
  const sortField = params.sortField ?? DEFAULT_SORT_FIELD;
  const sortDir = params.sortDir ?? DEFAULT_SORT_DIR;
```
2. Now what to do about `validField` and `validDir`? "[prune](https://github.com/beyondhb1079/s4us/wiki/Guide:-Adding-URL-parameter-support#pruneparam) each invalid parameter so that by the time the code reaches this spot the sorting direction and field are valid":
```js
  if (/* validField check fails */) {
    pruneQueryParam('sortField');
  }
  if (/* validDir check fails */) {
    pruneQueryParam('sortDir');
  }
  // Reaching here means sortDir and sortField are valid.
```

gonzalo1330: Ok done

src/components/FilterBar.jsx
```diff
@@ -37,14 +37,34 @@ const sortingOptions = {
...
+  const { setQueryParam, queryParams } = props;
   const classes = useStyles();
   const { minAmount, maxAmount } = queryParams;
 
   function updateSortingParams(sortOption) {
-    if (sortOption in sortingOptions) {
-      setSort(sortOption);
+    if (sortOption in sortParams) {
+      setQueryParam('sortDir', sortParams[sortOption].dir);
+      setQueryParam('sortField', sortParams[sortOption].field);
```
gonzalo1330, line 67: Not sure why it only updates and displays on the URL which ever is last? Starting to get frustrating as it should work

josueetcom: Hmm, that's because you're always calling `setQueryParam('sortDir', sortParams[sortOption].dir)`, which always calls `history.push` and refreshes the component before it can ever set `sortField`. I think this can be fixed by only calling `history.push` if it's necessary in `setQueryParam`. I'll leave some code suggestions in `Scholarships.jsx` that should fix this.

src/pages/Scholarships.jsx
```diff
@@ -30,63 +28,14 @@ function ScholarshipsPage() {
...
-  }
 
-  const sortField = validField ? params.sortField : DEFAULT_SORT_FIELD;
-  const sortDir = validDir ? params.sortDir : DEFAULT_SORT_DIR;
-
-  const location = useLocation();
   const history = useHistory();
-
+  const location = useLocation();
   const params = queryString.parse(location.search, { parseNumbers: true });
```
josueetcom, line 34: Add a line beneath this

src/lib/QueryParams.js
```diff
@@ -1,6 +1,5 @@
-const qParams = {
+/* eslint-disable import/prefer-default-export */
+export const qParams = {
```
josueetcom, line 2:
```suggestion
export default const qParams = {
```

src/pages/Scholarships.jsx
```diff
@@ -96,10 +45,27 @@ function ScholarshipsPage() {
     });
   };
 
-  const pruneQueryParam = (index) => {
-    delete params[index];
-    history.replace({ search: queryString.stringify(params) });
-  };
+  const DEFAULT_SORT_FIELD = 'deadline';
+  const DEFAULT_SORT_DIR = 'asc';
```
josueetcom, line 49: These would be better at the top of the file after the imports

src/pages/Scholarships.jsx
```diff
@@ -24,18 +24,18 @@ const useStyles = makeStyles((theme) => ({
...
   const history = useHistory();
-
+  const location = useLocation();
   const params = queryString.parse(location.search, { parseNumbers: true });
-  const { minAmount, maxAmount } = params;
-
+  const pruneQueryParam = (index) => {
+    delete params[index];
+    history.replace({ search: queryString.stringify(params) });
+  };
```
josueetcom, line 38: I suggest moving this below `setQueryParam` to minimize the diff.

src/pages/Scholarships.jsx
```diff
@@ -45,10 +45,27 @@ function ScholarshipsPage() {
...
+  if (!validateField) {
+    pruneQueryParam('sortField');
+    sortField = DEFAULT_SORT_FIELD;
+  }
+
+  const validateDir = sortDir === 'asc' || sortDir === 'desc';
+  if (!validateDir) {
+    pruneQueryParam('sortDir');
+    sortDir = DEFAULT_SORT_DIR;
+  }
```
josueetcom, line 66: This is not exactly what I meant by my previous [comment](https://github.com/beyondhb1079/s4us/pull/379#discussion_r575939396). `pruneQueryParam` calls `history.replace()` which will refresh this component. So, as my [other comment](https://github.com/beyondhb1079/s4us/pull/379#discussion_r578960875) was trying to show, you can simplify this to:
```suggestion
  const sortField = params.sortField ?? DEFAULT_SORT_FIELD;
  const sortDir = params.sortDir ?? DEFAULT_SORT_DIR;
  
  // setSort function (if you still keep it)
  
  if (sortField !== 'deadline' && sortField !== 'amount.min' && sortField !== 'amount.max') {
    pruneQueryParam('sortField');  // prunes the query parameter, which ends up refreshing the component
  }
  // From this point on, sortField is valid.
  if (sortDir !== 'asc' && sortDir != 'desc') {
    pruneQueryParam('sortDir');  // prunes the query parameter, which ends up refreshing the component
  }
  // From this point on, sortDir is valid.
```

src/lib/QueryParams.js
gonzalo1330: Because I include another const `sortParams` I think its needed the way it is unless there is a better way to do it for exporting both const ?

src/components/FilterDropdown.jsx
```diff
@@ -16,15 +18,22 @@ const useStyles = makeStyles((theme) => ({
...
   const classes = useStyles();
+  const location = useLocation();
+  const params = queryString.parse(location.search, { parseNumbers: true });
+  const [choice, setChoice] = useState(
+    value === 'sorting'
+      ? params.sortBy === undefined
+        ? 'deadlineAsc'
+        : params.sortBy
+      : ''
+  );
```
gonzalo1330, line 31: lint fails because of nested ternary expression, any suggestions on how to go about this?

gonzalo1330: Nvm got it, I think can you check to ensure I did it correctly?

gonzalo1330:
Everything is working now expect when an invalid param value is set to `sortBy`. Such as this -> https://s4us-pr-379.onrender.com/scholarships?sortBy=amountDescc .. any suggestions on how to fix that?

gonzalo1330:
Nice, last edge case that is not fully functioning is the sort filter drop down onDefault selection. It works when sortBy has params such as https://s4us-pr-379.onrender.com/scholarships?sortBy=amountDesc the amount (high to low) is selected and works when visiting https://s4us-pr-379.onrender.com/scholarships then the default of Deadline (Earliest to Latest) is selected.
But, when you visit https://s4us-pr-379.onrender.com/scholarships?sortBy=amountDescc essentially when prune function is called then there is none selected when it should be using the default selection of Deadline (Earliest to Latest). I’ll try and figure it out by any suggestions are welcomed

src/lib/QueryParams.js
josueetcom: Ah, that's because in general it's discouraged to export more than one thing form a file. Hmm, in that case, how about defining `sortParams` at the top of the `FilterBar.jsx` file and exporting it from there? Also I would rename it `sortOptions` since it's not _strictly_ tied to query parameters. The `sortingOptions` dictionary there can be renamed `sortItems` since it's the items needed for the dropdown.

src/pages/Scholarships.jsx
```diff
@@ -11,7 +11,9 @@ import Button from '@material-ui/core/Button';
 import Scholarships from '../models/Scholarships';
 import ScholarshipList from '../components/ScholarshipList';
 import FilterBar from '../components/FilterBar';
-import qParams from '../lib/QueryParams';
+import { qParams, sortParams } from '../lib/QueryParams';
+
+const DEFAULT_SORT_BY = 'deadlineAsc';
```
josueetcom, line 16: Putting `sortParams`/`sortOptions` next to this line (maybe before) is probably an ideal place since they're related.

src/pages/Scholarships.jsx
```diff
@@ -50,6 +50,23 @@ function ScholarshipsPage() {
...
   };
 
+  const sortBy = params.sortBy ?? DEFAULT_SORT_BY;
+
+  if (
+    sortBy !== 'deadlineAsc' &&
+    sortBy !== 'deadlineDesc' &&
+    sortBy !== 'amountAsc' &&
+    sortBy !== 'amountDesc'
+  ) {
```
josueetcom, line 60: I suggest using [`hasOwnProperty`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) here.

src/pages/Scholarships.jsx
```diff
@@ -50,6 +50,23 @@ function ScholarshipsPage() {
...
+    pruneQueryParam('sortBy');
+  }
+
+  const key = params.sortBy !== undefined ? sortBy : DEFAULT_SORT_BY;
```
josueetcom, line 64: This variable isn't necessary, you already made sure `sortBy` is valid so just use `sortBy`. (You could rename it to `sortKey` or something more readable though if it helps)

src/components/FilterDropdown.jsx
```diff
@@ -16,15 +18,18 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 function FilterDropdown(props) {
-  const { label, items, defaultValue, removeNone, onChange } = props;
-  const [choice, setChoice] = useState(defaultValue);
+  const { label, items, value, removeNone, onChange } = props;
   const classes = useStyles();
+  const location = useLocation();
+  const params = queryString.parse(location.search, { parseNumbers: true });
```
josueetcom, line 24: This component shouldn't be parsing the query string as it could be used for anything, including cases where the query string is not used.

src/components/FilterDropdown.jsx
```diff
@@ -16,15 +18,18 @@ const useStyles = makeStyles((theme) => ({
...
+  const sortByParams =
+    params.sortBy === undefined ? 'deadlineAsc' : params.sortBy;
```
josueetcom, line 26: This component should not be concerning itself over sorting dropdown specific logic. This is a generic dropdown component that can used as a dropdown for **anything**.

src/components/FilterDropdown.jsx
```diff
@@ -16,15 +18,18 @@ const useStyles = makeStyles((theme) => ({
...
+  const [choice, setChoice] = useState(value === 'sorting' ? sortByParams : '');
```
josueetcom, line 27: I'm not understanding why this is necessary. If the parent component passes what `value` should be and updates it whenever it needs to be something else, why set the choice to anything else? Using `value` also removes the need for the `choice` state variable.

src/components/FilterBar.jsx
```diff
@@ -82,18 +65,16 @@ export default function FilterBar(props) {
         <FilterDropDown
           label="Sorting"
           items={sortingOptions}
-          defaultValue="deadlineSoon"
+          value="sorting"
```
josueetcom, line 68: Why is this fixed? The value is just whatever key of `sortingOptions` should be currently selected, e.g. `deadlineAsc`. You can determine the value from the `queryParams` object that was passed in.

josueetcom (changes requested):

src/pages/Scholarships.jsx
gonzalo1330: So removing it from the import and just placing it on this current file? The sortParams but in another comment you mentioned to change the name to sortOptions right. Ill get this change in

src/lib/QueryParams.js
gonzalo1330: Ok done

src/components/FilterDropdown.jsx
gonzalo1330: Ah ok I left it how it was when it merged from main, so should it be changed to `const params = queryString.parse(location.search);`?

src/pages/Scholarships.jsx
gonzalo1330: Well when its pruned is when the issue comes because now it no longer is valid meaning it is now `undefined`

gonzalo1330: If this is what you mean then it does not work
```javascript
  if (
    !sortBy.hasOwnProperty('deadlineAsc') &&
    !sortBy.hasOwnProperty('deadlineDesc') &&
    !sortBy.hasOwnProperty('amountAsc') &&
    !sortBy.hasOwnProperty('amountDesc')
  ) {
    pruneQueryParam('sortBy');
  }

  const sortField = sortOptions[sortBy].field;
  const sortDir = sortOptions[sortBy].dir;
```

src/pages/Scholarships.jsx
gonzalo1330: Look at comment below and see if that's what you mean

src/components/FilterDropdown.jsx
gonzalo1330: I think what you mean is just pass the `sortBy` from Scholarships and that will determine what is used. But again when prune is used it is undefined or something because it does not read in anything if u see

gonzalo1330: Ah ok nvm I figured it out, nice you were completely correct the choice state is no longer needed. Cool thank you

src/pages/Scholarships.jsx
```diff
@@ -13,6 +13,26 @@ import ScholarshipList from '../components/ScholarshipList';
 import FilterBar from '../components/FilterBar';
 import qParams from '../lib/QueryParams';
 
+const sortOptions = {
+  deadlineAsc: {
```
josueetcom, line 17: I forgot to mention, I think you can keys like `deadline.asc` if it's in quotes, e.g. `'deadline.asc'`. I imagine you chose this route because it wasn't letting you with the dot (since there is no variable called `deadline`) but if you surround it with quotes it should work.

src/components/FilterDropdown.jsx
```diff
@@ -20,20 +18,15 @@ const useStyles = makeStyles((theme) => ({
...
-    setChoice(event.target.value);
     onChange(event.target.value);
   }
+
   return (
     <FormControl variant="outlined" className={classes.formControl}>
       <Select
-        value={choice}
+        value={value}
         onChange={selectChoice}
```
josueetcom, line 30: You can get rid of the `selectChoice` function now.
```suggestion
        onChange={(e) => onChange(e.target.value)}
```

src/pages/Scholarships.jsx
josueetcom: Let me ask you two questions. What is the type of `sortBy`? And then, what kind of object should `hasOwnProperty` be used against? Does it make any sense to check if a string has a property/key of a string?

One of the reasons I strongly dislike JavaScript is because it allows to write things like `"someString".hasOwnProperty("foo")` and it's still valid JavaScript.

Think about it a little more, is `sortBy` the dictionary you're checking in or the key you're checking for?

src/components/FilterDropdown.jsx
josueetcom: FilterDropdown should not be concerning itself over query parameters. It's just a visual component, it should not care what the URL is.

src/pages/Scholarships.jsx
```diff
@@ -53,17 +71,16 @@ function ScholarshipsPage() {
...
     sortBy !== 'amountDesc'
   ) {
     pruneQueryParam('sortBy');
   }
 
-  const key = params.sortBy !== undefined ? sortBy : DEFAULT_SORT_BY;
-  const sortField = sortParams[key].field;
-  const sortDir = sortParams[key].dir;
+  const sortField = sortOptions[params.sortBy ?? DEFAULT_SORT_BY].field;
+  const sortDir = sortOptions[params.sortBy ?? DEFAULT_SORT_BY].dir;
```
josueetcom, line 83: Again, `sortBy` is already equal to `params.sortBy ?? DEFAULT_SORT_BY`. There's no need to repeat the expression.

gonzalo1330: Ok

src/components/FilterDropdown.jsx
gonzalo1330: Ok nice

src/pages/Scholarships.jsx
gonzalo1330: Oh ok yea that was way I used the uppercase letter on the `dir`, but thats good to know. I did go the route of using the dot to separate the `field` and `dir` using the way you mentioned.

src/components/FilterDropdown.jsx
gonzalo1330: Ok I see what you mean, I believe the current` FilterDropdown.jsx `file includes the `const params `anymore.

gonzalo1330: Ok makes sense then I think the code below is what you mean
```javascript
if (!sortOptions.hasOwnProperty('deadline.asc')) {
    pruneQueryParam('sortBy');
}
```

src/pages/Scholarships.jsx
gonzalo1330: Ok makes sense then I think the code below is what you mean
```javascript
if (!sortOptions.hasOwnProperty('deadline.asc')) {
    pruneQueryParam('sortBy');
}
```

gonzalo1330: Its causing lint issues so should it be something like this?
if (!(sortBy in sortOptions)) {
    pruneQueryParam('sortBy');
}

src/components/FilterBar.jsx
```diff
@@ -30,38 +30,40 @@ const grades = {
...
-  deadlineSoon: 'Deadline (Earliest to Latest)',
-  deadlineLatest: 'Deadline (Latest to Earliest)',
-  amountLow: 'Amount (Low to High)',
-  amountHigh: 'Amount (High to Low)',
+const sortItems = {
+  'deadline.asc': 'Deadline (Earliest to Latest)',
+  'deadline.desc': 'Deadline (Latest to Earliest)',
+  'amount.asc': 'Amount (Low to High)',
+  'amount.desc': 'Amount (High to Low)',
+};
```
josueetcom, line 38: Not sure if you wanted to move the descriptions into `sortOptions` and then define `sortItems` using [Object.fromEntries](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries) like in the [stack overflow example](https://stackoverflow.com/a/14810722) we found earlier.

Feel free to punt this change to a separate PR though

src/components/FilterBar.jsx
```diff
@@ -30,38 +30,40 @@ const grades = {
...
-        changeSortBy('amount.min');
-        changeSortFormat('asc');
-        break;
-      case 'amountHigh':
-        changeSortBy('amount.max');
-        changeSortFormat('desc');
-        break;
-      default:
+  function updateSortingParams(sortOption) {
+    if (sortOption in sortOptions) {
```
josueetcom, line 65: Per comment above, this check would no longer be needed if those changes are made.

josueetcom (approved):

src/pages/Scholarships.jsx
josueetcom: Makes sense, thanks for proactively addressing the lint issue!

PR #380, opened 2021-02-08
grade level filter component

### LINKED ISSUE #273: Scholarship list: grade level filter
**User Story**
_As a user, I can filter the scholarship list by amount._

- [x] #270
- [x] #274
- [x] #275
- [x] #465
- [x] #265

### LINKED ISSUE #465: Grade level filter: basic component

gonzalo1330:
Dropdown for user to select grade filter.

## Motivation and Context

A user will be able to filter scholarships based on 1 or more grade level. 

Fixes #465 

## How Has This Been Tested?

Currently selection of grades have not been included but will test and update once the selection is available.

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-380.onrender.com.
2. Go to the scholarships link that is in the header.
3. On the top left there will be a selection for `Grade` that will be selectable of 1 or more.

## Checklist:


package.json
```diff
@@ -21,6 +21,7 @@
     "eslint-plugin-jest": "^24.0.2",
     "firebase": "^7.19.1",
     "formik": "^2.2.6",
+    "multiselect-react-dropdown": "^1.6.7",
```
josueetcom, line 24: This looks unused. `yarn remove`?

gonzalo1330: Sure it has been removed

src/components/GradeLevelFilter.jsx
```diff
@@ -0,0 +1,72 @@
...
+import ArrowDropDownIcon from '@material-ui/icons/ArrowDropDown';
+
+const useStyles = makeStyles((theme) => ({
+  buttonStyle: {
+    margin: theme.spacing(1),
+    height: theme.spacing(4),
+  },
+  popOverWindow: {
+    minWidth: '200px',
+  },
```
gonzalo1330, line 13: Styling currently not working as intended but will figure this out as on the console window when I change the min-width it works but does not work this way

src/components/GradeLevelFilter.jsx
```diff
@@ -0,0 +1,72 @@
...
+        transformOrigin={{ vertical: 'top', horizontal: 'left' }}>
+        <input
+          key="12"
+          type="checkbox"
+          // eslint-disable-next-line no-return-assign
+          onClick={() => (gradeList[12].isChecked = !gradeList[12].isChecked)}
+          checked={gradeList[12].isChecked}
+          value={gradeList[12].value}
+        />
+        {gradeList[12].value}
```
gonzalo1330, line 68: Im thinking this would be iterated through the gradeList object and then for each then there would be an input applied, is what im thinking at the moment to get the checkbox to be rendered for all the grades in the gradeList

src/components/GradeLevelFilter.jsx
```diff
@@ -0,0 +1,72 @@
...
+        open={Boolean(anchorEl)}
+        anchorEl={anchorEl}
+        onClose={() => setAnchorEl(null)}
+        anchorOrigin={{ vertical: 'bottom', horizontal: 'left' }}
+        transformOrigin={{ vertical: 'top', horizontal: 'left' }}>
+        <input
+          key="12"
+          type="checkbox"
+          // eslint-disable-next-line no-return-assign
+          onClick={() => (gradeList[12].isChecked = !gradeList[12].isChecked)}
```
gonzalo1330, line 64: Line 64, the change for the `isChecked` for the gradeList, does change but does not get rendered updated until the `popOver` is closed

src/components/GradeLevelFilter.jsx
```diff
@@ -0,0 +1,81 @@
...
+  },
+};
+
+export default function GradeLevelFilter() {
+  const classes = useStyles();
+  const [anchorEl, setAnchorEl] = useState(null);
+  const handleClick = (event) => setAnchorEl(event.currentTarget);
+
+  function updateGradeSelection(value) {
+    gradeList[value].isChecked = !gradeList[value].isChecked;
```
josueetcom, line 45: This is potentially unsafe as this is not a true constant variable. It's being modified by potentially multiple instances of `GradeLevelFilter`.

Instead of this I recommend using [local state](https://reactjs.org/docs/state-and-lifecycle.html#adding-local-state-to-a-class). Perhaps using `useState`.

src/components/GradeLevelFilter.jsx
```diff
@@ -0,0 +1,81 @@
...
+  },
+  15: {
+    value: 'College Junior',
+    isChecked: false,
+  },
+  16: {
+    value: 'College Senior',
+    isChecked: false,
+  },
+};
```
josueetcom, line 37: Instead of this you could map the string to a boolean. E.g.

```js
const gradeList = {
  'Highschool Senior': false,
  // etc.
};
```

Or even better, since the values are just booleans, you could use a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/Set) to track checked options.

```js
const selected = useState(Set());

// use selected.add() and selected.remove() on selecting/deselecting a checkbox.
```

src/components/GradeLevelFilter.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+export default function GradeLevelFilter() {
+  const classes = useStyles();
+  const [anchorEl, setAnchorEl] = useState(null);
+  const handleClick = (event) => setAnchorEl(event.currentTarget);
+
+  function updateGradeSelection(value) {
+    gradeList[value].isChecked = !gradeList[value].isChecked;
+    setAnchorEl(false);
```
josueetcom, line 46: This closes the popover, which might be annoying if people want to select more than one check box. Remove?

src/components/GradeLevelFilter.jsx
josueetcom: Although I think I see now maybe why this was done. Maybe in anticipation of URL parameters where encoding numbers might make more sense. That being the case, I think this PR can stay focused on just the styling if you prefer. Maybe we should discuss with the others what might make sense to encode in the URL for grade level selections.

josueetcom (commented):
Initial feedback

src/components/GradeLevelFilter.jsx
gonzalo1330: Sure, I did this for two reasons.
1. It took care of the rendering of check box not occur until the popover closes &
2. Technically a user would always need to only check 1 box for their specific grade, then if they are for example parents searching for multiple students then they would just go back and select the other grade
For now ill remove and we can figure out if it should be closed after a selection or kept open until user decided to close popover

src/components/GradeLevelFilter.jsx
gonzalo1330: Ok I can stay focused on the styling and then once we figure out how to encode the parameters for grades I can go ahead and figure out how to do this using a `Set`. Basically as a box is checked the onClick function would handle the adding of the grade selected and when the grade is already selected it would just search the set if found then just remove it, there would be no need for boolean, correct?

src/components/GradeLevelFilter.jsx
gonzalo1330: Hm, if we decide to go with a `set` then local state would not longer be needed since in the updateGradeSelection function would do something as follows:
```javascript
function updateGradeSelection(value) {
  if (gradeSet.has(value)) { // case when the grade selected is already in the set should be removed
    gradeSet.delete(value);
    return;
  }
  // happy case when the value just gets added to the set
  gradeSet.add(value); 
}
```

josueetcom: You need to consider the case where `GradeLevelFilter` is used more than once. There's no guarantee of that. `GradeLevelFilter` could be re-rendered, which creates another instance and another callback and if multiple callbacks access the same global variable and tamper with it they could mess up the state.

src/components/GradeLevelFilter.jsx
josueetcom: > once we figure out how to encode the parameters for grades I can go ahead and figure out how to do this using a Set

Not what I meant. I meant the map style you have could be necessary when encoding the selections via URL parameters. For styling/prototyping purposes a set is perfectly fine.

> Basically as a box is checked the onClick function would handle the adding of the grade selected and when the grade is already selected it would just search the set if found then just remove it, there would be no need for boolean, correct?

Right, clicking just toggles the presence of the selection.

src/components/GradeLevelFilter.jsx
josueetcom: > 1. It took care of the rendering of check box not occur until the popover closes &

This is why you use add [state](https://reactjs.org/docs/state-and-lifecycle.html) to the component. When the state changes the component re-renders.

2. Technically a user would always need to only check 1 box for their specific grade, then if they are for example parents searching for multiple students then they would just go back and select the other grade

That's reasonable. Though in that case I think we can just provide an experience where if they click anywhere outside the Popover it closes (which makes things easier if they apply multiple filters)

src/components/GradeLevelFilter.jsx
```diff
@@ -10,40 +10,38 @@ const useStyles = makeStyles((theme) => ({
...
-    setAnchorEl(false);
+  function updateGradeSelection(gradeSelected) {
+    const newSet = new Set(selectedGrades);
+    if (selectedGrades.has(gradeSelected)) {
+      newSet.delete(gradeSelected);
+      setSelectedGrades(newSet);
+      return;
+    }
+    newSet.add(gradeSelected);
+    setSelectedGrades(newSet);
```
josueetcom, line 44:
```suggestion
    if (selectedGrades.has(gradeSelected)) {
      newSet.delete(gradeSelected);
    } else {
      newSet.add(gradeSelected);
    }
    setSelectedGrades(newSet);
```

I consider the two flows to both be part of the main flow. The single if statement is usually used for error cases or for base cases where most of the work is done in the main flow.

src/components/GradeLevelFilter.jsx
```diff
@@ -56,24 +54,23 @@ export default function GradeLevelFilter() {
...
       <Popover
-        className={classes.popOverWindow}
         open={Boolean(anchorEl)}
         anchorEl={anchorEl}
         onClose={() => setAnchorEl(null)}
         anchorOrigin={{ vertical: 'bottom', horizontal: 'left' }}
         transformOrigin={{ vertical: 'top', horizontal: 'left' }}>
-        {Object.keys(gradeList).map((key) => (
-          <>
+        {Object.keys(gradeItems).map((key) => (
```
josueetcom, line 62: Hmm, I'm not entirely sure `Object.keys()` works with sets. It doesn't seem to work when I type this in the Web inspector JavaScript console:

```js
s = new Set(["bar", "foo"]);
Object.keys(s);  // outputs []
```

Another way you could write this is:
```suggestion
        {Array.from(gradeItems).map((grade) => (
```
And use `grade` for the key/value

src/components/GradeLevelFilter.jsx
```diff
@@ -10,40 +10,38 @@ const useStyles = makeStyles((theme) => ({
...
 export default function GradeLevelFilter() {
   const classes = useStyles();
   const [anchorEl, setAnchorEl] = useState(null);
   const handleClick = (event) => setAnchorEl(event.currentTarget);
+  const [selectedGrades, setSelectedGrades] = useState(new Set());
 
-  function updateGradeSelection(value) {
-    gradeList[value].isChecked = !gradeList[value].isChecked;
-    setAnchorEl(false);
+  function updateGradeSelection(gradeSelected) {
```
josueetcom, line 36: Nits on naming:
- `update` is a bit generic for what could more precisely be described as [toggling](https://www.google.com/search?q=define+toggle)
- The `Grade` in `updateGradeSelection` might be a bit too specific considering how the component is named `GradeLevelFilter` and there are no other functions here on toggling selection.
- The `Selected` in `gradeSelected` is a bit redundant considering the name of this function.

You might consider renaming this function something like `toggleSelection`, `toggleSelected`, `toggleSelectedState`, etc. and the parameter `grade`.

src/components/GradeLevelFilter.jsx
```diff
@@ -0,0 +1,78 @@
...
+  'College Junior',
+  'College Senior',
+  '5th Year',
+  'Post Grad',
+];
+
+export default function GradeLevelFilter() {
+  const classes = useStyles();
+  const [anchorEl, setAnchorEl] = useState(null);
+  const handleClick = (event) => setAnchorEl(event.currentTarget);
```
josueetcom, line 33: `handleClick` is a bit generic. I can't guess based on the name what it's supposed to do. Maybe `openPopover`?

src/components/FilterBar.jsx
```diff
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { makeStyles, Grid } from '@material-ui/core';
 import FilterDropDown from './FilterDropdown';
 import AmountFilter from './AmountFilter';
+import GradeFilter from './GradeLevelFilter';
```
josueetcom, line 6: Naming in inconsistent here. Either update the grade filter import name or the file name

src/components/GradeLevelFilter.jsx
```diff
@@ -0,0 +1,78 @@
...
+            <input
+              key={key}
+              type="checkbox"
+              onClick={() => updateGradeSelection(gradeItems[key])}
+              checked={selectedGrades.has(gradeItems[key])}
+              value={gradeItems[key]}
+            />
+            {gradeItems[key]}
+            <br />
+          </div>
```
josueetcom, line 73: Use [`Checkbox`](https://material-ui.com/components/checkboxes/) with [`FormControlLabel`](https://material-ui.com/components/checkboxes/#checkbox-with-formcontrollabel) and surround them all with a [`FormGroup`](https://material-ui.com/components/checkboxes/#checkboxes-with-formgroup) instead.

src/components/GradeLevelFilter.jsx
```diff
@@ -0,0 +1,78 @@
...
+        Grade
+      </Button>
+      <Popover
+        open={Boolean(anchorEl)}
+        anchorEl={anchorEl}
+        onClose={() => setAnchorEl(null)}
+        anchorOrigin={{ vertical: 'bottom', horizontal: 'left' }}
+        transformOrigin={{ vertical: 'top', horizontal: 'left' }}>
+        {Object.keys(gradeItems).map((key) => (
+          <div className={classes.popOverWindow}>
```
josueetcom, line 63: Hmm, I understand where this is coming from but the styling probably ought to be on `Popover` itself, or on the `FormGroup`/`FromControlLabel` once that's added.

josueetcom (changes requested):

src/components/GradeLevelFilter.jsx
```diff
@@ -2,77 +2,47 @@ import React, { useState } from 'react';
...
+        className={classes.popOverWindow}
+        multiple
+        limitTags={1}
+        options={gradeOptions}
+        getOptionLabel={(option) => option.title}
+        renderInput={(params) => (
+          <TextField
+            {...params}
+            variant="outlined"
+            label=" Grade"
```
josueetcom, line 41:
```suggestion
            label="Grade"
```

src/components/GradeLevelFilter.jsx
```diff
@@ -2,77 +2,47 @@ import React, { useState } from 'react';
...
-            />
-            {gradeItems[key]}
-            <br />
-          </div>
-        ))}
-      </Popover>
+      <Autocomplete
+        className={classes.popOverWindow}
+        multiple
+        limitTags={1}
```
josueetcom, line 34: Hmm, I'm not entirely fond of how the rest of the page shifts down as you select more options but then shifts back up once you click outside of the dropdown. Also the autocomplete is a bit wonky to use on mobile :P.

josueetcom: Considering that the options for selecting grade level are limited and easily groupable, can we just leave this as a Popover with Checkboxes? Be sure to address the comments from my previous review about following the [Checkbox](https://material-ui.com/components/checkboxes/) Material UI examples.

josueetcom: I think the Autocomplete we were talking about in the Figma design refers to the **major** filter, not the grade filter. The grade filter is just a popover with checkboxes.

josueetcom (changes requested):

src/components/GradeLevelFilter.jsx
gonzalo1330: Ah ok sounds good I changed it to the` if/else`

src/components/GradeLevelFilter.jsx
gonzalo1330: Hmm ok yea I see, made the changes

src/components/GradeLevelFilter.jsx
gonzalo1330: Ok yea does make sense, I do need to better with my naming conventions.

src/components/GradeLevelFilter.jsx
gonzalo1330: Ok `openPopover` it is

src/components/FilterBar.jsx
gonzalo1330: ok I changed the import name

src/components/GradeLevelFilter.jsx
gonzalo1330: Yeah ok we decided not to go with the auto complete so I just went back to the previous commit with the right changes.

src/components/GradeLevelFilter.jsx
gonzalo1330: I don't think this is on there anymore

src/components/GradeLevelFilter.jsx
gonzalo1330: Ok yea ill make this change once I get the `FormGroup` added

src/components/GradeLevelFilter.jsx
gonzalo1330: Ok im going to try that

gonzalo1330: Alright got it to work with the way you recommended

src/components/GradeLevelFilter.jsx
gonzalo1330: This is still pending for styling ill do that on the next push I do

.gitignore
```diff
@@ -1,5 +1,8 @@
 # See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
+#VSCode
```
josueetcom, line 3:
```suggestion
# VSCode
```

josueetcom (dismissed):

josueetcom (approved):

PR #374, opened 2021-02-07
Refactor dependency installation job in Validation workflow

josueetcom:
I just learned you can specify job order dependency in GitHub workflow actions. Since 3 of the validation checks rely on installing dependencies I've refactored that into a separate job.

## Motivation and Context

Each of the checks in our [validation workflow actions](https://github.com/beyondhb1079/s4us/actions?query=workflow%3AValidation) spends around a minute getting the yarn cache and installing dependencies from it.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] User visible change (users will notice UI or functional changes)

## Previewing Changes

1. Go to https://s4us-pr-NUMBER.onrender.com.
2.

## Checklist:

- [ ] I have added tests to cover my changes.

PR #383, opened 2021-02-12
Remove testing from pre-push hook

josueetcom:
## Motivation and Context

Now that #374 has sped up validation check tests significantly I think we can remove this overhead from pushing. We should really be pushing often and the 1.5-2.5 minutes this takes discourages you from pushing often. I had introduced it to catch things sooner but it's probably causing more toil than helping, especially now that it takes less than a minute to get feedback form the validation check.

Note: I've kept linting in the pre-push as that only takes around 30 seconds and it'll catch some things the build check will complain about.

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,7 +152,7 @@
   "husky": {
     "hooks": {
       "pre-commit": "lint-staged",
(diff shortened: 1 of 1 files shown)


Jobhdez (approved):
Looks good!

PR #412, opened 2021-03-05
pagination functionality

### LINKED ISSUE #94: Pagination: functionality

edwingl10:
## Motivation and Context

Fixes #94 

## How Has This Been Tested?
On laptop using Firefox.

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-412.onrender.com/scholarships.
1. Scroll down and click **Load More**.

## Checklist:


src/models/Scholarships.ts
```diff
@@ -46,11 +46,15 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
    *
    * @param opts list options.
    */
-  list(opts: {
-    sortDir?: SortDirection;
-    sortField?: string;
-    authorId?: string;
-  }): Promise<FirestoreModel<ScholarshipData>[]> {
+  async list(
```
josueetcom, line 49: No need to do `async`, this already returns a promise. The way you handle a promise that handles a promise is:

```js
return doSomething
  .then((result) => {/* do something with the result */});
```

See the [Chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises#chaining) guide Promises

src/models/Scholarships.ts
```diff
@@ -65,6 +69,12 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
     if (opts.authorId) {
       query = query.where('authorId', '==', opts.authorId);
     }
+
+    query = query.limit(5).startAfter(lastDoc || 0);
+
+    const snapShot = await query.get();
+    const last = snapShot.docs[snapShot.docs.length - 1];
+    setLastDoc(last);
     return FirestoreCollection.list(query);
```
josueetcom, line 78: Maybe what we can do is return some sort of response that has the results as well as a function to get the next page of results. E.g. 

```js
return FirestoreCollection.list(query)
  .then((scholarships) => { scholarships, next: () => Scholarships.list(opts, lastDoc)})
```

src/pages/Scholarships.jsx
```diff
@@ -48,6 +48,7 @@ function ScholarshipsPage() {
   const [scholarships, setScholarships] = useState([]);
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState();
+  const [lastDoc, setLastDoc] = useState(null);
```
josueetcom, line 51: Instead of a state variable I suggest just deriving this from `scholarships`' last item.

src/pages/Scholarships.jsx
```diff
@@ -125,9 +133,7 @@ function ScholarshipsPage() {
             <Button
               className={classes.loadMoreButton}
               color="primary"
-              onClick={() => {
-                alert('clicked');
-              }}>
+              onClick={() => displayScholarships()}>
```
josueetcom, line 136: More succinctly
```jsx
              onClick={displayScholarships}>
```

gonzalo1330:
The previous is looking nice

src/models/Scholarships.ts
```diff
@@ -72,10 +71,20 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
...
+    return FirestoreCollection.list(query).then((results) => {
+      let lastDocument: any;
+      query.get().then((snapshot) => {
+        lastDocument = snapshot.docs[snapshot.docs.length - 1];
+      });
+      const next = () => {
+        return this.list(opts, lastDocument);
+      };
+
+      return { results, next };
```
josueetcom, line 86: This is a handler for what to do after listing completes and you have the results. You shouldn't need to do anything with `query.get()` because `results` already contains the results. I think what you want here is this:

```js
return {results, next: results[results.length - 1].id};
```

src/models/Scholarships.ts
```diff
@@ -72,10 +71,20 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
...
+    });
   }
```
josueetcom, line 88: This is a handler for what to do after listing completes and you have the results. You shouldn't need to do anything with `query.get()` because `results` already contains the results. I think what you want here is this:

```js
    return FirestoreCollection.list(query).then((results) => {
      return {results, next: results[results.length - 1].id};
    });
  }
```
or more succinctly:
```js
    return FirestoreCollection.list(query).then((res) => ({res, next: res[res.length - 1].id}));
  }
```

josueetcom: Though, looking at the docs for [startAfter](https://firebase.google.com/docs/reference/js/firebase.firestore.Query#startafter) it looks like this is something that should probably be done in FirestoreCollection instead as the results from `FirestoreCollection.list(query)` are already [transformed](https://github.com/beyondhb1079/s4us/blob/main/src/models/base/FirestoreCollection.ts#L29) from `DocumentSnapshot`.

edwingl10:
Trying to figure out why it breaks when changing the `Sort by` option

src/pages/Scholarships.jsx
```diff
@@ -93,29 +94,40 @@ function ScholarshipsPage() {
...
   }
 
+  const displayScholarships = useCallback(
+    (scholarshipsList) => {
+      let mounted = true;
+      scholarshipsList
+        .then(({ results, next }) => {
+          setNextDocument(next);
+          return (
+            mounted &&
```
josueetcom, line 104: The `mounted` check should be done before calling `setNextDocument`.

gonzalo1330:
Hm discovered a bug while try the render, when all of the scholarships have been fetched and then the `Load More` is clicked again then it disappears all the scholarships from the page.
One other thing to note is that when the load more button is clicked it does not keep the user at the last recent shown prior to the one that load after the click it takes the page all the way down to the bottom again near the `Load More` button

src/models/base/FirestoreCollection.ts
```diff
@@ -20,14 +21,21 @@ export default abstract class FirestoreCollection<T> {
...
+
+    return firestoreQuery.get().then((qSnap: firestore.QuerySnapshot<E>) => {
+      return {
+        next: () => this.list(query, qSnap.docs[qSnap.docs.length - 1]),
+        results: qSnap.docs.map(
           (doc) => new FirestoreModel<E>(doc.ref, doc.data())
-        )
-      );
+        ),
+        empty: qSnap.empty,
```
josueetcom, line 37: You should try conditionally setting `next` (or instead of `empty`, `hasNext`) based on the amount of results you get vs. requested. E.g. you could do:

```ts
        next: () => this.list(query, qSnap.docs[qSnap.docs.length - 1]),
        results: qSnap.docs.map(
          (doc) => new FirestoreModel<E>(doc.ref, doc.data())
        ),
        hasNext: qSnap.empty || qSnap.size < limit,
```

or
 
```ts
        next: () => qSnap.empty ? null : this.list(query, qSnap.docs[qSnap.docs.length - 1]),
        results: qSnap.docs.map(
          (doc) => new FirestoreModel<E>(doc.ref, doc.data())
        ),
        empty: qSnap.empty,
```

I think the first version probably looks nicer so you don't have to deal with null `next` values.

src/pages/Scholarships.jsx
```diff
@@ -48,13 +48,15 @@ function ScholarshipsPage() {
   const [scholarships, setScholarships] = useState([]);
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState();
+  const [disableMoreBtn, setDisableMoreBtn] = useState(false);
```
josueetcom, line 51: Maybe `canLoadMore`, `hasMore` or `hasNext`? (Assuming other suggested renames/refactors are made)

src/pages/Scholarships.jsx
```diff
@@ -93,29 +95,45 @@ function ScholarshipsPage() {
...
+          if (mounted) {
+            setScholarships((prev) => [
+              ...prev,
+              ...results.filter((s) =>
+                s.data.amount.intersectsRange(minAmount, maxAmount)
+              ),
+            ]);
+          }
+          setLoadMoreFn(next);
+          setDisableMoreBtn(empty);
```
josueetcom, line 116: In case the component is no longer mounted everything should be inside the `if (mounted)` block. Alternatively instead of surrounding everything with `if (mounted)` you could also just short circuit return on `!mounted`:
```js
          if (!mounted) return;
          setScholarships((prev) => [
            ...prev,
            ...results.filter((s) =>
              s.data.amount.intersectsRange(minAmount, maxAmount)
            ),
          ]);
          setLoadMoreFn(next);
          setDisableMoreBtn(empty);
```

src/models/base/FirestoreCollection.ts
```diff
@@ -20,14 +21,21 @@ export default abstract class FirestoreCollection<T> {
...
-        querySnapshot.docs.map(
+    query: firestore.Query<E>,
+    lastDocument?: firestore.QueryDocumentSnapshot<E>
+  ): Promise<ScholarshipCollectionList<E>> {
+    let firestoreQuery: firestore.Query<E> = query.limit(5);
+    if (lastDocument !== undefined)
+      firestoreQuery = firestoreQuery.startAfter(lastDocument);
+
+    return firestoreQuery.get().then((qSnap: firestore.QuerySnapshot<E>) => {
+      return {
```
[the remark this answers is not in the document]
edwingl10: I was testing more things here and forgot to take it out when cleaning up.

src/models/base/FirestoreCollection.ts
edwingl10: Nice I went with the first option.

src/pages/Scholarships.jsx
```diff
@@ -148,9 +149,9 @@ function ScholarshipsPage() {
             <Button
               className={classes.loadMoreButton}
               color="primary"
-              disabled={disableMoreBtn}
+              disabled={canLoadMore}
               onClick={() => loadMoreScholarships(loadMoreFn)}>
-              Load More
+              {canLoadMore ? 'No more results' : 'Load More'}
```
josueetcom, line 154: This still shows the button. I'd rather it be something like:
```jsx
{canLoadMore ? <Button>... : 'No more results'}
```

Here and in the other page.

josueetcom (dismissed):
Just left a suggestion on not showing the load more button when there are no results to load. Otherwise looks good to me

src/pages/Scholarships.jsx
```diff
@@ -146,13 +146,17 @@ function ScholarshipsPage() {
...
-              color="primary"
-              disabled={canLoadMore}
-              onClick={() => loadMoreScholarships(loadMoreFn)}>
-              {canLoadMore ? 'No more results' : 'Load More'}
-            </Button>
+            {canLoadMore ? (
+              <Button
+                className={classes.loadMoreButton}
+                color="primary"
+                disabled={!canLoadMore}
```
josueetcom, line 153: This can be deleted

src/pages/UserHome.jsx
```diff
@@ -161,13 +161,17 @@ export default function UserHome() {
...
-            color="primary"
-            disabled={canLoadMore}
-            onClick={() => loadMoreScholarships(loadMoreFn)}>
-            {canLoadMore ? 'No more results' : 'Load More'}
-          </Button>
+          {canLoadMore ? (
+            <Button
+              className={classes.loadMoreButton}
+              color="primary"
+              disabled={!canLoadMore}
```
josueetcom, line 168: Same here.

josueetcom (approved):

PR #414, opened 2021-03-07
Fix promise rejections and simplify promises

### LINKED ISSUE #411: Fix UnhandledPromiseRejectionWarning warnings in tests
**Describe the bug**
If you look at our [validation checks](https://github.com/beyondhb1079/s4us/actions?query=workflow%3AValidation) the test logs are littered with UnhandledPromiseRejectionWarning and using `act()` where necessary. We ought to fix these issues. These seem to be blocking us from [upgrading react-scripts](https://github.com/beyondhb1079/s4us/pull/410), which would fix a current audit vulnerability we have.

josueetcom:
## Motivation and Context

Fixes #411. Sometimes we get warnings in the `yarn test` logs because of unhandled promises. This seems to mainly from the fact that in the test setup/teardown steps we wrote `async` but never `await`-ed for those calls to succeed. It turns out though that there, as well as in tests, instead of `await`-ing we can simply return the promise to wait on and Jest will do the rest.

Also, as part of this change I simplified some promises a bit as I just learned you can transform a promise instead of having to wrap it around a promise.

## How Has This Been Tested?

Using `yarn tests`. 

## Types of changes

## Previewing Changes

1. Go to the **Checks** tab in this PR.
2. Click on the **Validation** workflow.
3. Select the **test** job.
4. Expand the **Run tests** step.
5. Look for **UnhandledPromiseRejection**

## Checklist:


src/App.test.jsx
```diff
@@ -2,8 +2,13 @@ import React from 'react';
...
+// TODO: Figure out a cleaner solution..
+window.MutationObserver = require('mutation-observer');
+
 test('renders home page by default', () => {
   render(<App />);
-  const linkElement = screen.getByText(/Scholarships/i);
-  expect(linkElement).toBeInTheDocument();
+
+  const linkElement = screen.findByText(/Scholarships/i);
+  return expect(linkElement).resolves.toBeInTheDocument();
```
josueetcom, line 13: So this file's changes are a bit unintuitive. Basically what's going on is that the home page gets updated in `useEffect` due to an auth state change so I've returned a promise here to wait on even though it's not quite related to that effect :/.

src/components/Header.test.jsx
```diff
@@ -1,26 +1,32 @@
...
+// hacky workaround to allow findBy to work
+// TODO: Figure out a cleaner solution.
+window.MutationObserver = require('mutation-observer');
+
+test('does not render alert by default', async () => {
   delete window.location;
   window.location = new URL('https://www.example.com');
 
   render(<Header />, { wrapper: MemoryRouter });
+  await screen.findByText('Login');
```
josueetcom, line 18: This is to wait for the auth state to resolve itself and the component to reach its final render state. Without this we get warnings complaining about not handling promise rejections.

src/models/base/FirestoreCollection.test.ts
```diff
@@ -19,10 +19,8 @@ const names = new (class extends FirestoreCollection<NameData> {
   converter = converter;
 })();
 
-beforeEach(async () =>
-  clearFirestoreData(app.options as { projectId: string })
-);
-afterAll(async () => app.delete());
```
josueetcom, line 25: The issue before with lines like this is that something that creates a promise would be called but it'd never be waited on. So we need these functions to either be non-async so they can handle the promise themselves or `await` the promise in this `async` function.

edwingl10 (approved):

Jobhdez (approved):

Jobhdez (approved):

PR #463, opened 2021-04-20
Update autoapprove dependabot action to use github-script@v3

josueetcom:
This action has been failing lately:
https://github.com/beyondhb1079/s4us/actions/workflows/dependabot-autoapprove.yml

I think it might be because we're using an older version of github-script.


gonzalo1330 (approved):
Nice

PR #484, opened 2021-05-18
initial test

### LINKED ISSUE #394: Component unit tests: ProtectedRoute
We need unit tests for the [ProtectedRoute](https://github.com/beyondhb1079/s4us/blob/main/src/components/ProtectedRoute.jsx). Consider any edge cases. See the [Testing Library docs](https://testing-library.com/docs/) and existing tests for reference.

edwingl10:
## Motivation and Context

Fixes #394 

## How Has This Been Tested?

On laptop using Firefox. 

## Types of changes

## Previewing Changes

2.

## Checklist:

### DIFF
--- a/src/components/ProtectedRoute.test.jsx
+++ b/src/components/ProtectedRoute.test.jsx
@@ -1,0 +1,40 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { MemoryRouter, Route } from 'react-router-dom';
+import { clearFirestoreData, initializeTestApp } from '../lib/testing';
+import ScholarshipForm from '../pages/ScholarshipForm';
+
+window.MutationObserver = require('mutation-observer');
+
+function renderAtRoute(route) {
+  return render(
+    <MemoryRouter initialEntries={[route]}>
+      <Route path="/scholarships/new" component={ScholarshipForm} />
+    </MemoryRouter>
+  );
+}
+
+const app = initializeTestApp({
+  projectId: 'protected-route-test',
+  apiKey: 'fake-api-key',
+});
+
+beforeAll(() => clearFirestoreData(app.options));
+afterAll(() => app.delete());
+
+test('athenticated user can accesss page', () => {
+  renderAtRoute('/scholarships/new');
+
+  expect(screen.getByText(/Submit a Scholarship/i)).toBeInTheDocument();
+  expect(screen.getByLabelText('Scholarship Name *')).toBeInTheDocument();
+  expect(screen.getByLabelText('Deadline *')).toBeInTheDocument();
+  expect(screen.getByLabelText('Description *')).toBeInTheDocument();
+  expect(screen.getByLabelText('Website *')).toBeInTheDocument();
+
+  expect(screen.getByText('Amount Type *')).toBeInTheDocument();
+  expect(screen.getByText('Unknown')).toBeInTheDocument();
+  expect(screen.getByText('Fixed Amount:')).toBeInTheDocument();
+  // expect(screen.getByText('Range:')).toBeInTheDocument();
+  expect(screen.getByText('Full Tuition')).toBeInTheDocument();
+  expect(screen.getByRole('button', { name: /SUBMIT/i }));
+});


src/components/ProtectedRoute.test.jsx
```diff
@@ -0,0 +1,40 @@
...
+import { MemoryRouter, Route } from 'react-router-dom';
+import { clearFirestoreData, initializeTestApp } from '../lib/testing';
+import ScholarshipForm from '../pages/ScholarshipForm';
+
+window.MutationObserver = require('mutation-observer');
+
+function renderAtRoute(route) {
+  return render(
+    <MemoryRouter initialEntries={[route]}>
+      <Route path="/scholarships/new" component={ScholarshipForm} />
```
josueetcom, line 12: A few things: 
1. This should be a **ProtectedRoute**
2. The `path` is technically arbitrary but I'd expect it to be `route`?
3. The `component` is also arbitrary but instead of the **ScholarshipForm** (which has its own requirements) I suggest having this method accept a `component` which in the tests can something as simple as `<h1>Protected header</h1>`

The way this is currently written this looks more like what a test might look like for **AddScholarship** (as that's what this page was renamed to).

src/components/ProtectedRoute.test.jsx
```diff
@@ -0,0 +1,40 @@
...
+    </MemoryRouter>
+  );
+}
+
+const app = initializeTestApp({
+  projectId: 'protected-route-test',
+  apiKey: 'fake-api-key',
+});
+
+beforeAll(() => clearFirestoreData(app.options));
```
josueetcom, line 22: Firestore is irrelevant here.

src/components/ProtectedRoute.test.jsx
```diff
@@ -0,0 +1,40 @@
...
+afterAll(() => app.delete());
+
+test('athenticated user can accesss page', () => {
```
josueetcom, line 25: Does this test work?

Technically there are three test cases:
1. Initially the **CircularProgress** is displayed when `user` is undefined.
2. "You must log in first to access this page" when the `user` is not signed in.
3. The passed in component is displayed when `user` is signed in.

I'm not sure exactly how to mock the user but something you can manually set `firebase.auth().currentUser`.

Maybe you can even test the `onAuthStateChanged` thing using [firebase.auth().updateCurrentUser()](https://firebase.google.com/docs/reference/js/auth.auth.md#authupdatecurrentuser) and collapse the first test case with the other two.

src/components/ProtectedRoute.test.jsx
```diff
@@ -0,0 +1,40 @@
...
+  expect(screen.getByLabelText('Deadline *')).toBeInTheDocument();
+  expect(screen.getByLabelText('Description *')).toBeInTheDocument();
+  expect(screen.getByLabelText('Website *')).toBeInTheDocument();
+
+  expect(screen.getByText('Amount Type *')).toBeInTheDocument();
+  expect(screen.getByText('Unknown')).toBeInTheDocument();
+  expect(screen.getByText('Fixed Amount:')).toBeInTheDocument();
+  // expect(screen.getByText('Range:')).toBeInTheDocument();
+  expect(screen.getByText('Full Tuition')).toBeInTheDocument();
+  expect(screen.getByRole('button', { name: /SUBMIT/i }));
```
josueetcom, line 39: If you use a simple component you only need one statement.

josueetcom (changes requested):

josueetcom:
I think the fake auth testing behavior in #698 should help with writing these test cases!

src/components/ProtectedRoute.test.jsx
josueetcom: *authenticated

src/components/ProtectedRoute.test.jsx
```diff
@@ -0,0 +1,38 @@
...
+afterAll(() => app.delete());
+
+test('athenticated user can accesss page', () => {
+  renderAtRoute('/authenticated');
+
+  expect(screen.getByText('Protected Header')).toBeInTheDocument();
+});
+
+test("unathenticated user can't access page", () => {
+  app.auth().signOut();
```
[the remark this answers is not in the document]
josueetcom: Hmm it might be the fact **ProtectedRoute** uses `firebase.auth()`. In #698 I changed `Scholarships.ts` to use `firebase.app().auth()` instead of `firebase.auth()`:
https://github.com/beyondhb1079/s4us/blob/eb8ba2ffe3a1b6de8f32c9135a39c1164c024a3b/src/models/Scholarships.ts#L12

I think I had to do that because I mocked the test app's auth function:
https://github.com/beyondhb1079/s4us/blob/eb8ba2ffe3a1b6de8f32c9135a39c1164c024a3b/src/lib/testing.ts#L45

Maybe adding a `firebase.auth = app.auth` statement before `return app` in the `initializeTestApp` method would fix this? 🤔

src/components/ProtectedRoute.test.jsx
```diff
@@ -1,40 +1,38 @@
...
+
+  renderAtRoute('/unauthenticated');
+
+  expect(screen.getByText('Protected Header')).toBeInTheDocument();
```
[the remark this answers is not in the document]
edwingl10: yes but since it was failing, I just left it like this for now.

src/components/ProtectedRoute.test.jsx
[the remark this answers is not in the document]
edwingl10: hmm I tried it but it throws an error
Screen Shot 2021-10-31 at 11 09 37 AM

PR #489, opened 2021-05-25
Refactor loading components to ScholarshipList component

### LINKED ISSUE #487: Refactor scholarship loading rendering into ScholarshipList component
Take a look at these two files:
https://github.com/beyondhb1079/s4us/blob/27a0d0e693f1e18b9298c9813260a67b7d6489e4/src/pages/UserHome.jsx#L164-L174

https://github.com/beyondhb1079/s4us/blob/27a0d0e693f1e18b9298c9813260a67b7d6489e4/src/pages/Scholarships.jsx#L146-L160

They're very similar:
- Both show a list of scholarships
- Both have a **Load More** button
- Both show [circular progress indicators](https://material-ui.com/components/progress/#circular) when loading (or loading more) scholarships

What if we update one page but forget to update the other? What if we add something similar in a different page/component? 

`ScholarshipList` only takes care of displaying the list but not the dynamics around it. Perhaps it makes sense to merge these there. Along with the `scholarships` property maybe it makes sense to add a `loadingMore` boolean property and `onLoadMore` callback property.

josueetcom:
- `Scholarships.list` has been updated so that it can accept `minAmount` and `maxAmount` filters instead of having to do those on its return value.
- **ScholarshipList** now supports displaying everything related to loading scholarship results, including:
  - paginated loading
  - error message displaying
  - no results (or no more results) displaying

## Motivation and Context

Fixes #487, which simplifies how we load and display scholarships, as can be evidenced by the simplification in the **Scholarships** and **UserHome** pages.

## How Has This Been Tested?

I've tested the changes on my local server and a few tests for the postProcessing filter logic

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-489.onrender.com/scholarships.
2. Try loading more scholarships.
3. Notice slight changes in UI.

Also try logging in and making sure the UI looks correct for your dashboard page.

## Checklist:


src/components/ScholarshipList.jsx
```diff
@@ -35,27 +56,47 @@ function ScholarshipList({ scholarships }) {
...
+        <Box className={classes.loadMore}>
+          {onLoadMore ? (
+            <Button color="primary" onClick={onLoadMore}>
+              Load More
+            </Button>
+          ) : (
+            <Typography>End of results</Typography>
+          )}
+        </Box>
+      )}
```
josueetcom, line 85: Maybe this should be embedded in the Grid itself?

josueetcom: Figured it out!

src/pages/Scholarships.jsx
```diff
@@ -33,30 +27,14 @@ const sortOptions = {
...
-  const [canLoadMore, setCanLoadMore] = useState(false);
-
   const history = useHistory();
   const location = useLocation();
 
   const params = queryString.parse(location.search, { parseNumbers: true });
 
   const setQueryParam = (index, val) => {
-    setScholarships([]);
+    // setScholarships([]); needed?
```
josueetcom, line 37: TODO(me): Figure out if this is still needed

josueetcom: It is, elsewhere though.

src/models/Scholarships.ts
```diff
@@ -73,7 +75,10 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
       query = query.where('authorId', '==', opts.authorId);
     }
 
-    return FirestoreCollection.list(query);
+    const postProcessFilter = (s: FirestoreModel<ScholarshipData>) =>
+      s.data.amount.intersectsRange(opts.minAmount, opts.maxAmount);
```
edwingl10, line 79: Nice!

src/components/ScholarshipList.jsx
```diff
@@ -52,10 +134,12 @@ function ScholarshipList({ scholarships }) {
...
 
 ScholarshipList.propTypes = {
-  scholarships: PropTypes.arrayOf(PropTypes.object),
+  listFn: PropTypes.func,
+  noResultsNode: PropTypes.node,
 };
 ScholarshipList.defaultProps = {
-  scholarships: [],
+  listFn: undefined,
+  noResultsNode: <Typography>No scholarships found</Typography>,
```
edwingl10, line 142: ooo I didn't know we could set jsx as default props

edwingl10 (approved):

gonzalo1330 (approved):

src/components/ScholarshipList.jsx
josueetcom: Yeah! The only thing tricky about it is when you need to set a `className`. In that case, I would set this `undefined` and when rendering set the value there.

Issue #391, opened 2021-02-19
Component unit tests: HomeSection


josueetcom:
We need unit tests for the [HomeSection](https://github.com/beyondhb1079/s4us/blob/main/src/components/HomeSection.jsx). Consider any edge cases. See the [Testing Library docs](https://testing-library.com/docs/) and existing tests for reference.

PR #494, opened 2021-05-30
HomeSection Tests

Jobhdez:
## Motivation and Context

Simple tests for the HomeSection component i.e. issue #391.

I will add more tests. I just wanted to commit and start contributing to DreamScholars.

I wanted to get a review on the simple test I did because I have not tested web sites before, but I am reading a tutorial right now.

thanks.

## How Has This Been Tested?
`yarn start` and `yarn test` works.

## Types of changes

- [x] testing component

### DIFF
--- a/src/components/HomeSection.jsx
+++ b/src/components/HomeSection.jsx
@@ -68,6 +68,7 @@ HomeSection.defaultProps = {
   alignItems: 'flex-start',
   direction: 'row',
   tab: '',
+  buttons: '[]',
 };
 
 HomeSection.propTypes = {
--- a/src/components/HomeSection.jsx
+++ b/src/components/HomeSection.jsx
@@ -21,15 +21,8 @@ const useStyles = makeStyles((theme) => ({
 
 export default function HomeSection(props) {
   const classes = useStyles();
-  const {
-    alignItems,
-    direction,
-    tab,
-    title,
-    description,
-    buttons,
-    pic,
-  } = props;
+  const { alignItems, direction, tab, title, description, buttons, pic } =
+    props;
 
   return (
     <Container className={classes.root}>
--- a/src/components/HomeSection.jsx
+++ b/src/components/HomeSection.jsx
@@ -21,8 +21,15 @@ const useStyles = makeStyles((theme) => ({
 
 export default function HomeSection(props) {
   const classes = useStyles();
-  const { alignItems, direction, tab, title, description, buttons, pic } =
-    props;
+  const {
+    alignItems,
+    direction,
+    tab,
+    title,
+    description,
+    buttons,
+    pic,
+  } = props;
 
   return (
     <Container className={classes.root}>
@@ -69,6 +76,6 @@ HomeSection.propTypes = {
   tab: PropTypes.string,
   title: PropTypes.string.isRequired,
   description: PropTypes.string.isRequired,
-  buttons: PropTypes.arrayOf(PropTypes.element).isRequired,
+  buttons: PropTypes.arrayOf(PropTypes.element),
   pic: PropTypes.string.isRequired,
 };
--- a/src/components/HomeSection.jsx
+++ b/src/components/HomeSection.jsx
@@ -21,8 +21,15 @@ const useStyles = makeStyles((theme) => ({
 
 export default function HomeSection(props) {
   const classes = useStyles();
-  const { alignItems, direction, tab, title, description, buttons, pic } =
-    props;
+  const {
+    alignItems,
+    direction,
+    tab,
+    title,
+    description,
+    buttons,
+    pic,
+  } = props;
 
   return (
     <Container className={classes.root}>
@@ -61,7 +68,7 @@ HomeSection.defaultProps = {
   alignItems: 'flex-start',
   direction: 'row',
   tab: '',
-  buttons: '[]',
+  buttons: [],
 };
 
 HomeSection.propTypes = {
--- a/src/components/HomeSection.test.jsx
+++ b/src/components/HomeSection.test.jsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { render, screen } from '@testing-library/react';
-import { initializeTestApp } from '../lib/testing';
 import HomeSection from './HomeSection';
 
 describe('HomeSection', () => {
--- a/src/components/HomeSection.test.jsx
+++ b/src/components/HomeSection.test.jsx
@@ -31,16 +31,11 @@ describe('HomeSection', () => {
 
   test('renders no buttons by default', () => {
     render(
-      <HomeSection
-        title="Topology"
-        description="Math course"
-        buttons={[]}
-        pic="manifold"
-      />
+      <HomeSection title="Topology" description="Math course" pic="manifold" />
     );
     screen.debug();
-    const items = screen.queryByRole('button');
-    expect(items).toBe(null);
+    const item = screen.queryByRole('button');
+    expect(item).toBe(null);
   });
 
   test('render given buttons', async () => {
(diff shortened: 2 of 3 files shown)


src/components/HomeSection.test.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import HomeSection from './HomeSection';
+
+describe('HomeSection', () => {
+  test('renders HomeSection component', () => {
```
josueetcom, line 7: This test renders a `HomeSection` but doesn't test anything about it. You probably want to have multiple tests that try out the default property values and varying values and makes sure that:
1. They render and don't break unexpectedly
2. The properties given are reflected in the rendered DOM (e.g. the title is there, the number of buttons given are there, etc.) 

I suggest reading the [Testing Library Docs](https://testing-library.com/docs/) to get a better sense of what to test. Specifically, you'll want to learn about [Queries](https://testing-library.com/docs/queries/about/) and read about the [DOM Testing Library](https://testing-library.com/docs/dom-testing-library/intro) and [React Testing Library](https://testing-library.com/docs/react-testing-library/intro). There are plenty of examples there as well as in other `.test.jsx` files in our repo.

josueetcom (changes requested):

src/components/HomeSection.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+      <HomeSection
+        title="Topology"
+        description="Math course"
+        buttons={[]}
+        pic="manifold"
+      />
+    );
+    screen.debug();
+    const items = screen.queryByRole('button');
+    expect(items).toBe(null);
```
josueetcom, line 42: Good job identifying to use ***ByRole**!

src/components/HomeSection.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+        title="Topology"
+        description="Math course"
+        buttons={[]}
+        pic="manifold"
+      />
+    );
+    expect(screen.getByText('Math course')).toBeInTheDocument();
+  });
+
+  test('checks if it renders the given number of buttons', () => {
```
josueetcom, line 31:
```suggestion
  test('renders no buttons by default', () => {
```

Or better yet have a 'default render' test case that checks: no buttons are rendered and the title/description/pic are empty (you may need to set `defaultProps` for these in the component itself). Then, create another test case that checks the presence of one or more buttons, e.g. 'renders buttons given')

src/components/HomeSection.test.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import HomeSection from './HomeSection';
+
+describe('HomeSection', () => {
+  test('checks if the title is in the document', () => {
```
josueetcom, line 6:
```suggestion
  test('renders given title', () => {
```

src/components/HomeSection.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+        description="Math course"
+        buttons={[]}
+        pic="manifold"
+      />
+    );
+    screen.debug();
+    expect(screen.getByText('Topology')).toBeInTheDocument();
+  });
+
+  test('checks if the description is in the document', () => {
```
josueetcom, line 19:
```suggestion
  test('renders given description', () => {
```

src/components/HomeSection.test.jsx
josueetcom, line 31: A few more (optional) test case ideas:
- Renders given pic
- Renders given tab name (that `tab` property should maybe be called `tabName`)

The defaults can be part of that "default render" test case.

These are possible test cases too, but these are kind of obscure and might be a bit harder to test so you can skip these if you like:
- Renders in given direction
- Renders in given alignItems

josueetcom (changes requested):

src/components/HomeSection.test.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import HomeSection from './HomeSection';
+
+describe('HomeSection', () => {
```
josueetcom, line 5: I didn't know this existed! Looks handy for grouping tests

src/components/HomeSection.test.jsx
```diff
@@ -41,4 +42,30 @@ describe('HomeSection', () => {
...
+                    variant="outlined"
+                    color="primary">
+                    Hello
+                  </Button>,
+                  <Button
+                    to="/about"
+                    variant="outlined"
+                    color="primary">
+                    Hello
+                  </Button>,
```
josueetcom: You can simplify this by removing the optional properties:
```suggestion
                  <Button>Hello</Button>,
                  <Button>Hello</Button>,
```

You're not testing against those property values anyways.

src/components/HomeSection.test.jsx
```diff
@@ -28,7 +29,7 @@ describe('HomeSection', () => {
     expect(screen.getByText('Math course')).toBeInTheDocument();
   });
 
-  test('checks if it renders the given number of buttons', () => {
+  test('renders no buttons by default', () => {
```
josueetcom, line 32: The "by default" part here assumes that `buttons` is an optional property. The way it's written here, you're explicitly passing no buttons as opposed to rendering the default value. You need to make `buttons` an optional property in `HomeSection.jsx`.

src/components/HomeSection.test.jsx
```diff
@@ -0,0 +1,58 @@
...
+    );
+    expect(screen.getByText('Math course')).toBeInTheDocument();
+  });
+
+  test('renders no buttons by default', () => {
+    render(
+      <HomeSection
+        title="Topology"
+        description="Math course"
+        buttons={[]}
```
josueetcom, line 37: Delete this line as this test case is testing the default..

src/components/HomeSection.test.jsx
josueetcom: Btw, since `queryByRole` only returns one element I suggest replacing this with:

```js
    const item = screen.queryByRole('button');
    expect(item).toBe(null);
```
or:
```js
    const items = screen.queryAllByRole('button');
    expect(items.length).toBe(0);
```

src/components/HomeSection.jsx
```diff
@@ -61,6 +61,7 @@ HomeSection.defaultProps = {
   alignItems: 'flex-start',
   direction: 'row',
   tab: '',
+  buttons: '[]',
```
josueetcom, line 64: Look at line 73, it spells out `buttons`' `PropTypes` is an `arrayOf` stuff. So this should be an array, not a string.
```suggestion
  buttons: [],
```

josueetcom (changes requested):
Just a few tweaks otherwise should be good to merge next round.

src/components/HomeSection.test.jsx
Jobhdez: delete line 37?

josueetcom: Of course. Otherwise how else are you supposed to test the _default_ `buttons` value?

Jobhdez: alright - just making sure.

src/components/HomeSection.test.jsx
```diff
@@ -0,0 +1,43 @@
...
+  test('renders no buttons by default', () => {
+    render(
+      <HomeSection title="Topology" description="Math course" pic="manifold" />
+    );
+    screen.debug();
+    const item = screen.queryByRole('button');
+    expect(item).toBe(null);
+  });
+
+  test('render given buttons', async () => {
```
josueetcom, line 31: *`renders`

josueetcom (approved):
Just a typo fix otherwise LGTM. Note: you can fix the typo, commit, push, and then merge

PR #510, opened 2021-06-16
shareDialog Refactor

### LINKED ISSUE #373: Refactor ShareDialog to be dependent on location state
Currently, it's closely tied to [ScholarshipList](https://github.com/beyondhb1079/s4us/blob/ddef07f08dbc7cae3b03c56ef5e60e1b1fb0e84f/src/components/ScholarshipList.jsx#L123-L128) and its state. This means the [sharing logic](https://github.com/beyondhb1079/s4us/blob/main/src/components/ScholarshipList.jsx#L49-L67) will need to be copied/pasted/tweaked slightly between different components that want to use the ShareDialog component.

Instead of this, how about we encapsulate this logic based on Location state and conditionally show ShareDialog like we do with [LoginDialog](https://github.com/beyondhb1079/s4us/search?q=LoginDialog)? We can use state variables like we use [showLoginDialog](https://github.com/beyondhb1079/s4us/search?q=showLoginDialog) in order to determine whether to show the ShareDialog and what information to share.

edwingl10:
## Motivation and Context
Refactored logic for ShareDialog component to depend on location state. 3 main properties are passed in state, whether to show the shareDialog, the title to display, and the URL to display. 

Fixes #373 

## How Has This Been Tested?

On localhost 

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-510.onrender.com/scholarships.
2. Choose any scholarship
3. Click the share button

## Checklist:

### DIFF
--- a/src/components/ScholarshipDetailCard.jsx
+++ b/src/components/ScholarshipDetailCard.jsx
@@ -74,7 +74,7 @@ export default function ScholarshipDetailCard({ scholarship }) {
 
   const shareText = `${
     shareDialog.shareTitle
-  } | ${deadline.toLocaleDateString()}\n`;
+  } | ${deadline?.toLocaleDateString()}\n`;
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars, no-unused-vars
   const shareFn = () => {
--- a/src/components/ScholarshipDetailCard.jsx
+++ b/src/components/ScholarshipDetailCard.jsx
@@ -13,6 +13,7 @@ import {
 } from '@material-ui/core';
 import { Share, Send, Info } from '@material-ui/icons';
 import ScholarshipAmount from '../types/ScholarshipAmount';
+import { BRAND_NAME } from '../config/constants';
 
 const useStyles = makeStyles((theme) => ({
   actionSection: {
@@ -68,13 +69,13 @@ export default function ScholarshipDetailCard({ scholarship }) {
 
   const shareDialog = {
     showShareDialog: true,
-    shareTitle: `${amount?.toString()} - ${name}`,
+    shareTitle: `${amount?.toString()} - ${name} | ${BRAND_NAME}`,
     shareURL: `https://${window.location.hostname}/scholarships/${scholarship.id}`,
   };
 
   const shareText = `${
     shareDialog.shareTitle
-  } | ${deadline?.toLocaleDateString()}\n`;
+  } \n ${deadline?.toLocaleDateString()}\n`;
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars, no-unused-vars
   const shareFn = () => {
--- a/src/components/ScholarshipDetailCard.jsx
+++ b/src/components/ScholarshipDetailCard.jsx
@@ -67,32 +67,24 @@ export default function ScholarshipDetailCard({ scholarship }) {
     eligibility,
   } = scholarship.data;
 
-  const shareDialog = {
-    showShareDialog: true,
-    shareTitle: `${amount?.toString()} - ${name} | ${BRAND_NAME}`,
-    shareURL: `https://${window.location.hostname}/scholarships/${scholarship.id}`,
-  };
-
-  const shareText = `${
-    shareDialog.shareTitle
-  } \n ${deadline?.toLocaleDateString()}\n`;
-
   // eslint-disable-next-line @typescript-eslint/no-unused-vars, no-unused-vars
   const shareFn = () => {
+    const data = {
+      title: `${amount?.toString()} - ${name} | ${BRAND_NAME}`,
+      text: `${amount?.toString()} - ${name} | ${BRAND_NAME} \n ${deadline?.toLocaleDateString()}\n`,
+      url: `https://${window.location.hostname}/scholarships/${scholarship.id}`,
+    };
+
     if (navigator.share) {
       navigator
-        .share({
-          title: shareDialog.shareTitle,
-          url: shareDialog.shareURL,
-          text: shareText,
-        })
+        .share(data)
         // eslint-disable-next-line no-console
         .then(() => console.log('Thanks for sharing!'))
         // eslint-disable-next-line no-console
         .catch(console.error);
     } else {
       history.replace({
-        state: { shareDialog },
+        state: { showShareDialog: true, shareData: data },
       });
     }
   };
--- a/src/components/ScholarshipDetailCard.jsx
+++ b/src/components/ScholarshipDetailCard.jsx
@@ -66,6 +66,36 @@ export default function ScholarshipDetailCard({ scholarship }) {
     eligibility,
   } = scholarship.data;
 
+  const shareDialog = {
+    showShareDialog: true,
+    shareTitle: `${amount?.toString()} - ${name}`,
+    shareURL: `https://${window.location.hostname}/scholarships/${scholarship.id}`,
+  };
+
+  const shareText = `${
+    shareDialog.shareTitle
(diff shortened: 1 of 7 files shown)


src/components/ScholarshipDetailCard.jsx
```diff
@@ -111,7 +112,16 @@ export default function ScholarshipDetailCard({ scholarship }) {
...
           variant="outlined"
           className={classes.shareBtn}
-          startIcon={<Share />}>
+          startIcon={<Share />}
+          onClick={() =>
+            history.replace({
+              state: {
+                showShareDialog: true,
+                shareTitle: `${amount?.toString()} - ${name}`,
+                shareURL: `https://${window.location.hostname}/scholarships/${scholarship.id}`,
```
edwingl10, line 121: I tried doing something like
`<Button
component={Link} 
to={{state: showLoginDialog : true}}
replace/>` but it was not working. Not sure why.

src/components/ScholarshipDetailCard.jsx
```diff
@@ -111,7 +112,16 @@ export default function ScholarshipDetailCard({ scholarship }) {
...
+              },
+            })
+          }>
```
josueetcom, line 124: What's the issue with doing it this way?
```js
import { Link } from 'react-router-dom';
```
```suggestion
                component={Link}
                to={{ state: {
                  showShareDialog: true,
                  shareTitle: `${amount?.toString()} - ${name}`,
                  shareURL: `https://${window.location.hostname}/scholarships/${scholarship.id}`,
                } }}
                replace>
```

src/components/ScholarshipDetailCard.jsx
josueetcom, line 121: You could also use a state variable like this:
```js
shareDialog: {
  title: ...,
  url: ...
}
```

and then simply check for the presence of `shareDialog`.

src/components/ScholarshipList.jsx
```diff
@@ -26,32 +24,6 @@ const useStyles = makeStyles(() => ({
...
-        // eslint-disable-next-line no-console
-        .then(() => console.log('Thanks for sharing!'))
-        // eslint-disable-next-line no-console
-        .catch(console.error);
-    } else {
-      setShareSiteLink(url);
-      setShareSiteTitle(title);
-      setShareDialogOpen(true);
-    }
-  };
```
josueetcom, line 54: You could also keep this in the DetailCard and in the `else` case do the `history.replace` you ended up doing. I think where possible it may be better to use the native share dialog as it usually redirects to another app anyway.

src/components/ScholarshipDetailCard.jsx
edwingl10: The problem was that I was using the wrong `Link`

src/components/ScholarshipList.jsx
edwingl10: Wouldn't a better place for this be in `ShareDialog`

josueetcom: Hmm, potentially but in that case in `ShareDialog` you'll want to render nothing (`</>`) if the navigator supports sharing natively.

src/components/ShareDialog.jsx
```diff
@@ -36,51 +37,52 @@ const useStyles = makeStyles((theme) => ({
...
-export default function ShareDialog(props) {
+export default function ShareDialog() {
+  const location = useLocation();
+  const { showShareDialog, shareTitle, shareURL } = location.state || {
+    showShareDialog: false,
+  };
+
+  const history = useHistory();
+  const closeDialog = () =>
+    history.replace({ state: { showShareDialog: false } });
```
josueetcom, line 48: On Render, there's a slight glitch when you dismiss the ShareDialog. I guess maybe you should keep the title/link state but only change `show` to `false` so that when the dialog fades the glitch doesn't show up.

src/components/ScholarshipList.jsx
edwingl10: For now I just left it in `ScholarshipDetailCard`

src/components/ScholarshipDetailCard.jsx
edwingl10: I had made the changes but then kept the original version to tackle the glitch when closing the dialog.

src/components/ScholarshipDetailCard.jsx
```diff
@@ -63,6 +67,36 @@ export default function ScholarshipDetailCard({ scholarship }) {
...
 
+  const shareDialog = {
+    showShareDialog: true,
+    shareTitle: `${amount?.toString()} - ${name} | ${BRAND_NAME}`,
+    shareURL: `https://${window.location.hostname}/scholarships/${scholarship.id}`,
+  };
+
+  const shareText = `${
+    shareDialog.shareTitle
+  } \n ${deadline?.toLocaleDateString()}\n`;
```
josueetcom, line 78: These don't need to be defined outside of the `shareFn`.

src/components/ScholarshipDetailCard.jsx
```diff
@@ -63,6 +67,36 @@ export default function ScholarshipDetailCard({ scholarship }) {
...
+        })
+        // eslint-disable-next-line no-console
+        .then(() => console.log('Thanks for sharing!'))
+        // eslint-disable-next-line no-console
+        .catch(console.error);
+    } else {
+      history.replace({
+        state: { shareDialog },
+      });
+    }
```
josueetcom, line 97: I think to keep things simple we should use the [share `data` object parameter](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share#parameters) `navigator.share` expects and set `showShareDialog` as a separate state variable. E.g.:

```js
  const shareFn = () => {
    const data = { url: ..., text: ..., title: ...};
    if (navigator.share) {
      navigator.share(data)
        // eslint-disable-next-line no-console
        .then(() => console.log('Thanks for sharing!'))
        // eslint-disable-next-line no-console
        .catch(console.error);
    } else {
      history.replace({
        state: { showShareDialog: true, shareData: data },
      });
    }
  };
```

src/components/ShareDialog.jsx
```diff
@@ -36,51 +36,55 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-export default function ShareDialog(props) {
+export default function ShareDialog() {
+  const location = useLocation();
+  const { showShareDialog, shareTitle, shareURL } = location.state
+    ?.shareDialog || {
+    showShareDialog: false,
+  };
```
josueetcom, line 44: Using the other suggested format in my other comment I think this can be changed to:
```jsx
  const showShareDialog = location.state.showShareDialog ?? false;
  const { text, title, url } = location.state.shareData ?? {};
```

src/components/ShareDialog.jsx
```diff
@@ -36,51 +36,55 @@ const useStyles = makeStyles((theme) => ({
...
+
+  const history = useHistory();
+  const closeDialog = () =>
+    history.replace({
+      state: { shareDialog: { shareTitle, shareURL, showShareDialog: false } },
```
josueetcom, line 49: And then I think this would either be:
```suggestion
      state: { showShareDialog: false },
```
or if the `shareData` state isn't preserved above then:
```suggestion
      state: { shareData },
```

I think it's the latter.

josueetcom (changes requested):

src/components/ShareDialog.jsx
edwingl10: yup the latter works.

josueetcom (approved):
Looks good to me!

PR #535, opened 2021-07-02
Finally fix autoapprove dependabot workflow

josueetcom:
I finally understand why `dependabot` triggered checks like https://github.com/beyondhb1079/s4us/pull/534/checks?check_run_id=2969026240 keep failing.

For security reasons, the default [permissions](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token) on the GitHub token is read-only for Dependabot pull requests so we have to use the `pull_request_target` event instead.

Additional details in dependabot/dependabot-core#3253.

### DIFF
--- a/.github/workflows/approvals.yml
+++ b/.github/workflows/approvals.yml
@@ -4,53 +4,7 @@ on:
   pull_request:
     branches: [main]
 
-permissions:
-  contents: read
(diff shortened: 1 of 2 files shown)


Jobhdez (approved):
It looks great.

Issue #93, opened 2020-10-18
Scholarship model: add fields forms and filters rely on


josueetcom:
Namely, the majors and grade level requirements fields though maybe I missed others.

PR #562, opened 2021-07-12
Scholarship Model Schema design changes

### LINKED ISSUE #56: Schema design: Scholarship model
We need to establish what a Scholarship object consists of and how it will be stored in Firestore.

Some [prior related work](https://github.com/beyondhb1079/undocuscholar-web/blob/master/scholarships/Design.md).

Things to consider:
- Firestore is a NoSQL, document-oriented database. See their [Data model](https://cloud.google.com/firestore/docs/data-model).
- See [Structuring data](https://cloud.google.com/firestore/docs/concepts/structure-data) for tips on how to structure data and [Data types](https://cloud.google.com/firestore/docs/concepts/data-types) for info on what data types are supported.

The expectation here is to come up with the Scholarship model in some form.

edwingl10:
## Motivation and Context

Made changes to the scholarship schema design to better support various properties. 
Also updated wherever we called on the old properties. 

Fixes #56 and #93

## How Has This Been Tested?
Using Firestore emulator to check if fields are added.

## Types of changes

## Previewing Changes

Won't really be able to see the changes unless you use firestore emulator. 

2.

## Checklist:


josueetcom (changes requested):
Sorry, I meant to send these comments on Wednesday

src/models/Scholarships.test.ts
```diff
@@ -204,3 +204,11 @@ test('scholarships.list - filters by maxAmount', async () => {
     want.map(extractName).sort()
   );
 });
+
+test('dateAdded & lastModified gets set on saving a new scholarship', async () => {
```
josueetcom, line 208:
```suggestion
test('save new scholarship - dateAdded and lastModified get set', async () => {
```

Another test to add:
```suggestion
test('save updated scholarship - dateAdded unchanged but lastModified set', async () => {
```

src/models/Scholarships.ts
```diff
@@ -4,10 +4,9 @@ import FirestoreCollection from './base/FirestoreCollection';
 import FirestoreModelList from './base/FiretoreModelList';
 import ScholarshipEligibility from '../types/ScholarshipEligibility';
 import FirestoreModel from './base/FirestoreModel';
+import Model from './base/Model';
 
-interface ScholarshipData {
-  // TODO(https://github.com/beyondhb1079/s4us/issues/56):
-  // Update this to reflect the schema
+export interface ScholarshipData {
```
josueetcom, line 9: Does this need to be exported?

src/models/Scholarships.ts
```diff
@@ -100,6 +98,13 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
 
     return FirestoreCollection.list(query, postProcessFilter);
   }
+
+  save(data: ScholarshipData): Promise<Model<ScholarshipData>> {
+    const Data = { ...data, lastModified: new Date() };
```
josueetcom, line 103: [23.2](https://github.com/airbnb/javascript#naming--camelCase) Use camelCase when naming objects, functions, and instances.

```suggestion
    const newData = { ...data, lastModified: new Date() };
```

src/models/Scholarships.ts
```diff
@@ -100,6 +98,13 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
...
+    if (!data.dateAdded) Data.dateAdded = new Date();
```
josueetcom, line 104: There's a slight time difference between the two lines. I suggest doing this instead so that the date values are consistent for new scholarships:
```suggestion
    if (!data.dateAdded) newData.dateAdded = newData.lastModified;
```

src/models/Scholarships.ts
```diff
@@ -100,6 +98,13 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
...
+
+    return super.new(Data).save();
```
josueetcom, line 106: I think this doesn't work for existing scholarships (you might try adding a test case that proves/disproves my theory). If you call `Scholarships.save(data)` on an existing scholarship `.new()` creates a new document with a new ID (hence the function name) so the scholarship wouldn't be saved under the same ID. If you call `scholarship.save()` on an existing scholarship then none of the logic here will run.

Now how to make this work, hmmm.  There's a couple options:
- Create a class that extends `FirestoreModel<Scholarship>` and overrides the `.save()` method with this business logic.
- Remove the `.save()` method from `FirestoreModel<Scholarship>` and accept a `FirestoreModel<Scholarship>` instead.

I'm not sure which is easier but in any case I recommend adding a test case to make sure you're implementing it correctly!

src/types/ScholarshipEligibility.ts
```diff
@@ -7,5 +7,5 @@ export default interface ScholarshipEligibility {
   schools?: string[];
   grades?: string[];
   states?: String[];
-  gender?: GenderType;
+  gender?: Gender[];
```
josueetcom, line 10:
```suggestion
  genders?: Gender[];
```

src/models/Scholarships.test.ts
josueetcom: Might have to change where these tests are based on other comments.

josueetcom (changes requested):

src/models/Scholarships.ts
edwingl10: oops, was testing something and forgot to change it back.

src/models/base/ScholarshipModel.ts
```diff
@@ -0,0 +1,13 @@
+// import firebase from 'firebase/app';
+import FirestoreModel from './FirestoreModel';
+
+export default class ScholarshipModel<
+  T extends { lastModified?: Date; dateAdded?: Date }
```
edwingl10, line 5: without this it was complaining that `dateAdded` and `lastModified' did not exist on type T

src/models/base/ScholarshipModel.ts
```diff
@@ -0,0 +1,13 @@
+// import firebase from 'firebase/app';
+import FirestoreModel from './FirestoreModel';
+
+export default class ScholarshipModel<
```
josueetcom, line 4: **ScholarshipModel** is not a base class for anything. I suggest moving this up one folder.

src/models/base/ScholarshipModel.ts
```diff
@@ -0,0 +1,13 @@
...
+  T extends { lastModified?: Date; dateAdded?: Date }
+> extends FirestoreModel<T> {
```
josueetcom, line 6: Why does this class have generics? What does the *T* mean in **ScholarshipModel<T>**? What are the possible values of **T**? If it's only **ScholarshipData** then why not write this like this:
```ts
export default class ScholarshipModel extends FirestoreModel<ScholarshipData> {
```

src/models/base/FirestoreCollection.ts
```diff
@@ -14,8 +15,8 @@ export default abstract class FirestoreCollection<T> {
       .withConverter(this.converter);
   }
 
-  new(data?: T): Model<T> {
-    return new FirestoreModel<T>(this.collection.doc(), data ?? ({} as T));
+  new(data?: T): ScholarshipModel<T> {
+    return new ScholarshipModel<T>(this.collection.doc(), data ?? ({} as T));
```
josueetcom, line 19: This is incorrect. **FirestoreCollection** is a generic class for _any_ kind of collection, not just the Scholarships one. You can override this method in the _Scholarship-specific_ defined subclass in `models/Scholarships.ts` instead.

src/types/ScholarshipEligibility.ts
```diff
@@ -1,5 +1,11 @@
+import Gender from './Gender';
+
 export default interface ScholarshipEligibility {
   gpa?: number;
   ethnicities?: string[];
   majors?: string[];
+  schools?: string[];
+  grades?: string[];
+  states?: String[];
```
josueetcom, line 9:
```suggestion
  states?: string[];
```

josueetcom (changes requested):

src/models/base/ScholarshipModel.ts
edwingl10: I did this because I was running into a dependency cycle but I was able to fix it by moving the `ScholarshipData` into it's own file located in `types/ScholarshipData.ts`

src/types/ScholarshipEligibility.ts
edwingl10: oops typos! 😣

src/models/ScholarshipModel.test.ts
```diff
@@ -0,0 +1,49 @@
+import firebase from 'firebase/app';
+import { clearFirestoreData, initializeTestApp } from '../lib/testing';
+import ScholarshipAmount from '../types/ScholarshipAmount';
+import ScholarshipModel from './ScholarshipModel';
+import { converter } from './Scholarships';
+
+const app = initializeTestApp({ projectId: 'fs-model-test' });
```
josueetcom, line 7:
```suggestion
const app = initializeTestApp({ projectId: 'scholarship-model-test' });
```

src/models/ScholarshipModel.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  amount: new ScholarshipAmount(),
+};
+
+test('save new scholarship - dateAdded and lastModified get set', async () => {
+  const ref = scholarships.doc('scholarship 1');
+  const scholarship = new ScholarshipModel(ref, data);
+
+  await expect(scholarship.save()).resolves.toBeDefined();
+
+  const got = await scholarships.doc(scholarship.id).get();
```
josueetcom, line 31:
```suggestion
  const got = (await scholarships.doc(scholarship.id).get()).data();
```

src/models/ScholarshipModel.test.ts
```diff
@@ -0,0 +1,49 @@
...
+test('save updated scholarship - dateAdded unchanged but lastModified set', async () => {
+  const ref = scholarships.doc('scholarship 1');
+  const scholarship = new ScholarshipModel(ref, data);
+  await scholarship.save();
+
+  const { dateAdded, lastModified } = scholarship.data;
+  scholarship.data.name = 'updated scholarship';
+  await expect(scholarship.save()).resolves.toBeDefined();
+
+  const got = await scholarships.doc(scholarship.id).get();
```
josueetcom, line 45:
```suggestion
  const got = (await scholarships.doc(scholarship.id).get()).data();
```

src/models/Scholarships.test.ts
```diff
@@ -90,15 +93,19 @@ test('converter.fromFirestore', () => {
     description: 'description',
     deadline: firebase.firestore.Timestamp.fromDate(new Date('2019-02-20')),
```
josueetcom, line 94: You could extract these `new Date(...)` variables and use them instead. That way we don't get multiple dates mixed up :)

src/models/Scholarships.test.ts
```diff
@@ -110,7 +117,12 @@ test('converter.fromFirestore', () => {
     {}
   );
 
-  expect(got).toEqual({ ...snapdata, deadline: new Date('2019-02-20') });
+  expect(got).toEqual({
+    ...snapdata,
+    deadline: new Date('2019-02-20'),
+    dateAdded: got.dateAdded,
+    lastModified: got.lastModified,
```
josueetcom, line 124: Definitely want to use the dates from above (probably stored as a variable) cuz otherwise this only checks that these are what they already are.

src/models/Scholarships.ts
```diff
@@ -72,14 +73,21 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
...
     }
     if (opts.authorId) {
-      query = query.where('authorId', '==', opts.authorId);
+      query = query.where('author.id', '==', opts.authorId);
     }
 
     const postProcessFilter = (s: FirestoreModel<ScholarshipData>) =>
       s.data.amount.intersectsRange(opts.minAmount, opts.maxAmount);
 
     return FirestoreCollection.list(query, postProcessFilter);
```
josueetcom, line 82: I think both the `.get()` and `.list()` methods followed by `.save()` may no longer work because they return `FirestoreModel<ScholarshipData>` instead of `ScholarshipModel`. And it's hard to fix that atm :/

Leave a TODO comment here and file a GitHub issue to fix this (probably the overall base classes need fixing) and I'll take a look at it.

josueetcom (approved):

src/models/Scholarships.ts
```diff
@@ -72,14 +73,21 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
...
   }
+
+  new(data?: ScholarshipData): ScholarshipModel {
```
josueetcom, line 85: You should override `id()` as well so that it returns a **ScholarshipModel**

josueetcom (approved):

src/models/Scholarships.test.ts
```diff
@@ -77,9 +81,9 @@ test('converter.toFirestore', () => {
 
   expect(got).toEqual({
     ...data,
-    deadline: firebase.firestore.Timestamp.fromDate(data.deadline),
-    dateAdded: got.dateAdded,
-    lastModified: got.lastModified,
+    deadline: firebase.firestore.Timestamp.fromDate(deadline),
+    dateAdded: null,
+    lastModified: null,
```
josueetcom, line 86:
```suggestion
    deadline: firebase.firestore.Timestamp.fromDate(deadline),
```

josueetcom (approved):

PR #571, opened 2021-07-27
Remove the unused Model.subscribe() function

### LINKED ISSUE #92: Tests: FirestoreModel.subscribe()
We need to add unit tests for [**`FirestoreModel.subscribe()`**](https://github.com/beyondhb1079/s4us/blob/1d535cf645a1f98dd94dff6e0f4799f7db756dec/src/models/FirestoreModel.ts#L41-L49). 

Some test case ideas:
- **`subscribe`** invokes **`onChange`** observer
- **`subscribe`** invokes **`onError`** handler
- **`subscribe`**'s returned function unsubscribes from further updates (which means **`onChange`** and **`onError`** are no longer invoked)

josueetcom:
## Motivation and Context

We're not using this so let's remove it. It can be re-added later if needed. Fixes #92

### DIFF
--- a/src/models/base/FirestoreModel.ts
+++ b/src/models/base/FirestoreModel.ts
@@ -29,20 +29,4 @@ export default class FirestoreModel<T> implements Model<T> {
   delete(): Promise<void> {
     return this.ref.delete();
   }
-
-  subscribe(
-    onChange: (model: Model<T>) => void,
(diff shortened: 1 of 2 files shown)


Jobhdez (approved):
Removing clutter is good. Makes code more concise.

PR #596, opened 2021-08-28
Scholarship Form Rename and Refactor

### LINKED ISSUE #594: Scholarship form: rename + refactor
Rename `ScholarshipForm.jsx` file to better describe the pages purpose. 

Refactor the scholarshipForm into its own component so it can be re-used when editing a scholarship.

edwingl10:
## Motivation and Context

Fixes #594 

## How Has This Been Tested?

on laptop using firefox. 

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-596.onrender.com/scholarships/new.
2. It should look the same as before

## Checklist:


src/components/ScholarshipForm.jsx
```diff
@@ -98,7 +98,6 @@ function ScholarshipForm() {
   return (
     <Container maxWidth="md">
```
josueetcom, line 99: Now that this component is no longer a page component I suggest removing this as this is already in the page component.

src/components/ScholarshipForm.jsx
```diff
@@ -186,8 +185,15 @@ function ScholarshipForm() {
           </Button>
         </div>
       </form>
-      {submissionAlert}
     </Container>
   );
 }
+
+ScholarshipForm.propTypes = {
+  setSubmissionAlert: PropTypes.func,
```
josueetcom, line 193: Should this maybe be required so it doesn't default to `undefined`? Invoking `undefined` throws an exception I believe.

src/components/ScholarshipForm.jsx
```diff
@@ -23,9 +24,8 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-function ScholarshipForm() {
+function ScholarshipForm({ setSubmissionAlert }) {
   const classes = useStyles();
-  const [submissionAlert, setSubmissionAlert] = useState(null);
   const user = firebase.auth().currentUser;
 
   const formik = useFormik({
```
josueetcom, line 31: (For future reference) I'm realizing that the `formik` value maybe doesn't need to be passed in as a property. We could just pass in a **ScholarshipModel** that comes from either **Scholarships.id()** or **Scholarships.new()** and get the default/initial form values from there.

josueetcom (dismissed):

src/components/ScholarshipForm.jsx
edwingl10: interesting!

josueetcom (approved):

PR #613, opened 2021-09-01
Form general section styling

### LINKED ISSUE #601: General scholarship info styling
Styling for the General section when adding a new scholarship, per the [figma](https://www.figma.com/file/nwrW3UebcvSWyQZGb5ixNr/DreamScholar?node-id=0%3A1) design. 

Includes:
- implementing the [vertical stepper](https://material-ui.com/components/steppers/#vertical-stepper)
- changing current TextFields to fit the design
- desktop design
- mobile design

edwingl10:
## Motivation and Context

Fixes #601

## How Has This Been Tested?
On laptop and mobile.

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-613.onrender.com/scholarships/new.

## Checklist:

### DIFF
--- a/src/components/ScholarshipAmountField.jsx
+++ b/src/components/ScholarshipAmountField.jsx
@@ -1,28 +1,27 @@
 import React from 'react';
 import {
-  FormLabel,
-  Radio,
-  RadioGroup,
-  FormControl,
-  FormControlLabel,
+  InputLabel,
+  Grid,
+  Select,
+  MenuItem,
   FormHelperText,
 } from '@material-ui/core';
 import { makeStyles } from '@material-ui/core/styles';
+import RemoveIcon from '@material-ui/icons/Remove';
 import PropTypes from 'prop-types';
 import AmountType from '../types/AmountType';
 import AmountTextField from './AmountTextField';
 
-const useStyles = makeStyles({
-  amountFieldStyle: {
-    display: 'flex',
-    alignItems: 'center',
+const useStyles = makeStyles((theme) => ({
+  amountSelect: {
+    marginTop: theme.spacing(1),
+    minWidth: 150,
   },
-});
+}));
 
 function ScholarshipAmountField(props) {
   const classes = useStyles();
   const {
-    className,
     helperText,
     amountType,
     minAmount,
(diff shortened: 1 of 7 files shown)


src/components/ScholarshipAmountField.jsx
```diff
@@ -37,65 +41,76 @@ function ScholarshipAmountField(props) {
...
+      value={minAmount || ''}
+      onChange={(e) => {
+        const val = parseInt(e.target.value, 10);
+        updateAmount(val || 0, val || 0);
+      }}
+      disabled={!isFixed}
+    />
   );
-  labels[AmountType.FullTuition] = 'Full Tuition';
+  labels[AmountType.FullTuition] = <AmountTextField disabled />;
```
josueetcom, line 85:
```suggestion
  labels[AmountType.FullTuition] = <TextField disabled />;
```

josueetcom: Or maybe nothing at all?

src/components/ScholarshipAmountField.jsx
```diff
@@ -37,65 +41,76 @@ function ScholarshipAmountField(props) {
...
-        value={maxAmount || ''}
-        onChange={(e) => {
-          const val = parseInt(e.target.value, 10);
-          updateAmount(minAmount, val || 0);
-        }}
-        disabled={!isRange}
-      />
-    </div>
+    <Grid container alignItems="center">
+      <Grid item>
```
josueetcom, line 45: I don't think a `Grid` is necessary here. I'm pretty sure these'll render inline next to each other anyway

src/components/ScholarshipAmountField.jsx
```diff
@@ -37,65 +41,76 @@ function ScholarshipAmountField(props) {
...
+            displayEmpty>
+            <MenuItem value={null}>None</MenuItem>
+            <MenuItem value={AmountType.Unknown}>Unknown</MenuItem>
+            <MenuItem value={AmountType.Fixed}>Fixed</MenuItem>
+            <MenuItem value={AmountType.Range}>Range</MenuItem>
+            <MenuItem value={AmountType.FullTuition}>Full Tuition</MenuItem>
+          </Select>
+        </Grid>
+
+        <Grid item>{labels[amountType]}</Grid>
```
josueetcom, line 106: This'll fail if `amountType` isn't specified (it defaults to `null` at the moment)

src/components/ScholarshipAmountField.jsx
```diff
@@ -37,65 +41,76 @@ function ScholarshipAmountField(props) {
 
   const labels = {};
```
josueetcom, line 42: these aren't really labels btw, they're more like input fields.

src/components/ScholarshipAmountField.jsx
edwingl10: Nice, makes it more simple lol

src/components/ScholarshipAmountField.jsx
```diff
@@ -1,28 +1,32 @@
...
-  amountFieldStyle: {
-    display: 'flex',
-    alignItems: 'center',
+const useStyles = makeStyles((theme) => ({
+  amountSelect: {
+    minWidth: 150,
   },
-});
+  dash: {
+    marginTop: theme.spacing(2),
```
josueetcom, line 20: This is making some assumptions about the size of the text. Instead of this I suggest making this vertically centered:
```suggestion
    height: '100%',
    textAlign: 'center',
```

Got this from [SO](https://stackoverflow.com/a/8865463) and testing it out on the web page inspector for validation

src/components/ScholarshipAmountField.jsx
```diff
@@ -1,28 +1,32 @@
...
+  },
+  inputLabel: {
+    marginBottom: theme.spacing(1),
+  },
```
josueetcom, line 24: Interesting, elsewhere the property `labelStyle` is passed around and this is defined in the `ScholarshipForm` component. I suggest being consistent and either passing the style around everywhere or just assuming a default spacing that'll work well anywhere (e.g., instead of a bottom margin of 2 a top/bottom margin of 1).

src/components/ScholarshipAmountField.jsx
```diff
@@ -104,8 +111,7 @@ ScholarshipAmountField.propTypes = {
   helperText: PropTypes.string,
 };
 ScholarshipAmountField.defaultProps = {
-  className: null,
-  amountType: null,
+  amountType: '',
```
josueetcom, line 114: It's not a string type.
```suggestion
  amountType: undefined,
```

src/components/ScholarshipAmountField.test.jsx
```diff
@@ -1,25 +1,41 @@
...
+    />
   );
+}
 
-  expect(screen.getAllByRole('radio')).toHaveLength(4);
-  labels.forEach((name) =>
-    expect(screen.getByRole('radio', { name })).toBeInTheDocument()
-  );
+test('renders select options', () => {
+  renderWithAmountType('FIXED');
```
josueetcom, line 17: Use the [AmountType](https://github.com/beyondhb1079/s4us/blob/main/src/types/AmountType.ts) constant here and elsewhere.

src/components/ScholarshipForm.jsx
```diff
@@ -78,56 +90,111 @@ function ScholarshipForm({ setSubmissionAlert }) {
...
+          id="description"
+          labelStyle={classes.inputLabel}
+          formik={formik}
+          rows={8}
+        />
+      </Grid>
+    </Grid>
+  );
+
+  stepperItems.Requirements = 'requirements';
```
josueetcom, line 171: I'd remove this from the stepper items for now as it's not doing anything. You could add a **Confirmation** step though (maybe in another PR) where you pass the **Scholarship** object to the **ScholarshipDetailCard** component and have the user review/confirm details

src/components/ScholarshipForm.jsx
```diff
@@ -78,56 +90,111 @@ function ScholarshipForm({ setSubmissionAlert }) {
...
+                  disabled={activeStep === 0}
+                  onClick={() => setActiveStep((prevStep) => prevStep - 1)}>
+                  BACK
+                </Button>
+                <Button
+                  variant="contained"
+                  color="primary"
+                  onClick={() => setActiveStep((prevStep) => prevStep + 1)}>
+                  NEXT
+                </Button>
```
josueetcom, line 192: nit: I suggest making this the **SUBMIT** button when `activeStep == Object.keys(stepperItems).length` (or `key == 'Confirmation'`) so that it isn't possible to submit the form until the last step is shown.

josueetcom (approved):
Just a couple of things otherwise looks pretty good

src/components/ScholarshipForm.jsx
```diff
@@ -78,56 +90,111 @@ function ScholarshipForm({ setSubmissionAlert }) {
...
+              <div className={classes.stepperBtns}>
+                <Button
+                  disabled={activeStep === 0}
+                  onClick={() => setActiveStep((prevStep) => prevStep - 1)}>
+                  BACK
+                </Button>
+                <Button
+                  variant="contained"
+                  color="primary"
+                  onClick={() => setActiveStep((prevStep) => prevStep + 1)}>
```
josueetcom, line 190: Something I find concerning about this is how the **Next** button can be clicked without the user knowing there are errors in the step they just completed. You should fix this in a follow-up PR

src/components/ScholarshipAmountField.jsx
edwingl10: For now I passed in the style through props. I am wondering if we should just include that style in the different components instead of passing it, since it's just the styling for the label, I don't think it will change.

src/components/ScholarshipForm.jsx
edwingl10: for sure, I will address this in another PR.

src/components/ScholarshipForm.jsx
edwingl10: Nice, for now there is no back button when the submit button appears but I will fix this in another PR when adding the Review stage.

PR #622, opened 2021-09-04
Award amount functionality

### LINKED ISSUE #621: Award Amount Functionality
Add functionality to the Award Amount Select to support the `varies` option. A user should select varies if the amount is a range or if the amount is unknown. 
The `varies` option should accept either / none of the fields being filled and both of the fields filled. 

Things to change:
- formik validation rules
- some login in `ScholarshipAmount.ts` file

edwingl10:
## Motivation and Context

Fixes #621
Combined `Range` and `Unknown` into a new `Varies` field. Acts the same as the old `Range` but it also allows min and max to be empty (for unknown amounts).

based on #613 

## How Has This Been Tested?

laptop using Firestore emulator.

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-622.onrender.com/scholarships/new.
2. Select 'Varies' option from the Select
3. Play around with the values

## Checklist:


josueetcom (commented):
Some initial PRs I forgot to leave earlier

src/components/ScholarshipAmountField.jsx
```diff
@@ -1,28 +1,32 @@
...
 import AmountType from '../types/AmountType';
 import AmountTextField from './AmountTextField';
 
-const useStyles = makeStyles({
-  amountFieldStyle: {
-    display: 'flex',
-    alignItems: 'center',
+const useStyles = makeStyles((theme) => ({
+  amountSelect: {
+    minWidth: 150,
```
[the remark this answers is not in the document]
edwingl10: From the Figma design. 
I didn't use autoWidth because I wanted it to be consistent with whatever option the user chose. But I am noticing that you can't really notice the change in width, so `autoWidth` should be fine

src/components/AmountTextField.jsx
```diff
@@ -3,10 +3,9 @@ import { TextField, InputAdornment } from '@material-ui/core';
 import PropTypes from 'prop-types';
 import { makeStyles } from '@material-ui/core/styles';
 
-const useStyles = makeStyles((theme) => ({
+const useStyles = makeStyles(() => ({
   amountField: {
-    margin: theme.spacing(1),
-    maxWidth: 120,
+    maxWidth: 130,
```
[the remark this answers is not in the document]
edwingl10: I eye-balled it based on the Figma design. The TextFields were looking a little long.

Screen Shot 2021-09-10 at 12 20 06 AM

src/components/ScholarshipAmountField.jsx
```diff
@@ -35,45 +35,42 @@ function ScholarshipAmountField(props) {
...
   const error = !!helperText; // no error if helperText empty
 
   const inputFields = {};
-  inputFields[AmountType.Range] = (
+  inputFields[AmountType.Varies] = (
     <>
       <AmountTextField
-        error={isRange && error && (!!maxAmount || minAmount >= maxAmount)}
+        error={
+          error && (minAmount < 0 || (minAmount && minAmount >= maxAmount))
```
josueetcom, line 45: `minAmount < 0` is not possible per constraints in [AmountTextField](https://github.com/beyondhb1079/s4us/blob/main/src/components/AmountTextField.jsx#L23)
```suggestion
          error && (minAmount && minAmount >= maxAmount)
```

src/types/ScholarshipAmount.ts
```diff
@@ -23,9 +23,12 @@ export default class ScholarshipAmount {
     type: AmountType = AmountType.Unknown,
     data?: { min?: number; max?: number }
   ) {
-    this.type = type;
     const [min, max] = [data?.min, data?.max];
 
+    if (type === AmountType.Varies && !min && !max)
+      this.type = AmountType.Unknown;
+    else this.type = type;
```
josueetcom, line 30:
```suggestion
    this.type = type === AmountType.Varies && !min && !max ? AmountType.Unknown : type;
```

josueetcom (approved):

src/types/ScholarshipAmount.ts
edwingl10: Lol I had this at first but changed it cuz I thought the other one was more legible.

src/components/ScholarshipAmountField.jsx
edwingl10: I will pass the formik object to AmountTextField so we can directly know what field has the error.

src/components/ScholarshipAmountField.jsx
```diff
@@ -24,55 +25,61 @@ const useStyles = makeStyles(() => ({
...
           updateAmount(minAmount, val || 0);
         }}
-        disabled={!isRange}
       />
     </>
   );
   inputFields[AmountType.Fixed] = (
     <AmountTextField
-      error={isFixed && error}
+      error={getIn(formik.touched, 'amount.min') && Boolean(minError)}
```
edwingl10, line 77: without this, the field will get highlighted as an error even when the input is correct. I am planning to change when the validation runs to hopefully avoid these checks.

src/components/ScholarshipAmountField.jsx
```diff
@@ -11,40 +11,48 @@ import RemoveIcon from '@material-ui/icons/Remove';
...
+  const maxAmount = formik.values.amount.max;
+  const minError = getIn(formik.errors, 'amount.min');
+  const maxError = getIn(formik.errors, 'amount.max');
 
-  const error = !!helperText; // no error if helperText empty
+  function formatHelperText() {
+    if (!minError && !maxError) return;
+    if (getIn(formik.touched, 'amount.min') && minError) return minError;
+    if (getIn(formik.touche, 'amount.max') && maxError) return maxError;
+  }
```
josueetcom, line 49: Why not simply do:
```suggestion
  let helperText = ""; // or undefined or whatever
  if (getIn(formik.touched, 'amount.min') || getIn(formik.touched, 'amount.max')) {
    helperText = minError || maxError;
  }
```

src/components/ScholarshipAmountField.jsx
```diff
@@ -11,40 +11,48 @@ import RemoveIcon from '@material-ui/icons/Remove';
...
-    onTypeChange,
-    updateAmount,
-    labelStyle,
-  } = props;
+  const { labelStyle, formik } = props;
+
+  function updateAmount(min, max) {
+    formik.setFieldValue('amount.min', min, true);
+    formik.setFieldValue('amount.max', max, true);
+  }
```
josueetcom, line 33: You can inline these below instead (it'll make the onChange handlers look a bit more logical too)

src/components/ScholarshipAmountField.jsx
```diff
@@ -11,40 +11,48 @@ import RemoveIcon from '@material-ui/icons/Remove';
...
+
+  function onTypeChange(e) {
+    formik.setFieldValue('amount.type', e.target.value, true);
+  }
```
josueetcom, line 37: This can be inlined below too

src/components/ScholarshipAmountField.jsx
```diff
@@ -11,40 +11,48 @@ import RemoveIcon from '@material-ui/icons/Remove';
...
+
+  const amountType = formik.values.amount.type;
+  const minAmount = formik.values.amount.min;
+  const maxAmount = formik.values.amount.max;
```
josueetcom, line 41: With Formik doing the validation + suggested changes these are only used in one place so they can be inlined as well

josueetcom (changes requested):
Btw, looks like you need to update this branch so it has the changes in `main`.

src/components/ScholarshipAmountField.jsx
```diff
@@ -27,58 +27,50 @@ function ScholarshipAmountField(props) {
...
       <RemoveIcon className={classes.dash} />
 
       <AmountTextField
         error={Boolean(maxError)}
-        value={maxAmount || ''}
+        value={formik.values.amount.max || ''}
         onChange={(e) => {
           const val = parseInt(e.target.value, 10);
-          updateAmount(minAmount, val || 0);
+          formik.setFieldValue('amount.max', val || 0, true);
```
josueetcom, line 61: 2 things here:
1. It looks like `true` is the default value for `shouldValidate` according to [docs](https://formik.org/docs/api/formik#setfieldvalue-field-string-value-any-shouldvalidate-boolean--void) so it can be left out. If you don't want to validate on any text field change you can explicitly pass `false`. 
2. I just noticed the boolean is a bit confusing so maybe add a comment stating what it is. E.g.
```suggestion
          formik.setFieldValue('amount.max', val || 0, /* shouldValidate = */ true);
```

Same comment above for min and below for type.

josueetcom (approved):

src/components/ScholarshipAmountField.jsx
edwingl10: That is an interesting way of writing comments 🤔. I will make these changes in another PR.

PR #628, opened 2021-09-09
refactor formik initial values

### LINKED ISSUE #600: Refactor initial Formik values
Refactor initial Formik values found in the `ScholarshipForm`.

- One potential approach is to pass in a `ScholarshipModel` from either `Scholarships.id()` or `Scholarships.new()` and get the form values from there
- another approach is to refactor the `initialValues` into another file

edwingl10:
## Motivation and Context

Fixes #600 

## How Has This Been Tested?

laptop and firestore emulator

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-NUMBER.onrender.com
2.

## Checklist:


src/formik/InitialValues.js
```diff
@@ -0,0 +1,12 @@
+const InitialValues = {
```
josueetcom, line 1: [23.2](https://github.com/airbnb/javascript#naming--camelCase): Use camelCase when naming objects, functions, and instances.

src/components/ScholarshipForm.jsx
```diff
@@ -28,17 +29,7 @@ function ScholarshipForm({ setSubmissionAlert }) {
...
-      deadline: null,
-      description: '',
-      amount: {
-        type: null,
-        min: 0,
-        max: 0,
-      },
-      website: '',
-    },
+    initialValues: InitialValues,
```
josueetcom, line 32: Why is this being refactored? The initial values are not specific to Formik, they're specific to the ScholarshipForm. This should either be derivable in this component or passed in as a prop.

josueetcom (changes requested):

src/pages/AddScholarship.jsx
```diff
@@ -1,14 +1,48 @@
...
       <h1>Submit a Scholarship</h1>
-      <ScholarshipForm setSubmissionAlert={setSubmissionAlert} />
+      <ScholarshipForm
+        scholarship={scholarship}
+        submitFn={() =>
+          setSubmissionAlert(
+            <SubmissionAlert
+              id={scholarship.id}
+              name={scholarship.data.name}
+              closeFn={setSubmissionAlert(false)}
```
josueetcom, line 32: This isn't a function.
```suggestion
              closeFn={() => setSubmissionAlert(false)}
```

src/pages/AddScholarship.jsx
```diff
@@ -1,14 +1,48 @@
...
+            />
+          )
+        }
+        onSubmitError={() =>
+          setSubmissionAlert(
+            <Alert severity="error" onClose={() => setSubmissionAlert(null)}>
```
josueetcom, line 38: How come this one sets the submission alert to `null` but the other sets it to `false`? Also I just noticed this property's called `onClose`, I wonder if we should have a similarly named property for the **SubmissionAlert** component instead of `closeFn`.

src/components/ScholarshipForm.jsx
```diff
@@ -24,51 +21,24 @@ const useStyles = makeStyles((theme) => ({
...
       setSubmitting(true);
-      Scholarships.new({
+      scholarship.data = {
         ...values,
-        author: {
-          id: user?.uid,
-          email: user?.email,
-        },
-      })
+        author: { id: user?.uid, email: user?.email },
```
josueetcom, line 35: This is something that no longer belongs here as this form could be used for updating existing scholarships and this would change the author.

edwingl10: I moved it to `AddScholarship`

src/pages/AddScholarship.jsx
edwingl10: oops the latter was wrong, it was suppose to set it to null as well. Although `false` also seems to work, I think `null` is more suitable.

josueetcom (approved):
Approved. Though before merging I suggest getting the base branch merged in so this can be based on main instead

PR #630, opened 2021-09-12
Eligibility Requirements Styling

### LINKED ISSUE #602: Eligibility requirements styling
Styling for the Eligibility requirements section when adding a new scholarship, per the [Figma](https://www.figma.com/file/nwrW3UebcvSWyQZGb5ixNr/DreamScholar?node-id=0%3A1) design.

Includes: 
- desktop styling
- mobile styling

edwingl10:
## Motivation and Context

Fixes #602 

## How Has This Been Tested?

Laptop using FireFox.

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-630.onrender.com/scholarships/new.
2. Click next button to see eligibility section 

## Checklist:


src/components/ScholarshipForm.jsx
```diff
@@ -168,7 +203,90 @@ function ScholarshipForm({ setSubmissionAlert }) {
...
+          label="GPA"
+          id="gpa"
+          formik={formik}
+          labelStyle={classes.inputLabel}
+        />
+      </Grid>
+
+      <Grid item sm={6} xs={12}>
+        <InputLabel className={classes.inputLabel}>School</InputLabel>
+        <Autocomplete
```
josueetcom, line 248: I suggest making these autocompletes [freeSolo](https://mui.com/components/autocomplete/#free-solo) and allow [multiple](https://mui.com/components/autocomplete/#free-solo) values

edwingl10: yes, I was just testing some things.

src/components/FormikMultiSelect.jsx
```diff
@@ -0,0 +1,57 @@
...
+      width: 250,
+    },
+  },
+};
+
+function FormikMultiSelect(props) {
+  const { label, id, labelStyle, formik, options } = props;
+
+  return (
+    <>
```
josueetcom, line 23: You could place these inside a [FormControl](https://v4.mui.com/api/form-control/) like in the [demo](https://v4.mui.com/components/selects/#multiple-select).

src/components/FormikMultiSelect.jsx
```diff
@@ -0,0 +1,57 @@
...
+            `requirements.${id}`,
+            e.target.value,
+            /* shouldValidate = */ false
+          )
+        }
+        MenuProps={MenuProps}>
+        {options.map((name) => (
+          <MenuItem key={name} value={name}>
+            {name}
+          </MenuItem>
```
josueetcom, line 42: Optional: Have you considered using checkboxes? Or maybe letting the user decide how to display these? It seems like we can add anything here so in a later PR for something like grade level, let users check category check boxes for "All high schoolers", "All college undergraduates", "All college graduates".

Again, this is optional so maybe leave for a later PR.

src/components/FormikMultiSelect.jsx
```diff
@@ -0,0 +1,57 @@
...
+  },
+};
+
+function FormikMultiSelect(props) {
+  const { label, id, labelStyle, formik, options } = props;
+
+  return (
+    <>
+      <InputLabel className={labelStyle}>{label}</InputLabel>
+      <Select
```
josueetcom, line 25: Optional: Have you considered setting the `renderValue` so that chips are displayed? This'd help show all the selected options.

Alternatively you may want the caller to provide a custom `renderValue` instead. E.g. maybe for grade level, if all High school grades are selected we display "All high schoolers" instead of "HS Freshman, HS Sophomore, ....". 

Again, this is optional so maybe leave for a later PR.

src/components/FormikAutocomplete.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { InputLabel, TextField } from '@material-ui/core';
+import { Autocomplete } from '@material-ui/lab';
+import PropTypes from 'prop-types';
+
+/* eslint-disable react/jsx-props-no-spreading */
+function FormikAutocomplete(props) {
```
josueetcom, line 7: It'd be nice if this accepted a placeholder value that indicates any/all values apply. E.g. For the Majors one it could say "All/Any majors acceptable" or "No major requirements"

src/components/FormikMultiSelect.jsx
```diff
@@ -0,0 +1,57 @@
...
+  },
+  PaperProps: {
+    style: {
+      maxHeight: 250,
+      width: 250,
+    },
+  },
+};
+
+function FormikMultiSelect(props) {
```
josueetcom, line 19: Similar to above a placeholder explaining what no selection means would be nice. E.g. "No ethnicity requirements", etc.

src/components/ScholarshipForm.jsx
```diff
@@ -8,13 +8,42 @@ import {
...
+  'Asian',
+  'Black or African American',
+  'Hispanic or Latino',
+  'Native Hawaiian or Other Pacific Islander',
+  'White',
+];
+
+const schoolOptions = [{ title: 'school1' }, { title: 'school2' }];
+const stateOptions = [{ title: 'CA' }, { title: 'WA' }];
+const majorOptions = [{ title: 'major1' }, { title: 'major2' }];
```
josueetcom, line 46: Why not just use arrays here?

src/components/ScholarshipForm.jsx
```diff
@@ -114,6 +145,90 @@ function ScholarshipForm({ scholarship, submitFn, onSubmitError }) {
...
+        <FormikAutocomplete
+          label="State(s)"
+          id="states"
+          labelStyle={classes.inputLabel}
+          options={stateOptions}
+          formik={formik}
+        />
+      </Grid>
+
+      <Grid item xs={6}>
```
josueetcom, line 209:
```suggestion
      <Grid item sm={6} xs={12}>
```

src/components/ScholarshipForm.jsx
```diff
@@ -114,6 +145,90 @@ function ScholarshipForm({ scholarship, submitFn, onSubmitError }) {
...
+          label="Major(s)"
+          id="majors"
+          labelStyle={classes.inputLabel}
+          options={majorOptions}
+          freeSolo
+          formik={formik}
+        />
+      </Grid>
+
+      <Grid item xs={6}>
```
josueetcom, line 220:
```suggestion
      <Grid item sm={6} xs={12}>
```

josueetcom (changes requested):

src/components/ScholarshipForm.jsx
```diff
@@ -114,6 +145,90 @@ function ScholarshipForm({ scholarship, submitFn, onSubmitError }) {
     </Grid>
   );
 
+  stepperItems['Eligibility Requirements'] = (
```
josueetcom, line 148: Please guard this with an experiment. Or at least the grades and ethnicities ones. We haven't quite decided the data model for how to store these.

src/components/FormikMultiSelect.jsx
```diff
@@ -0,0 +1,57 @@
...
+          formik.setFieldValue(
+            `requirements.${id}`,
+            e.target.value,
+            /* shouldValidate = */ false
+          )
+        }
+        MenuProps={MenuProps}>
+        {options.map((name) => (
+          <MenuItem key={name} value={name}>
+            {name}
```
josueetcom, line 41: Something to consider here is that the value may be different from the display name. E.g. "High School Senior" may map to the value `GradeLevel.HS_SENIOR` which under the hood may be the value `12`. I suggest thinking about how to support this use case. You might want to make `options` a dictionary object and do something like:
```suggestion
        {options.map(([name, value]) => (
          <MenuItem key={value} value={value}>
            {name}
```

src/components/ScholarshipForm.jsx
josueetcom: In retrospect this is the sort of pattern you may want for the gradeOptions and ethnicityOptions instead since those could be displayed differently than their values.

src/components/ScholarshipForm.jsx
josueetcom: Actually, you may want to guard the whole thing (or most of the fields) for now as some of the options are limited (e.g. the state dropdown doesn't have all 50 states).

Or you could add to `constants.ts` some of those as lists:
```ts
// Source: https://gist.github.com/bubblerun/a624de5b4fa8ff0980010054a7220977
const STATES = [...];

// Source: https://www.nomajordrama.co.nz/Majors
const MAJORS = [...];

// Source: https://www.university-list.net/us/universities-2000.htm
const COLLEGES = [...];
```
Feel free to use different sources and scrape things however you like (e.g. you could include the state for the college, the major's general area so that they're groupable and more easily selectable by general area, etc.)

Alternatively you could break up this PR into multiple ones and add a single field or a group of similar fields for each if you'd like.

src/components/FormikMultiSelect.jsx
edwingl10: Interesting, I will keep this in mind for another PR.

src/components/FormikMultiSelect.jsx
edwingl10: yeah for sure I can explore this in another PR. The only thing with chips though, is that they might enlarge the Select field and that might look a little weird.

josueetcom: Yeah enlarging the Select field may look a bit weird. But maybe that's ok. On the other hand, maybe in some cases displaying a [limited number of chips](https://mui.com/components/autocomplete/#limit-tags) makes more sense. Maybe a custom `renderValue` prop would be good.

src/components/FormikAutocomplete.jsx
edwingl10: added!

src/components/FormikMultiSelect.jsx
edwingl10: I tried it, but it messes up the styling. I can try to fix this in another PR.

Screen Shot 2021-10-01 at 10 56 25 AM

src/types/Ethnicity.ts
```diff
@@ -0,0 +1,10 @@
+enum Ethnicity {
```
josueetcom, line 1: One more is Mixed/Two or more races

src/components/ScholarshipForm.jsx
```diff
@@ -114,6 +146,96 @@ function ScholarshipForm({ scholarship, submitFn, onSubmitError }) {
...
+            placeholder="No state requirements"
+          />
+        </Grid>
+
+        <Grid item sm={6} xs={12}>
+          <FormikAutocomplete
+            label="Major(s)"
+            id="majors"
+            labelStyle={classes.inputLabel}
+            options={[...MAJORS]}
```
josueetcom, line 219: Why not simply 
```suggestion
            options={MAJORS}
```

src/components/ScholarshipForm.jsx
```diff
@@ -114,6 +146,96 @@ function ScholarshipForm({ scholarship, submitFn, onSubmitError }) {
...
+            labelStyle={classes.inputLabel}
+          />
+        </Grid>
+
+        <Grid item sm={6} xs={12}>
+          <FormikAutocomplete
+            label="School(s)"
+            id="schools"
+            labelStyle={classes.inputLabel}
+            options={[...SCHOOLS]}
```
josueetcom, line 196: Why not simply

```suggestion
            options={SCHOOLS}
```

src/components/FormikMultiSelect.jsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export default FormikMultiSelect;
+
+FormikMultiSelect.propTypes = {
+  label: PropTypes.string.isRequired,
+  id: PropTypes.string.isRequired,
+  labelStyle: PropTypes.string.isRequired,
+  formik: PropTypes.object.isRequired,
+  options: PropTypes.arrayOf(PropTypes.string).isRequired,
```
josueetcom, line 71:
```suggestion
  options: PropTypes.objectOf(PropTypes.string).isRequired,
```

src/components/FormikMultiSelect.jsx
```diff
@@ -0,0 +1,76 @@
...
+        value={formik.values.requirements[id]}
+        onChange={(e) =>
+          formik.setFieldValue(
+            `requirements.${id}`,
+            e.target.value,
+            /* shouldValidate = */ false
+          )
+        }
+        renderValue={(selected) =>
+          selected.length === 0 ? placeholder : selected.join(', ')
```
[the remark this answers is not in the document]
josueetcom: Also noticed this renders the all caps and underscored names instead. Since the section is hidden I recommend fixing that in a follow-up PR.

src/types/GradeLevel.ts
```diff
@@ -0,0 +1,13 @@
+enum GradeLevel {
```
josueetcom, line 1: I was actually thinking of using numbers here but we can fix that in a later PR since this is hidden functionality anyways.

src/components/ScholarshipForm.jsx
```diff
@@ -8,13 +8,43 @@ import {
...
 import FormikTextField from './FormikTextField';
+import FormikMultiSelect from './FormikMultiSelect';
+import FormikAutocomplete from './FormikAutocomplete';
+import { SCHOOLS, STATES, MAJORS } from '../types/options';
+import GradeLevel from '../types/GradeLevel';
+import Ethnicity from '../types/Ethnicity';
+import experiments from '../lib/experiments';
+
+const gradeOptions = {
+  'Middle School': GradeLevel.MiddleSchool,
```
josueetcom, line 28: (for a follow-up PR)

I recommend having these string representations be fetched from a util class method or a method in the same namespace:
https://stackoverflow.com/a/28151986

Then you could dynamically generate this list.

josueetcom (approved):

src/components/FormikAutocomplete.jsx
```diff
@@ -0,0 +1,57 @@
...
+          )
+        }
+        options={options}
+        renderInput={(params) => (
+          <TextField
+            {...params}
+            variant="outlined"
+            fullWidth
+            placeholder={
+              formik.values.requirements[id].length == 0 ? placeholder : ''
```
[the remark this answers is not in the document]
edwingl10: Screen Shot 2021-10-02 at 10 50 35 PM

unfortunately, it will keep the placeholder even if the TextField is not empty 🙃

src/components/ScholarshipForm.jsx
edwingl10: interesting 🤔

src/components/ScholarshipForm.jsx
edwingl10: it throws the following error: `Uncaught TypeError: options.filter is not a function`. Probably because the `SCHOOLS` is a set

src/components/ScholarshipForm.jsx
edwingl10: same as the SCHOOLS 🙃

src/types/GradeLevel.ts
edwingl10: so for the middle school option: would it be '6-8'? or simply 8.

src/components/FormikAutocomplete.jsx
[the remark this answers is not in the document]
josueetcom: Ah darn, that's unfortunate. Let's keep it as is then. Maybe they fixed in MUIv5 and we can revisit this later.

src/components/ScholarshipForm.jsx
josueetcom: Got it.

src/components/ScholarshipForm.jsx
josueetcom: Got it.

src/types/GradeLevel.ts
josueetcom: I was originally thinking 8. Though this is actually a good point to discuss. You could write a one-pager with a proposal for how to represent the various grade levels. I just realized you might have cases of 5th year high schoolers, 5th year undergrad students, etc. that might not fit well with the standard grade level-ing system.

josueetcom (approved):

PR #634, opened 2021-09-25
no requirements checkbox functionality

edwingl10:
## Motivation and Context

Added functionality to the `No Requirements` checkbox. When clicked, it disables all the fields. It keeps values entered before checking the box but doesn't submit those values with the scholarship. 

## How Has This Been Tested?

Laptop using localhost. 

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-NUMBER.onrender.com/scholarships/new.
2. Click on the next button to access the next section
3. check the checkbox and notice that all the other fields in that section get disabled. 

## Checklist:

### DIFF
--- a/src/components/ScholarshipForm.jsx
+++ b/src/components/ScholarshipForm.jsx
@@ -60,7 +60,7 @@ const useStyles = makeStyles((theme) => ({
 function ScholarshipForm({ scholarship, submitFn, onSubmitError }) {
   const classes = useStyles();
   const [activeStep, setActiveStep] = useState(0);
-  const [hasReqs, setHasReqs] = useState(false);
+  const [noReqsChecked, setNoReqsChecked] = useState(false);
 
   const formik = useFormik({
     initialValues: scholarship.data,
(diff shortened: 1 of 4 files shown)


src/components/ScholarshipForm.jsx
```diff
@@ -60,14 +61,24 @@ const useStyles = makeStyles((theme) => ({
...
     validationSchema,
     onSubmit: (values, { setSubmitting, resetForm }) => {
       setSubmitting(true);
       scholarship.data = { ...values };
+
+      if (noReqsChecked) {
+        Object.keys(scholarship.data.requirements).forEach(
+          (k) => (scholarship.data.requirements[k] = [])
+        );
+        scholarship.data.requirements.gpa = 0;
```
edwingl10, line 79: If the user chooses some requirements and then checks the 'No requirements checkbox', this will 'save' their choices, but not submit them to the scholarship.

Another thing we could do is clear any selections they made before checking the checkbox. Not sure what route we wanna take.

PR #677, opened 2021-10-05
ignore expired scholarships.

Jobhdez:
## Motivation and Context

the issue was that the scholarships past the deadline were showing up. this is a fix.

## Testing
I created two scholarships. One that was past the deadline and one that wasnt. the one that was passed the deadling didnt show up.

### DIFF
--- a/src/models/Scholarships.ts
+++ b/src/models/Scholarships.ts
@@ -70,7 +70,9 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
     }
     // Sort ties by deadline earliest to earliest
     if (opts.sortField !== 'deadline') {
-      query = query.orderBy('deadline', 'asc');
+      query = query
+        .where('deadline', '>', new Date())
+        .orderBy('deadline', 'asc');
     }
     if (opts.authorId) {
       query = query.where('author.id', '==', opts.authorId);


src/models/Scholarships.ts
```diff
@@ -70,7 +70,9 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
     }
     // Sort ties by deadline earliest to earliest
     if (opts.sortField !== 'deadline') {
-      query = query.orderBy('deadline', 'asc');
+      query = query
+        .where('deadline', '>', new Date())
```
josueetcom, line 74: 3 issues with this:
1. This is inflexible. The caller has no option to show expired scholarships, which would be relevant in other use cases.
2. This is only executed if you're not sorting by deadline.
3. You have no tests to prove the correctness of this. You should add test cases to verify this is correct.

josueetcom (changes requested):

src/models/Scholarships.ts
josueetcom: One more issue based on the failed tests but isn't obvious: Firestore doesn't let you actually do this unless the sorting field is `deadline` so you'll need to handle both cases.

PR #682, opened 2021-10-08
Igore scholarhips past deadline

ghost:
i tested my changes by adding scholarhips and seeing if the scholarships were ignored which they were.


josueetcom:
superseded #682

PR #704, opened 2021-10-24
Edit scholarship

### LINKED ISSUE #558: Allow users to edit submitted scholarships
Users need to be edit/delete from the home page and also from the scholarship page, but only if they are the author of the scholarship.

edwingl10:
## Motivation and Context

Fixes #558 

## How Has This Been Tested?

local

## Types of changes

## Previewing Changes

2. click on a scholarship you added
3. click the edit button

## Checklist:


firestore.rules
```diff
@@ -6,6 +6,7 @@ service cloud.firestore {
     }
     match /scholarships/{scholarshipId} {
       allow write: if request.auth != null;
+      allow update: if request.auth.uid == resource.data.atuhor.id;
```
josueetcom, line 9:
```suggestion
      allow update: if request.auth.uid == resource.data.author.id;
```

Please add a test for this.

src/pages/EditScholarship.jsx
```diff
@@ -0,0 +1,78 @@
...
+    if (loading) {
+      Scholarships.id(id)
+        .get()
+        .then((s) => mounted && setScholarship(s))
+        .catch((e) => mounted && setError(e));
+    }
+    return () => {
+      mounted = false;
+    };
+  }, [id, loading]);
```
josueetcom, line 29: "if we need to load it" and the `if (loading)` checks come from `ViewScholarship` relying on passed in state. I think we can keep this simple and remove the effect's dependence on `loading`.

src/pages/EditScholarship.jsx
```diff
@@ -0,0 +1,78 @@
...
+import ScholarshipForm from '../components/ScholarshipForm';
+import SubmissionAlert from '../components/SubmissionAlert';
+import Scholarships from '../models/Scholarships';
+import { Helmet } from 'react-helmet';
+
+function EditScholarship() {
+  const { id } = useParams();
+  const [scholarship, setScholarship] = useState(undefined);
+  const [error, setError] = useState();
+  const loading = !error && (!scholarship || !scholarship.data);
```
josueetcom, line 14: I'm curious where the `!scholarship.data` comes from?

src/components/ProtectedRoute.jsx
```diff
@@ -42,7 +42,7 @@ function ProtectedRoute({ component: Component, path }) {
           );
         }
         if (isSignedIn === true) {
-          return <Component />;
+          return <Component {...props} />;
```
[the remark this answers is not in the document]
josueetcom: Sweet! Also I just realized you probably got those react router params from:
https://github.com/beyondhb1079/s4us/blob/eb8ba2ffe3a1b6de8f32c9135a39c1164c024a3b/src/pages/ViewScholarship.jsx#L8

I guess **ProtectedRoute** behaves a little differently from normal **Route**. I wonder if there's a pattern we should be using so these router props are passed automatically or if this is the way 🤔

firestore.rules
edwingl10: oops! typos 😶
Added the tests!

src/pages/EditScholarship.jsx
edwingl10: Was most likely a check when we were passing scholarship in state.

src/pages/EditScholarship.jsx
edwingl10: nice, I thought we still needed it to show the '...Loading' screen but we can also show it when `error || !scholarship`

src/firestore.rules.test.js
```diff
@@ -56,3 +61,38 @@ test('allows write to scholarships doc when signed in', () => {
...
+
+  testDoc.then((s) => console.log(s));
+
+  const updatedDoc = authedApp2
+    .firestore()
+    .collection('scholarships')
+    .doc('KLJASDQW')
+    .set({ ...testDoc, name: 'updated name' });
+
+  return assertFails(updatedDoc);
```
edwingl10, line 97: I'm trying to figure out why this succeeds when it's suppose to fail 🤔

src/firestore.rules.test.js
```diff
@@ -56,3 +61,38 @@ test('allows write to scholarships doc when signed in', () => {
...
+    .doc('KLJASDQW')
+    .set(newScholarship);
+
+  const updatedDoc = authedApp
+    .firestore()
+    .collection('scholarships')
+    .doc('KLJASDQW')
+    .set({ ...testDoc, name: 'updated name' });
+
+  return assertSucceeds(updatedDoc);
```
josueetcom, line 78: `.set` returns a Promise you need to wait on to finish before you can update it.
```js
  return assertSucceeds(authedApp
    .firestore()
    .collection('scholarships')
    .doc('KLJASDQW')
    .set(newScholarship)
    .then((s) => s.set({ ...newScholarship, name: 'updated name' });
```

This might be what's causing the bug below too.

src/firestore.rules.test.js
josueetcom: My guess is that (like in the other test), you're not waiting for the `.set()` promise to complete so perhaps there's a race condition going on and the second promise succeeds because it checks against the `create` rule instead of the `update` one.

Try something like this:

```js
  await assertSucceeds(authedApp
    .firestore()
    .collection('scholarships')
    .doc('KLJASDQW')
    .set(newScholarship));

  return assertFails(authedApp2
    .firestore()
    .collection('scholarships')
    .doc('KLJASDQW')
    .set({ ...newScholarship, name: 'updated name' }));
```

src/firestore.rules.test.js
```diff
@@ -26,6 +26,21 @@ const authedApp = initializeTestApp({
...
+const converter = {
+  toFirestore: (data) => {
+    const user = { uid: 'alice', email: 'test' };
+    const author =
+      data.author ?? (user ? { id: user?.uid, email: user?.email } : {});
+
+    console.log({ ...data, author });
+    return { ...data, author };
+  },
+};
```
edwingl10, line 43: just testing some things, but with this at least the user is added lol

josueetcom: I suggest debugging what `firebase.app()` is as that's what gets used in the normal converter:
https://github.com/beyondhb1079/s4us/blob/eb8ba2ffe3a1b6de8f32c9135a39c1164c024a3b/src/models/Scholarships.ts#L12

Maybe add `console.log(JSON.stringify())` statements or set a debug breakpoint? (I've yet to figure out how to run the debugger for tests but I'm sure there's a way).

josueetcom: Ah wait, of course it's like this! None of the tests below are using the converter defined in `Scholarship.ts`:
https://github.com/beyondhb1079/s4us/blob/eb8ba2ffe3a1b6de8f32c9135a39c1164c024a3b/src/models/Scholarships.ts#L9

They're all just simply doing doing `firestore().collection('scholarships')` instead of what `FirestoreCollection.collection` does:
https://github.com/beyondhb1079/s4us/blob/eb8ba2ffe3a1b6de8f32c9135a39c1164c024a3b/src/models/base/FirestoreCollection.ts#L10-L15

Make sure to pass `.withConverter(Scholarships.converter)` everywhere you need to rely on converter logic.

josueetcom: Actually, you know what, it might still not work because there are _multiple_ apps and `Scholarships.ts` always uses the _default_ one (which I'm not sure how it's determined). I think we need to separate the Firebase app from these collection classes so that we can have multiple handles using different apps. Ideally you'd be able to do something like:
```js
const app1Scholarships = Scholarships.connect(app1);
const app2Scholarships = Scholarships.connect(app2);

app1Scholarships.new()...
app2Scholarships.new()...
```

Feel free to skip getting the second test to work (you can comment it out with a TODO) and file an issue to support specifying the Firestore connection for our Firestore collections.

josueetcom (approved):

PR #717, opened 2021-11-02
ScholarshipForm improvements

josueetcom:
- Do not assume a particular path for Formik input types
- Initialize new scholarships with required field defaults
- Support unset optional fields in the form.

## Motivation and Context

Helps isolate better the responsibilities of the page component vs. the form component. 

## How Has This Been Tested?

Locally

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-717.onrender.com.
1. Log in.
1. Try adding or editing a scholarship you own.

## Checklist:

### DIFF
--- a/src/components/ScholarshipForm.jsx
+++ b/src/components/ScholarshipForm.jsx
@@ -58,11 +58,14 @@ function ScholarshipForm({ scholarship, onSubmit }) {
       scholarship
         .save()
         .then(onSubmit)
+        .then(() => setActiveStep(0))
         .catch(setSubmissionError)
         .finally(() => setSubmitting(false));
     },
   });
 
+  // Initially requirements is null but is set to {} when the "no requirements"
+  // checkbox is explicitly set.
   const noReqsChecked = JSON.stringify(formik.values.requirements) === '{}';
 
   const stepperItems = {};
--- a/src/components/ScholarshipForm.jsx
+++ b/src/components/ScholarshipForm.jsx
@@ -231,9 +231,11 @@ function ScholarshipForm({ scholarship, onSubmit }) {
   );
 
   function validationCheck() {
-    const noReqsGiven = Object.values(formik.values.requirements).every(
-      (val) => val == [] || val == ''
-    );
+    const noReqsGiven =
+      !formik.values.requirements ||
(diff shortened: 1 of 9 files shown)


src/components/ScholarshipForm.jsx
```diff
@@ -41,29 +43,28 @@ const useStyles = makeStyles((theme) => ({
...
-        .then(() => setNoReqsChecked(false))
-        .then(resetForm)
-        .catch(onSubmitError)
+        .then(onSubmit)
+        .catch(setSubmissionError)
         .finally(() => setSubmitting(false));
     },
   });
 
+  const noReqsChecked = JSON.stringify(formik.values.requirements) === '{}';
```
edwingl10, line 66: Nice! 
I was curious why `JSON.stringify` was needed but I am guessing so we can make this check easier.

src/components/ScholarshipForm.jsx
```diff
@@ -41,29 +43,28 @@ const useStyles = makeStyles((theme) => ({
...
       setSubmitting(true);
       scholarship.data = { ...values };
       scholarship
         .save()
-        .then(submitFn)
-        .then(() => setActiveStep(0))
-        .then(() => setNoReqsChecked(false))
-        .then(resetForm)
-        .catch(onSubmitError)
+        .then(onSubmit)
```
edwingl10, line 60: right now after submission / edit, it doesn't reset the `ActiveStep`. 
If you spam the submit button after adding a new scholarship, it will make duplicates. 

Since we will fix this later with a dialog, for now I think we can disable the buttons, forcing them to either view the scholarship or go to another page. OR just take them back to the beginning of the form.

edwingl10 (approved):

src/components/ScholarshipForm.jsx
josueetcom: Oh weird, I thought changing the `scholarship` prop of the form would prompt a recreation of this component. I'll have it go back to the beginning of the form for now.

src/components/ScholarshipForm.jsx
josueetcom: Yepp. Added a comment to help explain it since it's not intuitive at first glance.

PR #720, opened 2021-11-03
ScholarshipAmount bug fixes

### LINKED ISSUE #719: ScholarshipAmount bugs
While editing a scholarship, noticed the following. 

1. stored values (intended for backend) show when editing a scholarship. E.g. when editing a scholarship that had an amount type set to `Full Tuition`, and changing it to `varies` shows. 

Screen Shot 2021-11-02 at 8 41 22 PM

2. amount values from a previous selection carry on to new selection. E.g. when first choosing the `Fixed` type amount for a scholarship and then changing it to `Varies`, min and max are set to fixed amount you inputted at first.
Screen Shot 2021-11-02 at 8 41 48 PM

3. `toString` shows wrong text when adding a new scholarship and only setting the min. 
Screen Shot 2021-11-02 at 8 42 28 PM

edwingl10:
## Motivation and Context

Fixes #719.
Adds a `fromStorage` method. Similar to the `toStorage` method we already have.
Resets `min` and `max` when the scholarship amount `type` is changed. 

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-720.onrender.com/scholarships.
2. Edit a scholarship and you should not see values intended for backend
3. change the amount types and notice the min and max resetting

## Checklist:


src/components/ScholarshipAmountField.jsx
```diff
@@ -79,7 +79,11 @@ function ScholarshipAmountField(props) {
             className={classes.amountSelect}
             variant="outlined"
             value={amountType}
-            onChange={formik.handleChange}>
+            onChange={(e) => {
+              formik.setFieldValue('amount.min', 0);
+              formik.setFieldValue('amount.max', 0);
+              formik.handleChange('amount.type')(e);
```
edwingl10, line 85: got this from [here.](https://stackoverflow.com/questions/61384089/how-to-call-two-function-in-onchange-using-formik)

src/types/ScholarshipAmount.ts
```diff
@@ -108,6 +108,24 @@ namespace ScholarshipAmount {
     return { min, max, type };
   }
 
+  // Translates 'amount' to the format it should be displayed as
```
josueetcom, line 111: Use JSDoc style as in the previous function (that way IDEs will display this comment when you hover over the `fromStorage` method).

src/types/ScholarshipAmount.ts
```diff
@@ -117,7 +135,7 @@ namespace ScholarshipAmount {
         return `$${amount.min}`;
       case AmountType.Varies:
         if (!amount.min && !amount.max) return 'Varies';
-        if (amount.min && amount.max !== RANGE_MAX) {
+        if (amount.min && amount.max && amount.max !== RANGE_MAX) {
```
josueetcom, line 138: Can you add a test case for this?

src/types/ScholarshipAmount.ts
```diff
@@ -108,6 +108,24 @@ namespace ScholarshipAmount {
     return { min, max, type };
   }
 
+  // Translates 'amount' to the format it should be displayed as
+  export function fromStorage(amount: ScholarshipAmount): ScholarshipAmount {
```
josueetcom, line 112: I don't exactly recall off the top of my head why I decided not to include this method when I added the other one. I'm guessing it has to do something with the fact that storing it again with the weird values should be ok and `toString` should be able to handle the weird values. Though I guess certain elements need the raw values and the UI would need to sort of do the same kind of handling this method does.

josueetcom: We should add test cases for `toStorage` and this one 🤔.

src/types/ScholarshipAmount.test.ts
```diff
@@ -134,3 +150,110 @@ test('toString() - Range - unbounded', () => {
...
+  expect(res).toEqual({
+    type: AmountType.Unknown,
+    min: UNKNOWN_MIN,
+    max: UNKNOWN_MAX,
+  });
+});
+
+test('fromStorage() - Fixed', () => {
+  const amount = fixed(1000);
+  const res = fromStorage(amount);
```
josueetcom, line 217: For the `fromStorage()` tests you might want to do something like this instead to make sure you're converting from the storage represenation:
```suggestion
  const res = fromStorage(toStorage(amount));
```

josueetcom (dismissed):

josueetcom (approved):

PR #722, opened 2021-11-05
Add VS Code Chrome and test debugging targets

josueetcom:
## Motivation and Context

Adds debugging targets everyone can use.

## How Has This Been Tested?

Locally.


edwingl10 (approved):
looks good!

PR #726, opened 2021-11-13
Grade filter param support

### LINKED ISSUE #275: Grade level filter: URL parameter support
URL parameters usually encode information about filters to make pages easier to share. E.g. compare the `q` and `start` parameter values with what's presented in the query and the page number here: https://www.google.com/search?q=url+parameters+filter&start=10
 
- [x] Update URL parameters when grade level filter is applied
- [x] Parse URL parameters to determine grade level filters
- [x] Ensure grade level filter component reflects URL parameters
- [x] Ensure fetched data reflects filters set via the URL

edwingl10:
## Motivation and Context

Fixes #275
prunes duplicate grade values, and removes grade values that are not represented by the `GradeLevel` enum.

## How Has This Been Tested?

local

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-726.onrender.com/scholarships?expShowGradeFilter=1.
2. click some grades and watch the parameters change with it
3. you can also provide some grades in the url and see the changes reflected in the filter
4. try inputting some invalid grade values, should prune those values

## Checklist:

### DIFF
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -40,7 +40,7 @@ export default function FilterBar(props) {
         {experiments.expShowGradeFilter && (
           <GradeLevelFilter
             grades={grades ?? []}
-            changeFn={(e) => setQueryParam('grades', e)}
+            changeFn={(e) => setQueryParam(qParams.GRADES, e)}
           />
         )}
         <AmountFilter
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -39,7 +39,7 @@ export default function FilterBar(props) {
         )}
         {experiments.expShowGradeFilter && (
           <GradeLevelFilter
-            grades={grades ?? []}
+            grades={new Set(grades)}
             changeFn={(e) => setQueryParam(qParams.GRADES, e)}
           />
         )}
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -29,7 +29,7 @@ const majors = {
 export default function FilterBar(props) {
   const { setQueryParam, queryParams } = props;
   const classes = useStyles();
-  const { minAmount, maxAmount } = queryParams;
+  const { minAmount, maxAmount, grades } = queryParams;
 
   return (
     <Grid container spacing={2} className={classes.root}>
@@ -37,7 +37,12 @@ export default function FilterBar(props) {
         {experiments.expShowMajorFilter && (
           <FilterDropDown label="Major" items={majors} />
         )}
-        {experiments.expShowGradeFilter && <GradeLevelFilter />}
+        {experiments.expShowGradeFilter && (
+          <GradeLevelFilter
+            grades={grades ?? []}
+            changeFn={(e) => setQueryParam('grades', e)}
+          />
+        )}
         <AmountFilter
           min={minAmount ?? 0}
           max={maxAmount ?? 0}
--- a/src/components/FormikMultiSelect.jsx
+++ b/src/components/FormikMultiSelect.jsx
@@ -43,7 +43,7 @@ function FormikMultiSelect(props) {
         }
         MenuProps={MenuProps}>
         {Object.entries(options).map(([val, stringRep]) => (
-          <MenuItem key={val} value={val}>
+          <MenuItem key={val} value={parseInt(val) || val}>
             {stringRep}
           </MenuItem>
         ))}
--- a/src/components/GradeLevelFilter.jsx
+++ b/src/components/GradeLevelFilter.jsx
@@ -21,20 +21,23 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-export default function GradeLevelFilter() {
+export default function GradeLevelFilter(props) {
   const classes = useStyles();
   const [anchorEl, setAnchorEl] = useState(null);
   const openPopover = (event) => setAnchorEl(event.currentTarget);
-  const [selectedGrades, setSelectedGrades] = useState(new Set());
+
+  const { grades, changeFn } = props;
 
   function toggleSelection(grade) {
-    const newSet = new Set(selectedGrades);
-    if (selectedGrades.has(grade)) {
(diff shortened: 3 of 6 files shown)


src/components/FormikMultiSelect.jsx
```diff
@@ -43,7 +43,7 @@ function FormikMultiSelect(props) {
         }
         MenuProps={MenuProps}>
         {Object.entries(options).map(([val, stringRep]) => (
-          <MenuItem key={val} value={val}>
+          <MenuItem key={val} value={parseInt(val) || val}>
```
edwingl10, line 46: the grades were being stored as strings instead of integers. 
Instead of doing this, I can use `GradeLevel.keys()` and then use `GradeLevel.toString(key)` for the value. 
I would also have to add the `keys()` method for the `Ethnicity` type but we would avoid parsing the integer here and in other places.

edwingl10, line 46: grades were being stored as strings.

src/components/GradeLevelFilter.jsx
```diff
@@ -63,20 +56,23 @@ export default function GradeLevelFilter() {
...
-                    checked={selectedGrades.has(grade)}
-                    onChange={() => toggleSelection(grade)}
-                    name="grade"
-                  />
-                }
-                label={grade}
-              />
-            ))}
+            {Object.entries(GradeLevel.values()).map(([val, stringRep]) => {
+              val = parseInt(val);
```
edwingl10, line 60: need to convert to integer since the params are converted to integers as well.

src/components/GradeLevelFilter.jsx
```diff
@@ -63,20 +56,23 @@ export default function GradeLevelFilter() {
...
+              return (
+                <FormControlLabel
+                  key={val}
+                  control={
+                    <Checkbox
+                      color="primary"
+                      checked={grades.indexOf(val) > -1}
```
josueetcom, line 67: I suggest leaving `grades` as a Set and setting its PropType to be a set. You'll have to convert an array to a set in the caller.

src/components/FilterBar.jsx
```diff
@@ -29,15 +29,20 @@ const majors = {
...
   return (
     <Grid container spacing={2} className={classes.root}>
       <Grid item className={classes.alignText}>
         {experiments.expShowMajorFilter && (
           <FilterDropDown label="Major" items={majors} />
         )}
-        {experiments.expShowGradeFilter && <GradeLevelFilter />}
+        {experiments.expShowGradeFilter && (
+          <GradeLevelFilter
+            grades={grades ?? []}
```
josueetcom, line 42:
```suggestion
            grades={new Set(grades)}
```

src/pages/ListScholarships.jsx
```diff
@@ -58,6 +79,23 @@ function ListScholarships() {
...
+      GradeLevel.keys().includes(g)
+    );
+    if (prunedInvalid.length !== grades.length) {
+      replaceQueryParam(qParams.GRADES, prunedInvalid);
+    }
+  }
+  if (
+    (grades !== undefined && !Array.isArray(grades)) ||
+    JSON.stringify(grades) === '[]'
+  )
```
josueetcom, line 96: Surround both `if` statements with blocks with the `undefined` and `isArray` checks so that these other conditions are simplified. E.g.

```js
if (grades !== undefined) {
  if (!Array.isArray(grades)) {
    // prune
  }
  // validate/prune array values
  // param altogether if it's empty.
}
```

josueetcom: Use `grade.length` to check for emptiness

josueetcom (changes requested):

src/pages/ListScholarships.jsx
```diff
@@ -58,6 +79,21 @@ function ListScholarships() {
...
+    if (Array.isArray(grades) && grades.length > 0) {
+      const prunedInvalid = [...new Set(grades)].filter((g) =>
+        GradeLevel.keys().includes(g)
+      );
+
+      if (prunedInvalid.length !== grades.length) {
+        replaceQueryParam(qParams.GRADES, prunedInvalid);
+      }
+    } else pruneQueryParam(qParams.GRADES);
+  }
```
edwingl10, line 95: tested against the following scenarios:
`grades`
`grades[]`
`grades[]=`
`grades[]=1,2,8`

src/components/GradeLevelFilter.jsx
```diff
@@ -63,23 +55,34 @@ export default function GradeLevelFilter() {
...
-                    color="primary"
-                    checked={selectedGrades.has(grade)}
-                    onChange={() => toggleSelection(grade)}
-                    name="grade"
-                  />
-                }
-                label={grade}
-              />
-            ))}
+            {Object.entries(GradeLevel.values()).map(([val, stringRep]) => {
```
josueetcom, line 58: How about:
```suggestion
            {Object.entries(GradeLevel.values()).map(([grade, stringRep]) => {
```

src/components/GradeLevelFilter.jsx
```diff
@@ -63,23 +55,34 @@ export default function GradeLevelFilter() {
...
     </>
   );
 }
+
+GradeLevelFilter.propTypes = {
+  grades: PropTypes.instanceOf(Set),
+  changeFn: PropTypes.func.isRequired,
+};
+GradeLevelFilter.defaultProps = {
+  grades: undefined,
```
josueetcom, line 87:
```suggestion
  grades: new Set(),
```

src/pages/ListScholarships.jsx
josueetcom: I'm guessing this won't prune though when you do `grades[]=1,2,3` when all the values are invalid.

josueetcom (approved):

src/pages/ListScholarships.jsx
edwingl10: it does, first it replaces it with `[]` and then that gets pruned 😃

PR #727, opened 2021-11-14
Update README.md

Jobhdez:
I bought a new computer and after following the instructions I realized that `yarn start`, at least, needs JAVA or OpenJDK. It needs it. new users need to be aware that they also need Java installed to contribute.


Jobhdez:
java is not needed. my bad

josueetcom:
So I actually ran into this issue on my new Macbook too. Are you sure this isn't needed?

README.md
```diff
@@ -10,6 +10,7 @@ The following are required:
 
 - [Git](https://git-scm.com/)
 - [Node.js](https://nodejs.org/)
+- [Java](https://docs.oracle.com/en/java/javase/11/install/overview-jdk-installation.html#GUID-8677A77F-231A-40F7-98B9-1FD0B48C346A)
```
josueetcom, line 13: How about this link instead?
https://www.oracle.com/java/technologies/downloads/

Jobhdez: okay

Jobhdez:
> So I actually ran into this issue on my new Macbook too. Are you sure this isn't needed

> So I actually ran into this issue on my new Macbook too. Are you sure this isn't needed?

i ran into the same problem with the macbook i bought. but i returned and using my ubuntu computer. on my ubuntu computer i uninstalled java and still worked; it might be a mac dependency

josueetcom:
> i ran into the same problem with the macbook i bought. but i returned and using my ubuntu computer. on my ubuntu computer i uninstalled java and still worked; it might be a mac dependency

Maybe Ubuntu already has Java installed? I believe the emulator depends on the Java runtime, which may already be bundled into Ubuntu.

josueetcom (approved):

Issue #636, opened 2021-09-28
Upgrade Material UI to 5


josueetcom:
This is the latest stable version per https://mui.com/getting-started/usage/

Per [migration guide](https://mui.com/guides/migration-v4), we need to:

- [x] Upgrade React to v17
- [x] Upgrade (if needed) react-scripts, @types/react, @types/react-dom
- [x] Setup ThemeProvider properly
- [x] Update MUI version
- [x] Run codemods
- [x] Handle individual breaking changes
  - [x] Theme structure
  - [x] [Pickers](https://mui.com/guides/pickers-migration/)
- [x] #847

PR #736, opened 2021-11-19
Migrate some JSS to SX props

josueetcom:
## Motivation and Context

Helps with #636. Basically, we should use `sx` property wherever styling applies to only one element. But if we want a styling to apply to multiple elements then we should used the [`styled`](https://mui.com/system/styled/#main-content) API.

Additional context: https://mui.com/guides/migration-v4/#migrate-from-jss

## How Has This Been Tested?

Existing tests should provide decent coverage. 🤷 

## Types of changes


josueetcom:
I recommend hiding whitespace when reviewing this! There's only about 84 additions but like 200+ deletions

edwingl10 (approved):

PR #742, opened 2021-11-23
grade filter functionality

### LINKED ISSUE #274: Grade level filter: functionality

edwingl10:
## Motivation and Context

Fixes #274

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-742.onrender.com/scholarships.
2. choose some grade filters

## Checklist:


src/types/GradeLevel.ts
```diff
@@ -46,6 +46,14 @@ namespace GradeLevel {
   export function toString(level: GradeLevel): string {
     return toStringMappings[level];
   }
+
+  export function includesGrade(
+    grades?: number[],
+    paramGrades?: number[]
+  ): boolean {
+    if (!paramGrades || !grades) return true;
+    return grades.some((g) => paramGrades.includes(g));
```
josueetcom, line 55: This can be written as one line:
```suggestion
    return !paramGrades || !grades || grades.some((g) => paramGrades.includes(g));
```

Actually, other filters will probably be similar (e.g. majors, states, etc.) so I suggest either inlining this or refactoring this as a generic function in Scholarships.ts

src/models/Scholarships.test.ts
```diff
@@ -19,6 +19,7 @@ function create(data: {
   amount?: ScholarshipAmount | undefined;
   name?: string;
   deadline?: Date;
+  grades?: number[];
```
josueetcom, line 22:
```suggestion
  grades?: GradeLevel[];
```

You're going to need to use the `GradeLevel` types elsewhere in tests (you can unwrap them to avoid the `GradeLevel.` prefix if you'd like).

src/types/GradeLevel.test.ts
```diff
@@ -0,0 +1,63 @@
+import GradeLevel from './GradeLevel';
+
+test('check Grade key values', () => {
+  const keys = [];
+  for (let k = 8; k <= 22; k++) keys.push(k);
+
+  expect(GradeLevel.keys()).toEqual(keys);
+});
+
+test('toString() - middle school', () => {
```
josueetcom, line 10: These `toString()` tests aren't particularly interesting because they all test the same thing: that a value from a mapping gets used. Also after high school the grade level values become a bit hard to decipher. You should use `GradeLevel` enums themselves.

More interesting test cases would be if there were some sort of `rangeToString()` function that can simplify the string represenation of multiple values. E.g. `High School Freshman, Sophomore, Junior` if those three are selected or `All High School students` if all four High School options are selected.

src/models/Scholarships.test.ts
edwingl10: what do you mean by unwrapping them?

src/types/GradeLevel.test.ts
edwingl10: interesting, I can come back to this in the near future.

src/types/GradeLevel.ts
edwingl10: I refactored this as a generic function

src/models/Scholarships.test.ts
josueetcom: I guess in JS they call it [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment). I.e. if you can't do
```js
import { MIDDLE_SCHOOL, ... } from '../types/GradeLevel.ts'
```
You could do this at the top of the tests:
```js
const { MIDDLE_SCHOOL, ... } = GradeLevel;
```
So that you don't need to write the `GradeLevel.` prefix everywhere

src/models/Scholarships.ts
```diff
@@ -155,6 +159,16 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
...
+
+  /**
+   *
+   * @param reqs is the scholarship requirements e.g. (grades, states, etc.)
+   * @param paramFilters
+   * @returns whether a scholarship's requirements are included in the given filters
+   */
+  includesFilter(reqs?: any[], paramFilters?: any[]): boolean {
+    return !paramFilters || !reqs || reqs.some((r) => paramFilters.includes(r));
+  }
```
josueetcom, line 171: I suggest defining this outside of the class (probably before the class).

src/models/Scholarships.ts
```diff
@@ -155,6 +159,16 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
...
   }
+
+  /**
+   *
+   * @param reqs is the scholarship requirements e.g. (grades, states, etc.)
+   * @param paramFilters
+   * @returns whether a scholarship's requirements are included in the given filters
+   */
+  includesFilter(reqs?: any[], paramFilters?: any[]): boolean {
+    return !paramFilters || !reqs || reqs.some((r) => paramFilters.includes(r));
```
josueetcom, line 170: You should handle the cases where either array is empty:
```suggestion
    return !paramFilters?.length || !reqs?.length || reqs.some((r) => paramFilters.includes(r));
```

Probably need some test cases for them.

src/models/Scholarships.ts
```diff
@@ -155,6 +159,16 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
...
     return new FirestoreModel<ScholarshipData>(this.collection.doc(), data);
   }
+
+  /**
+   *
+   * @param reqs is the scholarship requirements e.g. (grades, states, etc.)
+   * @param paramFilters
+   * @returns whether a scholarship's requirements are included in the given filters
+   */
+  includesFilter(reqs?: any[], paramFilters?: any[]): boolean {
```
josueetcom, line 169: Hmm, I'm realizing this is a hard method to name because we want it to return true when:
- there are no requirements OR
- there are no filters for this requirement OR
- there is an intersection between the requirements and the filters

Ideally this could be named something like `containsAny` or `intersects` and its semantics are well understood but that would normally imply the first two cases are false rather than true.

josueetcom: How about `requirementMatchesFilter`? You might want to give a meaningful JSDoc summary. E.g. 

```
Returns whether the given requirement list matches the given filter list. True is returned when any of the following conditions are met:
- There are no requirements.
- There are no filters.
- The requirement list intersects with the filter list.
```

src/models/Scholarships.test.ts
```diff
@@ -276,6 +279,38 @@ test('scholarships.list - filters by maxAmount', async () => {
...
+  expect(got.results.map(extractName).sort()).toEqual(
+    want.map(extractName).sort()
+  );
+});
+
+test('scholarships.list - filters by grades (all grades)', async () => {
+  await Promise.all(gradeScholarships.map((s) => s.save()));
+
+  const got = await Scholarships.list({
+    grades: [8, 11, 17, 20],
```
josueetcom, line 305: Again I suggest using the enum values themselves here and elsewhere.

src/models/Scholarships.ts
edwingl10: initially I didn't think it was necessary since we prune empty arrays, but I think it's still a good check to have.

src/models/Scholarships.test.ts
```diff
@@ -369,23 +404,43 @@ test('scholarships.new - default values', async () => {
 });
 
 test('includesFilter() - no requirements', () => {
```
josueetcom, line 406: Gonna want to update these names. Also you can use simple array values (instead of `GradeLevel` ones) for these since it's a generic function comparing two arrays.

josueetcom (approved):

src/models/Scholarships.test.ts
```diff
@@ -403,44 +403,24 @@ test('scholarships.new - default values', async () => {
...
-test('includesFilter() - requirement not in param filters', () => {
-  const grades = [7, GradeLevel.CollegeFreshman, GradeLevel.CollegeSophomore];
-  const paramGrades = [
-    GradeLevel.MiddleSchool,
-    GradeLevel.HsSophomore,
-    GradeLevel.CollegeJunior,
-  ];
+test('requirementMatchesFilter() - requirement not in param filters', () => {
+  const grades = [7, 13, 14];
+  const paramGrades = [8, 10, 15];
```
josueetcom, line 424: nit: name these something more generic as it's applicable beyond grades.

josueetcom (dismissed):

josueetcom (approved):

PR #743, opened 2021-11-23
added tests for the filter bar component.

Jobhdez:
## Motivation and Context
added some tests for the filter bar component. tests are important because it will allow you to refactor down the line and it will prevent you from making changes the break the web system.

## How Has This Been Tested?
This has been tested by running `yarn test.` all the tests I made passed.

## Types of changes

- [x] Added tests

## Checklist:

- [ ] Maybe add more tests

### DIFF
--- a/src/components/FilterBar.test.jsx
+++ b/src/components/FilterBar.test.jsx
@@ -1,0 +1,37 @@
+import { render, screen } from '@testing-library/react';
+import queryString from 'query-string';
+import { createTheme, ThemeProvider } from '@mui/material/styles';
+import FilterBar from './FilterBar';
+
+const renderWithTheme = (ui) =>
+  render(<ThemeProvider theme={createTheme()}>{ui}</ThemeProvider>);
+
+const queryOptions = {
+  arrayFormat: 'bracket-separator',
+  arrayFormatSeparator: ',',
+};
+
+const params = queryString.parse(
+  { sortBy: 'amount.asc' },
+  {
+    parseNumbers: true,
+    ...queryOptions,
+  }
+);
+
+const setQueryParam = () => {};
+
+describe('HomeSection', () => {
+  test('renders the component', () => {
+    renderWithTheme(<FilterBar queryParams={params} {...{ setQueryParam }} />);
+  });
+  test('checks the number of buttons', () => {
+    renderWithTheme(<FilterBar queryParams={params} {...{ setQueryParam }} />);
+    const buttons = screen.queryAllByRole('button');
+    expect(buttons).toHaveLength(2);
+  });
+  test('check the name of the button', () => {
+    renderWithTheme(<FilterBar queryParams={params} {...{ setQueryParam }} />);
+    expect(screen.getByText('Amount')).toBeInTheDocument();
+  });
+});


src/components/FilterBar.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+  arrayFormatSeparator: ',',
+};
+
+const params = queryString.parse(
+  { sortBy: 'amount.asc' },
+  {
+    parseNumbers: true,
+    ...queryOptions,
+  }
+);
```
edwingl10, line 20: you can just set `params` to an object that contains `{sortBy: 'amount.asc'}`
queryString is probably not needed.

src/components/FilterBar.test.jsx
```diff
@@ -0,0 +1,37 @@
+import { render, screen } from '@testing-library/react';
+import queryString from 'query-string';
+import { createTheme, ThemeProvider } from '@mui/material/styles';
+import FilterBar from './FilterBar';
+
+const renderWithTheme = (ui) =>
+  render(<ThemeProvider theme={createTheme()}>{ui}</ThemeProvider>);
+
+const queryOptions = {
```
edwingl10, line 9: not needed

src/components/FilterBar.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+    parseNumbers: true,
+    ...queryOptions,
+  }
+);
+
+const setQueryParam = () => {};
+
+describe('HomeSection', () => {
+  test('renders the component', () => {
+    renderWithTheme(<FilterBar queryParams={params} {...{ setQueryParam }} />);
```
edwingl10, line 26: not testing anything here, but if you write tests checking for `sort` and `amount filter` buttons, then that would imply that `FilterBar` is rendering

src/components/FilterBar.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+  });
+  test('checks the number of buttons', () => {
```
edwingl10, line 28:
```suggestion
test('Amount filter rendered', () => {
```

src/components/FilterBar.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+    renderWithTheme(<FilterBar queryParams={params} {...{ setQueryParam }} />);
+    const buttons = screen.queryAllByRole('button');
+    expect(buttons).toHaveLength(2);
+  });
+  test('check the name of the button', () => {
+    renderWithTheme(<FilterBar queryParams={params} {...{ setQueryParam }} />);
+    expect(screen.getByText('Amount')).toBeInTheDocument();
```
edwingl10, line 35: these test can be grouped together.

Also you should probably write tests that checks if the sort button is rendered and that it renders the correct text based on the `params`

edwingl10 (changes requested):

src/components/FilterBar.test.jsx
Jobhdez: `renderWithTheme` seems to be needed. I ran the code without it and it didnt work. the error that I got mentioned `ThemeProvider`

edwingl10: I was refering to the `queryOptions` object (:

src/components/FilterBar.test.jsx
Jobhdez: can I just leave them the way they are? each behavior has its own test. isnt this more modular?

src/components/FilterBar.test.jsx
josueetcom: ^ I think that's supposed to be the name of the test below.

This could be a `"renders amount filter and sorting dropdown by default"` and you could add `expect` calls that additionally make sure each are rendered and the default sorting value is set.

src/components/FilterBar.test.jsx
```diff
@@ -0,0 +1,47 @@
...
+
+describe('FilterBar', () => {
+  test('SortBy filter (deadline.asc) rendered', () => {
+    const params = { sortBy: 'deadline.asc' };
+    renderWithTheme(<FilterBar queryParams={params} {...{ setQueryParam }} />);
+    expect(
+      screen.getByText('Deadline (Earliest to Latest)')
+    ).toBeInTheDocument();
+  });
+  test('SortBy filter (deadline.desc) rendered', () => {
```
josueetcom, line 18: nit: prefer spaces between tests.

here and elsewhere

src/components/FilterBar.test.jsx
```diff
@@ -0,0 +1,47 @@
...
+import { createTheme, ThemeProvider } from '@mui/material/styles';
+import FilterBar from './FilterBar';
+
+const renderWithTheme = (ui) =>
+  render(<ThemeProvider theme={createTheme()}>{ui}</ThemeProvider>);
+
+const setQueryParam = () => {};
+
+describe('FilterBar', () => {
+  test('SortBy filter (deadline.asc) rendered', () => {
```
josueetcom, line 11: nit: prefer to use the active tense in test names. E.g. `renders SortBy filter (deadline.asc)`

src/components/FilterBar.test.jsx
```diff
@@ -0,0 +1,47 @@
...
+    expect(
+      screen.getByText('Deadline (Latest to Earliest)')
+    ).toBeInTheDocument();
+  });
+  test('SortBy filter (amount.desc) rendered', () => {
+    const params = { sortBy: 'amount.desc' };
+    renderWithTheme(<FilterBar queryParams={params} {...{ setQueryParam }} />);
+    expect(screen.getByText('Amount (High to Low)')).toBeInTheDocument();
+  });
+  test('SortBy filter (amount.asc) rendered', () => {
```
josueetcom, line 30: These tests seem to be testing `FilterDropdown` more than `FilterBar`. I think what you want to do is have tests that check:
- How the component renders by default. E.g. `"renders amount filter and sorting dropdown by default"`
- How the component renders when params are given (you could do this as one test or as a single test per param). E.g. `"renders passed in sorting option"`.
- How the component renders when experiments are enabled (again, either as one test or as a single test per experiment). E.g. `"renders majors filter when experiment enabled"`.
- How the component interacts with `setQueryParam`, maybe once per filter value. This involves interacting with the UI element itself using `get*By*` and setting some property. See [this guide](https://reactjs.org/docs/test-utils.html#act) for an example. E.g. `"calls setQueryParam when sorting option changed"`

That being said, feel free to merge these changes in separate stages/PRs. Some of our test files are actually missing some of these pieces too.

josueetcom (changes requested):

src/components/FilterBar.test.jsx
josueetcom: You can even get some of the bullets done above and then file a separate issue to add the remaining tests.

PR #745, opened 2021-11-29
added tests for the aboutcard component.

Jobhdez:
## Motivation and Context
Added tests for the aboutCard component

Fixes #
adds tests
## How Has This Been Tested?
when I `yarn test` the tests I wrote pass

## Types of changes
tests

- [x] component tests
## Previewing Changes

2.

## Checklist:

### DIFF
--- a/src/components/AboutCard.test.jsx
+++ b/src/components/AboutCard.test.jsx
@@ -6,17 +6,9 @@ import testPic from '../logo.svg';
 const renderWithTheme = (ui) =>
   render(<ThemeProvider theme={createTheme()}>{ui}</ThemeProvider>);
 
-const team = [
-  {
-    name: 'Job Hernandez',
-    img: testPic,
-    description: 'I love math and science and technology.',
-  },
-];
-
-const name = team[0].name;
-const img = team[0].img;
-const description = team[0].description;
+const name = 'Job Hernandez';
+const img = testPic;
+const description = 'I love math and science and technology.';
 
 describe('AboutCard', () => {
   test('renders props', () => {
--- a/src/components/AboutCard.test.jsx
+++ b/src/components/AboutCard.test.jsx
@@ -10,7 +10,7 @@ const team = [
   {
     name: 'Job Hernandez',
     img: testPic,
-    description: <>I love math and science and technology.</>,
+    description: 'I love math and science and technology.',
   },
 ];
 
@@ -19,26 +19,16 @@ const img = team[0].img;
 const description = team[0].description;
 
 describe('AboutCard', () => {
-  test('renders the given name', () => {
+  test('renders props', () => {
     renderWithTheme(<AboutCard {...{ name, img, description }} />);
-    screen.debug();
-    expect(screen.getByText('Job Hernandez')).toBeInTheDocument();
-  });
 
-  test('renders the description', () => {
-    const { container } = renderWithTheme(
-      <AboutCard {...{ name, img, description }} />
-    );
-    container.querySelector(
-      'MuiTypography-root MuiTypography-body2 css-e784if-MuiTypography-root'
-    );
-    expect(container.firstChild).toBeInTheDocument();
-    screen.debug();
-  });
+    expect(screen.getByText(name)).toBeInTheDocument();
 
-  test('renders the image', () => {
-    renderWithTheme(<AboutCard {...{ name, img, description }} />);
-    const imageTitle = screen.getByTitle('picture of Job Hernandez');
-    expect(imageTitle).toBeInTheDocument();
+    const image = screen.getByRole('image', {
+      name: 'picture of Job Hernandez',
+    });
+    expect(image).toBeInTheDocument();
+
+    expect(screen.getByText(description)).toBeInTheDocument();
   });
 });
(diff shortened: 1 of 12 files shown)


src/components/AboutCard.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+];
+
+const name = team[0].name;
+const img = team[0].img;
+const description = team[0].description;
+
+describe('AboutCard', () => {
+  test('renders the given name', () => {
+    renderWithTheme(<AboutCard {...{ name, img, description }} />);
+    screen.debug();
```
josueetcom, line 24: Remove? This looks like it was meant to help with debugging.

Here and elsewhere.

src/components/AboutCard.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  test('renders the description', () => {
+    const { container } = renderWithTheme(
+      <AboutCard {...{ name, img, description }} />
+    );
+    container.querySelector(
+      'MuiTypography-root MuiTypography-body2 css-e784if-MuiTypography-root'
+    );
+    expect(container.firstChild).toBeInTheDocument();
```
josueetcom, line 35: why not use `getByText` like you did in the other test?

src/components/AboutCard.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+    screen.debug();
+  });
+
+  test('renders the image', () => {
+    renderWithTheme(<AboutCard {...{ name, img, description }} />);
+    const imageTitle = screen.getByTitle('picture of Job Hernandez');
```
josueetcom, line 41:
```suggestion
    const image = screen.getByRole('image', { name: 'picture of Job Hernandez' })；
```

josueetcom (changes requested):

src/components/AboutCard.test.jsx
Jobhdez: `getByText` doesn’t work. I tried it already and it can’t find the description

src/components/AboutCard.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+    expect(imageTitle).toBeInTheDocument();
+  });
```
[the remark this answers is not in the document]
Jobhdez: can i please leave like that. i think different tests might be better. have things a little bit more modular.

src/components/AboutCard.test.jsx
Jobhdez: I will remove it

src/components/AboutCard.test.jsx
Jobhdez: done

src/components/AboutCard.test.jsx
josueetcom: FYI, no commits were pushed to this branch.

src/components/AboutCard.test.jsx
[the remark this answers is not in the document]
josueetcom: Generally speaking different tests should test different cases/scenarios. In each of these tests they're testing the same scenario, specifically this one:
```jsx
renderWithTheme(<AboutCard {...{ name, img, description }} />);
```

There's no need to write three tests for the same scenario when you can write one.

Now, where would separate tests for each property like these make sense? If each one of them were optional. Because then you'd test these cases:
```jsx
renderWithTheme(<AboutCard />);                            // no props set
renderWithTheme(<AboutCard  name={name} />);               // name set
renderWithTheme(<AboutCard  img={img} />);                 // image set
renderWithTheme(<AboutCard  description={description} />); // description set
```

Note that this doesn't even cover 4 other theoretical scenarios (3 where two props are set + 1 where all 3 props are set). But they cover all the conditional logic one would expect as a user. As a user my assumption is whether or not I set one property is independent from whether or not I set another.

Now imagine having separate tests for _each_ property (like you have in your code) for _each_ scenario. That's anywhere between 4-8 scenarios * 3 prop tests/scenario = 12-24 tests. That's not too bad, but it's much more than just writing one test per useful scenario.

Now imagine having _n_ optional properties. that's _2^n_ possible combinations/scenarios (any prop could be set or unset) and _n*2^n_ tests if you want to do one test per property per scenario. So for 5 properties, 32 scenarios and 160 tests. For 10 properties, 1024 scenarios and 10k tests. It's not practical nor useful to exhaustively test all scenarios and write several tests for the same scenario when one suffices. What's most practical is to test the default case, one case per property, and have those tests validate everything that's relevant about them. If some properties interact with other ones then there'd be separate cases for those scenarios that aren't already covered in other tests.

src/components/AboutCard.test.jsx
josueetcom: That might be because `description` is a node element here (it has `<></>`). If you make it a string or do `getByText('some substring of the description')` or `getByText(description.innerText)` it should work.

src/components/AboutCard.test.jsx
josueetcom: I don't see any new commits

src/components/AboutCard.test.jsx
```diff
@@ -0,0 +1,34 @@
...
+  {
+    name: 'Job Hernandez',
+    img: testPic,
+    description: 'I love math and science and technology.',
+  },
+];
+
+const name = team[0].name;
+const img = team[0].img;
+const description = team[0].description;
```
josueetcom, line 19: If you're going to just unwrap the only element of `team` then you may as well just define the variables you need.
```suggestion
const name = 'Job Hernandez';
const img = testPic;
const description = 'I love math and science and technology.';
```

josueetcom (approved):
Just one minor change. You should be able to push it and then merge.

PR #760, opened 2021-12-02
public home page styling

### LINKED ISSUE #43: Public home page: styling
Let's make it look more like the mocks

edwingl10:
## Motivation and Context

Fixes #43

## How Has This Been Tested?
locally

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-760.onrender.com.
2. make sure you are logged out

## Checklist:


src/components/HomeSection.jsx
```diff
@@ -5,61 +5,62 @@ import { Container, Grid, Typography } from '@mui/material';
 const HomeSection = ({
```
edwingl10, line 5: was thinking of renaming this component to something more generic since we can use this in other pages. 
something like `contentGridSection` or `gridSection`

src/components/HomeSection.jsx
```diff
@@ -5,61 +5,62 @@ import { Container, Grid, Typography } from '@mui/material';
...
   alignItems,
   direction,
-  tab,
   title,
+  main,
```
edwingl10, line 9: used to change typography and image size. I noticed on the figma design that the top section had bigger text and a bigger image.

josueetcom: What if we just pass in the whole `img` node instead?

src/components/HomeSection.jsx
josueetcom: What other pages were you thinking? AFAIK we only use this on the home page and we may as well make this an unexported component in `PublicHome` 😛

src/pages/PublicHome.jsx
```diff
@@ -1,33 +1,55 @@
...
-      <ScholarshipsMadeSimple />
       <HomeSection
-        alignItems="center"
-        direction="row-reverse"
-        title="For and by the Community"
-        description="Interested in helping students find scholarships? Join the family of community contributors, help find scholarships that are open to anyone regardless of citizen status."
+        direction="row"
+        title="Scholarships for everyone."
+        main
+        description="Find your dream scholarship together."
```
josueetcom, line 16: FYI this wording was changed recently. I'm actually going to leave some comments on the Figma design for what our wording for a bunch of things could be.

src/components/HomeSection.jsx
josueetcom: Oh nvm, you also need to make the typography a little bigger too.

src/components/HomeSection.jsx
```diff
@@ -5,61 +5,62 @@ import { Container, Grid, Typography } from '@mui/material';
...
   buttons: [],
+  description: undefined,
 };
 
 HomeSection.propTypes = {
   alignItems: PropTypes.string,
   direction: PropTypes.string,
-  tab: PropTypes.string,
   title: PropTypes.string.isRequired,
-  description: PropTypes.string.isRequired,
```
josueetcom, line 62: Shouldn't this be required instead of undefined?

src/components/HomeSection.jsx
edwingl10: We could use it for the top section in the `About page` but that piece of code is not that complex, so it might not be necessary.

src/components/HomeSection.jsx
edwingl10: The last part of the page also uses this component and doesn't have any description text.

src/pages/PublicHome.jsx
edwingl10: nice, I'll look out for those.

josueetcom (approved):

src/components/HomeSection.jsx
```diff
@@ -5,61 +5,62 @@ import { Container, Grid, Typography } from '@mui/material';
...
+            ))}
+          </Grid>
+        </Grid>
+        <Grid item xs={12} sm={6}>
+          <img
+            src={pic}
+            alt=""
+            style={{
+              width: picSize,
+              height: picSize,
```
josueetcom, line 40: One thing I'm noticing is that the height of each tab is slightly different

josueetcom: It's actually collapsed when a picture is loading.

edwingl10: fixed by setting size of grid item

josueetcom (approved):

PR #789, opened 2021-12-14
i18n setup & Translations

### LINKED ISSUE #675: Support Spanish in the app
Investigate how we can support Spanish in the app. There's a few ways we can go about this:
- Natively support Spanish
- Use something like Google Translate to translate the site

Personally I prefer the former. But then there are a few ways to support Spanish natively as well:
- Hardcoded strings for each language stored in the code
- Hardcoded strings loaded from a config, either locally or remotely using something like [Firebase Remote Config](https://firebase.google.com/docs/remote-config)

There are trade-offs doing things one way or another so I'd like to see a doc exploring the various options and explaining why we should adopt a particular design vs. others.

Additional resources for MUI:https://mui.com/guides/localization/

edwingl10:
## Motivation and Context

Fixes #675. Uses [react-i18next](https://react.i18next.com/) library for translation capability.

## How Has This Been Tested?
locally.
planning to add translation tests in another PR.

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-789.onrender.com.
2. to change language, you will have to change language in index.html file. 

## Checklist:


src/components/ScholarshipList.jsx
```diff
@@ -84,11 +86,10 @@ function ScholarshipList({ noResultsNode, listFn }) {
 
 ScholarshipList.propTypes = {
   listFn: PropTypes.func,
-  noResultsNode: PropTypes.node,
+  noResultsNode: PropTypes.node.isRequired,
```
edwingl10, line 89: made this required so it can be translated.

src/i18n/setup.js
```diff
@@ -0,0 +1,24 @@
...
+
+i18n
+  .use(initReactI18next)
+  .use(LanguageDetector)
+  .init({
+    resources,
+    supportedLngs: ['en', 'es'],
+    fallbackLng: 'en',
+    detection: {
+      order: ['htmlTag', 'cookie', 'localStorage', 'path', 'subdomain'],
```
edwingl10, line 20: order will change once we allow language to be changed.
I was thinking `['cookie', 'localStorage', 'path', 'subdomain', 'htmTag']`

src/i18n/translations/english.json
```diff
@@ -0,0 +1,88 @@
...
+      "gonzalo": "My passion for computers has lead me to pursue a career in Software Engineering. I am currently a UWB CSS student completing my final year in the Computer Science and Software Engineering program. I am 24 years old, bilingual and currently a part of the DACA program. I intend to contribute as much as possible to building a better tomorrow.",
+      "sergio": "I’m a UX Designer that is interested in using technology as a means to solve problems. My personal experiences have inspired my focus on empathy and universal design. I find joy in the ability to explore ideas with a central aim to create solutions that will help people solve meaningful problems.",
+      "job": "I am driven to solve problems with software; software has tremendous power to advance society. So, I would like to contribute and have an impact."
+    }
+  },
+
+  "contact": {
+    "bugTitle": "Found a bug?",
+    "bugDesc": "You can file it <1>here</1>.",
```
edwingl10, line 44: `<1></1>` might be confusing but they are needed to allow `React` components

src/pages/About.jsx
```diff
@@ -9,76 +9,46 @@ import {
...
+
+  const team = [
+    {
+      name: 'Josue Rios',
+      img: 'https://avatars.githubusercontent.com/u/8023233?v=4',
+      description: (
+        <Trans i18nKey="about.team.josue">
+          start
+          <MuiLink href="http://beyondhb1079.org/">Beyond HB 1079</MuiLink> end
+        </Trans>
```
edwingl10, line 27: to learn more about the [Trans](https://react.i18next.com/latest/trans-component) component.

Basically it breaks up the text into the following parts
1 -> start
2 -> <MuiLink> </MuiLink>
3 -> end

I could have left the oiginal text instead of `start/end` but I thought those saved some space.

src/pages/AddScholarship.jsx
```diff
@@ -18,15 +21,11 @@ function AddScholarship() {
...
           <Typography variant="h4" gutterBottom>
-            Additional information
-          </Typography>
-          <Typography>
-            Help our team provide the best search results for students and
-            community members looking for scholarships by contributing as much
-            detailed information possible.
+            {t('addScholarship.title')}
           </Typography>
+          <Typography>{t('addScholarship.desc')}</Typography>
```
edwingl10, line 28: was also thinking of translating the instructions, but I'm not sure if it's necessary since the actual scholarship form is not translated.

PR #806, opened 2021-12-28
i18next setup & header translations

### LINKED ISSUE #675: Support Spanish in the app

edwingl10:
## Motivation and Context

Fixes #675 
Setup files and Header translations for proof of concept.

## How Has This Been Tested?
locally

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-806.onrender.com.
2. click on icon to change language

## Checklist:


src/i18n/translations/english.json
```diff
@@ -0,0 +1,20 @@
...
+  "navbar": {
+    "scholarships": "Scholarships",
+    "add": "Add",
+    "about": "About",
+    "contact": "Contact"
+  },
+
+  "btn": {
+    "browse": "Browse Scholarships",
+    "add": "Add a Scholarship",
```
edwingl10, line 11: used in the other stacked PRs.
Included it here so I wouldn't have to add them for every PR.

src/components/Header.jsx
```diff
@@ -130,10 +142,36 @@ function Header() {
...
+      <Menu
+        open={Boolean(anchorEl)}
+        anchorEl={anchorEl}
+        onClose={() => setAnchorEl(null)}
+        anchorOrigin={{ horizontal: 'center', vertical: 'bottom' }}
+        transformOrigin={{ horizontal: 'center', vertical: 'top' }}>
+        {Object.entries(languages).map(([k, v], index) => (
+          <MenuItem
+            key={k}
+            selected={index === selectedIndex}
```
josueetcom, line 165: instead of using `index` and `selectedIndex` can you compare `v` with the current language?

edwingl10: We could do this in a future PR where we support the language via params since I don't think there is a method that outputs what language is being used.

edwingl10: Nice

src/components/Header.jsx
```diff
@@ -108,17 +111,25 @@ const useStyles = makeStyles((theme) => ({
...
-  Contact: '/contact',
+  'navbar.scholarships': '/scholarships',
+  'navbar.add': '/scholarships/new',
+  'navbar.about': '/about',
+  'navbar.contact': '/contact',
+};
+
+const languages = {
+  English: 'en',
+  Español: 'es',
```
josueetcom, line 122: Nit: Not sure how well special characters like ñ are supported as keys (seems to work here) but perhaps we can flip this so that it maps from language code to the name of the language? If we ever decide to support other languages with more complex characters like 한국어 or 中文 I'm not sure how well they'll work as keys (maybe they can as strings).

src/components/Header.jsx
```diff
@@ -130,10 +141,35 @@ function Header() {
...
         </Grid>
       </Grid>
+
+      <Menu
+        open={Boolean(anchorEl)}
+        anchorEl={anchorEl}
+        onClose={() => setAnchorEl(null)}
+        anchorOrigin={{ horizontal: 'center', vertical: 'bottom' }}
+        transformOrigin={{ horizontal: 'center', vertical: 'top' }}>
+        {Object.entries(languages).map(([k, v]) => (
```
josueetcom, line 161: Nit: I suggest using slightly better names than `k` and `v` as it's not entirely clear what maps to what here. Maybe `lang` and `code`/`abbr`?

josueetcom (approved):

Issue #538, opened 2021-07-02
Major filter: basic component


josueetcom:
Basic component that supports selecting multiple majors and unblocks other related issues.

PR #849, opened 2022-01-04
Majors filter styling

### LINKED ISSUE #308: Major filter: styling
Conform to the design in #310

edwingl10:
## Motivation and Context

Fixes #308 #538
Chosen options are rendered as chips on top of search area.

## How Has This Been Tested?
local

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-849.onrender.com/scholarships?expShowMajorFilter=1.
2. Search and choose majors

## Checklist:


src/components/MajorFilter.jsx
```diff
@@ -0,0 +1,63 @@
...
+          />
+        ))}
+        <Autocomplete
+          multiple
+          open={isOpen}
+          value={majors}
+          onChange={(e, val) => setMajors(val)}
+          onInputChange={(e, val) => {
+            if (val.length > 0) setIsOpen(true);
+            else setIsOpen(false);
```
edwingl10, line 43: prevents listbox from showing when there is nothing in the input field like below,

Screen Shot 2022-01-04 at 2 44 29 PM

src/components/MajorFilter.jsx
```diff
@@ -0,0 +1,63 @@
...
+      </Button>
+      <Popover
+        open={Boolean(anchorEl)}
+        anchorEl={anchorEl}
+        onClose={() => setAnchorEl(null)}
+        anchorOrigin={{ vertical: 'bottom', horizontal: 'left' }}
+        transformOrigin={{ vertical: 'top', horizontal: 'left' }}
+        sx={{ maxWidth: 400 }}>
+        {majors.map((major) => (
+          <Chip
```
edwingl10, line 28: This might not be a problem when we implement it with the collapsable menu.

Screen Shot 2022-01-04 at 2 14 54 PM

josueetcom: I think setting a limit is fine. Let's say maybe 5 or 10?

src/components/MajorFilter.jsx
```diff
@@ -0,0 +1,66 @@
...
+        ))}
+        <Autocomplete
+          multiple
+          open={isOpen}
+          value={majors}
+          onChange={(e, val) => setMajors(val)}
+          onInputChange={(e, val) => {
+            if (val.length > 0) setIsOpen(true);
+            else setIsOpen(false);
+          }}
```
josueetcom, line 47: Get rid of boolean zen.
```suggestion
          onInputChange={(e, val) => setIsOpen(val.length > 0)}
```

Also your original code had a bug as there was a `;` before the `else`.

src/components/MajorFilter.jsx
```diff
@@ -0,0 +1,66 @@
...
+            variant="outlined"
+            color="primary"
+            key={major}
+            onDelete={() =>
+              setMajors((chips) => chips.filter((chip) => chip !== major))
+            }
+            sx={{ mx: 1, mt: 1, color: '#000' }}
+          />
+        ))}
+        <Autocomplete
```
josueetcom, line 39: Add `freeSolo`?

src/components/FilterBar.jsx
```diff
@@ -24,9 +19,7 @@ export default function FilterBar({ setQueryParam, queryParams }) {
       justifyContent="space-between"
       sx={{ flexGrow: 1 }}>
       <Grid item>
-        {experiments.expShowMajorFilter && (
-          <FilterDropDown label="Major" items={majors} />
```
josueetcom, line 28: Hmm we're not using this component anymore. I wonder if there's any value in keeping it.

src/components/MajorFilter.jsx
```diff
@@ -0,0 +1,66 @@
...
+            else setIsOpen(false);
+          }}
+          options={[...MAJORS]}
+          renderTags={() => null}
+          renderInput={(params) => (
+            <TextField
+              ref={params.InputProps.ref}
+              inputProps={params.inputProps}
+              autoFocus
+              placeholder="Filter Majors"
```
josueetcom, line 55:
```suggestion
              placeholder="Enter a major to filter by..."
```

src/components/MajorFilter.jsx
josueetcom: Maybe instead we can initialize this to false and show that dropdown arrow that toggles showing/hiding the list? The ones in the [playground](https://mui.com/components/autocomplete/#playground) have a button on the right to hide.

josueetcom (dismissed):

src/components/FilterBar.jsx
edwingl10: removed it.

src/components/MajorFilter.jsx
edwingl10: Added the dropdown arrow, the only "problem" is that the arrow flips up even when the input has no options.

Screen Shot 2022-01-06 at 9 10 18 PM

edwingl10: Update:
removing`onInputChange` will avoid this bug but it will only show the options IF the drop down is clicked.

src/components/MajorFilter.jsx
```diff
@@ -0,0 +1,85 @@
...
+  IconButton,
+} from '@mui/material';
+import { MAJORS } from '../types/options';
+import ArrowDropDownIcon from '@mui/icons-material/ArrowDropDown';
+import ArrowDropUpIcon from '@mui/icons-material/ArrowDropUp';
+
+function MajorFilter() {
+  const [anchorEl, setAnchorEl] = useState(null);
+  const [majors, setMajors] = useState([]);
+  const [isOpen, setIsOpen] = useState(false);
```
josueetcom, line 17: How about initializing this to true?

josueetcom (approved):

src/components/MajorFilter.jsx
josueetcom: Or we can we make it such that `isOpen` changes when the user starts typing.

edwingl10: I added `onInputChange` back.

PR #850, opened 2022-01-05
Major filter functionality

### LINKED ISSUE #307: Major filter: functionality

edwingl10:
## Motivation and Context

Fixes #307 #309
Added support for param support and functionality.

## How Has This Been Tested?
local 

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-850.onrender.com/scholarships.
2. Select some majors to filter by

## Checklist:


edwingl10:
I noticed when inputting an invalid option for `grades` or `majors`, it prunes all the filter options even when there are valid values. 
E.g. dreamscholars/scholarships?grades[]=8&majors[]=test
will clear everything when it's suppose to keep `grades[]=8`

src/components/MajorFilter.jsx
```diff
@@ -30,17 +31,15 @@ function MajorFilter() {
...
+            onDelete={() => changeFn(majors.filter((m) => m !== major))}
             sx={{ mx: 1, mt: 1, color: '#000' }}
           />
         ))}
         <Autocomplete
           multiple
           open={isOpen}
           value={majors}
-          onChange={(e, val) => setMajors(val)}
+          onChange={(e, val) => changeFn(val)}
```
josueetcom, line 42: I'm noticing that if I select something already selected it deselects it. Maybe `options` should exclude anything already selected? You should be able to use set operations for this.

src/pages/ListScholarships.jsx
```diff
@@ -44,15 +45,15 @@ function ListScholarships() {
...
-    history.replace({ search: queryString.stringify(params) });
+    history.replace({ search: queryString.stringify(params, queryOptions) });
   };
 
   /**
    * when dealing with invalid values in an array
    * rather than pruning the entire param, we can prune out the invalid values
    */
   const replaceQueryParam = (index, newVal) => {
-    params.grades = newVal;
```
josueetcom, line 55: lol

src/pages/ListScholarships.jsx
```diff
@@ -44,15 +45,15 @@ function ListScholarships() {
...
     delete params[index];
-    history.replace({ search: queryString.stringify(params) });
+    history.replace({ search: queryString.stringify(params, queryOptions) });
   };
 
   /**
    * when dealing with invalid values in an array
    * rather than pruning the entire param, we can prune out the invalid values
    */
   const replaceQueryParam = (index, newVal) => {
```
josueetcom, line 55: Hmm `index` feels kind of weird. Could you rename these functions' `index` parameter to `field` or `key`?

src/pages/ListScholarships.jsx
```diff
@@ -96,13 +97,27 @@ function ListScholarships() {
     } else pruneQueryParam(qParams.GRADES);
   }
 
+  /**
+   * prunes inalid major values not present in the MAJORS set
```
josueetcom, line 101: No need to prune as we allow freeform major values.

src/models/Scholarships.ts
```diff
@@ -155,6 +156,11 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
                 data.requirements?.grades,
                 opts.grades
               ) &&
+              // major filter
+              requirementMatchesFilter(
+                data.requirements?.majors,
+                opts.majors
```
josueetcom, line 162: Something that just came to mind is how we match up free form values when case mismatches. Maybe we should transform each of these lists to lower case so that case doesn't matter? I think you need something like `.map(s => s.toLowerCase())`.

josueetcom: Also I'm realizing it's going to be hard to actually make these grade level and majors filters work when we don't have that data inputted yet for all scholarships. Something we should probably follow up on (I left a comment on Slack).

src/pages/ListScholarships.jsx
edwingl10: This was causing a frustrating bug 😫

src/pages/ListScholarships.jsx
edwingl10: got it.
just pruned if `majors` is not an array/empty array

src/pages/ListScholarships.jsx
```diff
@@ -96,13 +96,19 @@ function ListScholarships() {
     } else pruneQueryParam(qParams.GRADES);
   }
 
+  if (majors !== undefined) {
+    if (!Array.isArray(majors) || majors.length === 0 || majors[0] === '')
+      pruneQueryParam(qParams.MAJORS);
+  }
+
```
edwingl10, line 103: pruned empty array or if

src/pages/ListScholarships.jsx
```diff
@@ -96,13 +96,19 @@ function ListScholarships() {
     } else pruneQueryParam(qParams.GRADES);
   }
 
+  if (majors !== undefined) {
+    if (!Array.isArray(majors) || majors.length === 0 || majors[0] === '')
```
edwingl10, line 100: `majors[0] === ''` takes care of the case where `?majors[]=` is inputted in the URL

src/components/FilterBar.jsx
```diff
@@ -19,7 +19,13 @@ export default function FilterBar({ setQueryParam, queryParams }) {
       <Grid item>
         <MajorFilter
           majors={majors}
-          changeFn={(e) => setQueryParam(qParams.MAJORS, e)}
+          onSelect={(e) => setQueryParam(qParams.MAJORS, e)}
+          onDelete={(e) =>
```
josueetcom, line 23: `e` is normally a shorthand for `event` in listeners and the value is normally `e.target.value`. How about using `m` instead?

josueetcom (approved):

Issue #816, opened 2021-12-30
Header is narrower than rest of the page


josueetcom:
Somehow our header is narrower than the rest of the page which conflicts with Figma [design](https://www.figma.com/file/nwrW3UebcvSWyQZGb5ixNr/DreamScholar?node-id=711%3A15). Also looks a bit awkward IMO

![image](https://user-images.githubusercontent.com/8023233/147722722-f3871624-ec64-454f-b419-d1ce035a243c.png)

PR #851, opened 2022-01-06
Header bug fixes

### LINKED ISSUE #817: Header height is different when logged in vs. not logged in
**Describe the bug**
When I log in the header has a computed height of 80px. When I log out it's 72px.

**To Reproduce**
1. Go to https://dreamscholars.org/about
2. Log in.
3. Take note of the header and how the page shifts down slightly once your avatar loads.

Also reproducible when logging out.

**Expected behavior**
Consistent header height and no page shifts when logging in/out.

**Additional context**
Running this in Chrome on Chrome OS.

edwingl10:
## Motivation and Context

Fixes #817 #816 

## How Has This Been Tested?
local

## Types of changes

## Previewing Changes

1. Go to https://s4us-pr-NUMBER.onrender.com/about.
2. login
3. logout
4. header height should not change

## Checklist:


src/components/Header.jsx
```diff
@@ -137,17 +139,15 @@ function Header() {
             {BRAND_NAME.toUpperCase()}
           </MuiLink>
         </Grid>
-        <Grid item>
-          <HeaderNavMenu links={links} />
-        </Grid>
 
-        <Grid item>
-          <IconButton onClick={(e) => setAnchorEl(e.currentTarget)}>
+        <Grid item sx={{ height: 80 }}>
```
edwingl10, line 143: putting them under one `Grid item` prevents them from being spaced out like before.

src/components/Header.jsx
```diff
@@ -73,14 +73,16 @@ const AuthZoomButton = ({ t }) => {
             aria-controls="menu-appbar"
             aria-haspopup="true"
             onClick={(e) => setAnchorEl(e.currentTarget)}
-            color="inherit">
+            color="inherit"
+            sx={{ height: '100%', width: 64 }}>
```
edwingl10, line 77: `height` added to prevent slight vertical shift in links when logging in / logging out
`width` added to prevent horizontal shift in links when logging in / logging out

src/components/Header.jsx
josueetcom: Putting them together sounds good but I'm not entirely convinced that manually setting the height here is the best idea. The height difference is caused by the different styling properties on the buttons themselves. I think the issue is the padding up top because the Login **Button** takes up 36.5px + 24px = 60.5px in height but the Avatar takes up 56 + 24 = 80px in height. So I think the fix lies in fixing the height of the AuthZoomButton.

src/components/HeaderNavMenu.jsx
```diff
@@ -21,7 +21,8 @@ function HeaderNavMenu({ links }) {
       indicatorColor="primary"
       scrollButtons="auto"
       value={currentTab}
-      variant="scrollable">
+      variant="scrollable"
+      sx={{ display: 'inline-flex' }}>
```
josueetcom, line 25: What does this do?

edwingl10: It prevents the following. 
Screen Shot 2022-01-07 at 8 38 14 PM

meanwhile `display: inline` causes the following:
Screen Shot 2022-01-07 at 8 39 14 PM

josueetcom (approved):

PR #859, opened 2022-01-11
firestore validation rules

### LINKED ISSUE #572: Scholarship model: Firestore data validation rules
Currently our [Firestore rules](https://github.com/beyondhb1079/s4us/blob/main/firestore.rules) limit access to certain documents/collections based on auth. But once you have write access to a collection you can essentially write whatever you want.

It looks like we can use Firestore security rules for data validation purposes as well:
https://firebase.google.com/docs/rules/data-validation#firestore

Let's add security rules that enforce creates/writes are only allowed if required fields are set. This way, if we tinker around with the database from in the Firestore Console UI or programmatically, we're not accidentally missing required fields.

edwingl10:
## Motivation and Context

Fixes #572 

## How Has This Been Tested?
by running yarn test

## Types of changes

## Previewing Changes

2.

## Checklist:


firestore.rules
```diff
@@ -1,12 +1,52 @@
...
+      && (!('ethnicities' in docData.requirements) || docData.requirements.ethnicities is list)
+      && (!('majors' in docData.requirements) || docData.requirements.majors is list)
+      && (!('schools' in docData.requirements) || docData.requirements.schools is list)
+      && (!('grades' in docData.requirements) || docData.requirements.grades is list)
+      && (!('states' in docData.requirements) || docData.requirements.states is list)
+      && (!('genders' in docData.requirements) || docData.requirements.genders is list)
+    }
+
+    function validAuthor(docData){
+      return !('author' in docData) || docData.author is map
```
josueetcom, line 24: Author should be required

firestore.rules
```diff
@@ -1,12 +1,52 @@
...
+      return docData.name is string 
+      && validAmount(docData)
+      && docData.deadline is timestamp 
+      && docData.description is string 
+      && docData.website is string
+      // OPTIONAL FIELDS
+      && (!('organization' in docData) || docData.organization is string)
+      && (!('tags' in docData) || docData.tags is list)
+      && (!('dateAdded' in docData) || docData.dateAdded is timestamp)
+      && (!('lastModified' in docData) || docData.lastModified is timestamp)
```
josueetcom, line 39: These should be required

firestore.rules
```diff
@@ -1,12 +1,52 @@
...
+    }
+
     match /scholarships/{document=**} {
       allow read: if true;
     }
 
     match /scholarships/{scholarshipId} {
-      allow create: if request.auth != null;
+      allow create: if request.auth != null && validData(request.resource.data);
       allow update: if request.auth.uid == resource.data.author.id || request.auth.token.admin;
```
josueetcom, line 50: Should we add the check for update too?

firestore.rules
edwingl10: I'm assuming the id and email should also be required?

firestore.rules
```diff
@@ -1,13 +1,53 @@
 rules_version = '2';
 service cloud.firestore {
   match /databases/{database}/documents {
+    function validAmount(docData){
```
edwingl10, line 4: should I also add validation for `min/max` when the type is a certain type?

firestore.rules
```diff
@@ -1,13 +1,53 @@
...
+      return docData.amount is map 
+      && docData.amount.type in ['UNKNOWN', 'FIXED','VARIES','FULL_TUITION']
+      && docData.amount.min is int
+      && docData.amount.max is int
+    }
+
+    function validReqs(docData){
+      return !('requirements' in docData)
```
josueetcom, line 12: Might be nice to ensure this is always set even if empty (though not sure of current db state).

firestore.rules
```diff
@@ -1,13 +1,53 @@
...
+      && docData.amount.min is int
+      && docData.amount.max is int
+    }
+
+    function validReqs(docData){
+      return !('requirements' in docData) 
+      || docData.requirements is map
+      && (!('gpa' in docData.requirements) || docData.requirements.gpa is number && docData.requirements.gpa > 0 && docData.requirements.gpa <= 4)
+      && (!('ethnicities' in docData.requirements) || docData.requirements.ethnicities is list)
+      && (!('majors' in docData.requirements) || docData.requirements.majors is list)
```
josueetcom, line 16: I wonder if you can refactor these "not set or is list" checks but that probably requires being able to do something like `reqs[someKey]`. However if you take in `docData.requirements` as simply `reqs` then maybe these lines won't look too long.

firestore.rules
josueetcom: I wouldn't worry about it. Something you could do here (and elsewhere) though is take in `amount` directly so you don't need to do `docData.amount`.

firestore.rules
edwingl10: I check the db and the one that were added via google sheets don't have a`requirements` field

josueetcom: Ah right, I did it that way so that when we actually edit them we remember to set the requirements. Alright in a few weeks then.

src/firestore.rules.test.js
```diff
@@ -157,10 +163,186 @@ test('allows scholarship delete when user is admin', async () => {
...
+    });
+
+    test('fails when scholarship genders is not an array', () => {
+      scholarship.requirements.genders = 'Male';
+      return assertFails(collection.set(scholarship));
+    });
+  });
+});
+
+describe('validation rules - update', () => {
```
josueetcom, line 304: These don't seem specific to update? You can group these with the other one. I don't think we need to have two sets of tests for update and create. Just one set and maybe a test case that tests update goes through the same validation failure process

josueetcom (approved):

src/firestore.rules.test.js
```diff
@@ -171,76 +171,186 @@ test('allows scholarship delete when user is admin', async () => {
   );
 });
 
-describe('validation rules - create', () => {
-  const collection = aliceApp.firestore().collection('scholarships').doc();
+describe('scholarship alidation rules', () => {
```
josueetcom, line 174: *validation

src/firestore.rules.test.js
```diff
@@ -171,76 +171,186 @@ test('allows scholarship delete when user is admin', async () => {
...
 
-  test('fails when no scholarship name', () => {
-    const scholarship = { ...newScholarship };
-    delete scholarship.name;
+  test('fails create when name not a string', () => {
+    return assertFails(collection('abc').set({ ...newScholarship, name: 34 }));
+  });
 
-    return assertFails(collection.set(scholarship));
+  test('fails update when name not a string', async () => {
```
josueetcom, line 182: Oh, I didn't mean make a corresponding update test for each one. But rather just make sure `validData()` is run for update operations (could've just been a single test).

But if you wanted to do this you could also probably refactor this so that firestore is prepopulated with an existing scholarship and just change the scholarship used based on whether it's existing or not. There are probably other refactoring opportunities for similar scholarships.

But these refactors are not really high priority right now. Better to address them later when we convert everything to TypeScript.

josueetcom (approved):

src/firestore.rules.test.js
```diff
@@ -157,10 +163,290 @@ test('allows scholarship delete when user is admin', async () => {
...
+
+    test('fails update when gpa is greater than 4', async () => {
+      await assertSucceeds(collection('gp').set(newScholarship));
+      return assertFails(
+        collection('gp').set({ ...newScholarship, requirements: { gpa: 4.1 } })
+      );
+    });
+
+    test('fails create when ethnicities not an array', () => {
+      scholarship.requirements.ethnicities = '[]';
```
josueetcom, line 423: Btw it's dangerous to set a global variable from within a test as it could impact other tests. Other tests may fail/pass unexpectedly because of it.

src/firestore.rules.test.js
```diff
@@ -157,10 +163,290 @@ test('allows scholarship delete when user is admin', async () => {
...
+      return assertFails(collection.set(scholarship));
```
josueetcom: `scholarship` seems to be undefined, do you mean to do `{...newScholarship, requirements: {ethnicities: '[]'}`?

firestore.rules
```diff
@@ -1,13 +1,51 @@
...
+
+    function validReqs(reqs){
+      return reqs is map
+      && (!('gpa' in reqs) || reqs.gpa is number && reqs.gpa > 0 && reqs.gpa <= 4)
+      && (!('ethnicities' in reqs) || reqs.ethnicities is list)
+      && (!('majors' in reqs) || reqs.majors is list)
+      && (!('schools' in reqs) || reqs.schools is list)
+      && (!('grades' in reqs) || reqs.grades is list)
+      && (!('states' in reqs) || reqs.states is list)
+      && (!('genders' in reqs) || reqs.genders is list)
```
josueetcom, line 18: This is definitely much cleaner to look at!

firestore.rules
```diff
@@ -1,13 +1,51 @@
 rules_version = '2';
 service cloud.firestore {
   match /databases/{database}/documents {
+    function validAmount(amount){
```
josueetcom, line 4: nit: here and elsewhere
```suggestion
    function validAmount(amount) {
```

src/firestore.rules.test.js
edwingl10: changed!

src/firestore.rules.test.js
edwingl10: lol cool I'll leave as is for now.

src/firestore.rules.test.js
edwingl10: I got rid of the global variable and ended up doing this

Issue #858, opened 2022-01-11
Show errors for editable scholarships


josueetcom:
At the bottom of the ScholarshipCard we should add a red/yellow/orange section that contains a list of errors we think the scholarship has based on heuristics from the description.

We should show an error if the description:
- [x] Contains what seems to be an outdated school year.
- [ ] Is a single line when it looks like there should be more (e.g. `" - "` appears multiple times).
- [x] Contains dollar amounts  **but** does not contain the min/max specified.
- [ ] Contains a date/deadline **but** does not contain the specified deadline.
- [x] Contains grade level keywords (e.g. "freshmen", "1st year", "12th grade") **but** no grade level requirements.
- [x] Contains the word `GPA` or `Grade Point Average` **but** no GPA requirement is set.
- [x] Contains one of the schools we list **but** no school requirements.
- [x] Contains a state or state abbreviation **but** no state requirements.
- [x] Contains a major  **but** no majors requirements.
- [x] Contains ethnicity keywords (e.g. "hispanic", "black") **but** no ethnicity requirements.

PR #863, opened 2022-01-13
Add lint check for min GPA

josueetcom:
## Motivation and Context

Addresses one of tasks in #858.

## How Has This Been Tested?

Added tests for the new files. The card tests are missing but the card view has been tested locally.

## Types of changes

## Previewing Changes

1. Go to Render preview URL.
2. Try adding/editing a scholarship so that the minimum GPA is mentioned in the description.
3. See the error on the card.
4. Log out and make sure you can't see the error.

## Checklist:

### DIFF
--- a/src/components/ScholarshipCard.jsx
+++ b/src/components/ScholarshipCard.jsx
@@ -110,7 +110,7 @@ export default function ScholarshipCard({ scholarship, style }) {
   }, [author, currentUser, preview]);
 
   const CardAreaComponent = detailed ? Box : CardActionArea;
-  const lintIssues = canEdit ? lint(scholarship.data) : [];
+  const lintIssues = canEdit || preview ? lint(scholarship.data) : [];
   return (
     <Card variant="outlined">
       <CardAreaComponent
--- a/src/components/ScholarshipCard.jsx
+++ b/src/components/ScholarshipCard.jsx
@@ -286,12 +286,10 @@ export default function ScholarshipCard({ scholarship, style }) {
           <AlertTitle>
             <strong>{lintIssues.length} potential issues detected</strong>
           </AlertTitle>
-          <Box>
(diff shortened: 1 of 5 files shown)


src/components/ScholarshipCard.jsx
```diff
@@ -107,6 +110,7 @@ export default function ScholarshipCard({ scholarship, style }) {
   }, [author, currentUser, preview]);
 
   const CardAreaComponent = detailed ? Box : CardActionArea;
+  const lintIssues = canEdit ? lint(scholarship.data) : [];
```
edwingl10, line 113: It would also be nice to see the warnings when adding a scholarship.

src/lib/lint.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      'GPAs of 3.0-3.5 encouraged to apply.',
+    ].forEach((d) => expect(parseMinGPA(d)?.value).toBe('3.0')));
+
+  test('matches minimum GPA value', () =>
+    [
+      'Min GPA of 3.0 out of 4.0',
+      'Min GPA: High school 3.5, College 3.0',
+      'Must have a GPA of 3.0 out of 4.0 or 4.0 out of 5.0',
+      'GPAs of 3.0-3.5 encouraged to apply.',
+    ].forEach((d) => expect(parseMinGPA(d)?.value).toBe('3.0')));
```
edwingl10, line 39: same test as above

edwingl10, line 39: same test as above

edwingl10 (approved):

src/components/ScholarshipCard.jsx
josueetcom: oh, I hadn't noticed it didn't appear for new scholarships. Fixed.

PR #875, opened 2022-01-25
Refactor filters

edwingl10:
## Motivation and Context

Refactors filters since they had the same `Button` and `Popover` logic. 

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. go to scholarships page
3. fitlers should look & behave the same

### DIFF
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -34,7 +34,7 @@ const Filter = (props) => {
         onClose={() => setAnchorEl(null)}
         anchorOrigin={{ vertical: 'bottom', horizontal: 'left' }}
         transformOrigin={{ vertical: 'top', horizontal: 'left' }}
-        PaperProps={{ style: { width: 280 } }}>
+        PaperProps={{ style: { minWidth: 280 } }}>
         {filter}
       </Popover>
     </>
@@ -77,11 +77,20 @@ export default function FilterBar({ setQueryParam, queryParams }) {
           }
         />
 
-        <AmountFilter
-          min={minAmount ?? 0}
-          max={maxAmount ?? 0}
-          onMinChange={(e) => setQueryParam(qParams.MIN_AMOUNT, e.target.value)}
-          onMaxChange={(e) => setQueryParam(qParams.MAX_AMOUNT, e.target.value)}
+        <Filter
+          title="Amount"
+          filter={
+            <AmountFilter
+              min={minAmount ?? 0}
+              max={maxAmount ?? 0}
+              onMinChange={(e) =>
+                setQueryParam(qParams.MIN_AMOUNT, e.target.value)
+              }
+              onMaxChange={(e) =>
+                setQueryParam(qParams.MAX_AMOUNT, e.target.value)
+              }
+            />
+          }
         />
       </Grid>
 
--- a/src/components/FilterBar.jsx
+++ b/src/components/FilterBar.jsx
@@ -56,19 +56,24 @@ export default function FilterBar({ setQueryParam, queryParams }) {
       justifyContent="space-between"
       sx={{ flexGrow: 1 }}>
       <Grid item>
-        <MajorFilter
-          majors={majors}
-          onSelect={(m) => setQueryParam(qParams.MAJORS, m)}
-          onDelete={(m) =>
-            setQueryParam(
-              qParams.MAJORS,
-              majors.filter((major) => major !== m)
-            )
+        <Filter
+          title="Majors"
+          filter={
+            <MajorFilter
+              majors={majors}
+              onSelect={(m) => setQueryParam(qParams.MAJORS, m)}
+              onDelete={(m) =>
+                setQueryParam(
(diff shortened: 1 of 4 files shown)


src/components/FilterBar.jsx
```diff
@@ -17,27 +55,46 @@ export default function FilterBar({ setQueryParam, queryParams }) {
...
-          majors={majors}
-          onSelect={(m) => setQueryParam(qParams.MAJORS, m)}
-          onDelete={(m) =>
-            setQueryParam(
-              qParams.MAJORS,
-              majors.filter((major) => major !== m)
-            )
+        <Filter
+          title="Majors"
+          filter={
```
josueetcom, line 60: Instead of passing a property you can just embed this like normal HTML and read it as the `children` prop.

https://codeburst.io/a-quick-intro-to-reacts-props-children-cb3d2fce4891

src/components/FilterBar.jsx
```diff
@@ -1,11 +1,49 @@
...
+        sx={{ m: 1, height: (theme) => theme.spacing(4) }}>
+        {title}
+      </Button>
+      <Popover
+        open={Boolean(anchorEl)}
+        anchorEl={anchorEl}
+        onClose={() => setAnchorEl(null)}
+        anchorOrigin={{ vertical: 'bottom', horizontal: 'left' }}
+        transformOrigin={{ vertical: 'top', horizontal: 'left' }}
+        PaperProps={{ style: { minWidth: 280 } }}>
```
josueetcom, line 36: Hmm, for the major filters this makes it take up an arbitrary width. I suggest either:
- Setting a `maxWidth` OR
- Stacking major selections on top of each other

src/components/FilterBar.jsx
```diff
@@ -1,11 +1,49 @@
...
+  Popover,
+} from '@mui/material';
 import AmountFilter from './AmountFilter';
 import GradeLevelFilter from './GradeLevelFilter';
 import qParams from '../lib/QueryParams';
 import sortOptions, { DEADLINE_ASC } from '../lib/sortOptions';
 import MajorFilter from './MajorFilter';
+import ArrowDropDownIcon from '@mui/icons-material/ArrowDropDown';
+
+const Filter = ({ title, filter }) => {
```
josueetcom, line 18: Maybe `FilterButton`? Since it isn't the same type of component as other `*Filter` ones.

josueetcom (changes requested):

src/components/FilterBar.jsx
edwingl10: The reason I made it this width was so that you can seel the Input's placeholder.

edwingl10: Screen Shot 2022-01-25 at 9 29 33 AM

VS

Screen Shot 2022-01-25 at 9 30 07 AM

src/components/FilterBar.jsx
```diff
@@ -34,15 +35,14 @@ const Filter = ({ title, filter }) => {
...
+        {props.children}
       </Popover>
     </>
   );
 };
 
-Filter.propTypes = {
+FilterButton.propTypes = {
   title: PropTypes.string.isRequired,
-  filter: PropTypes.node.isRequired,
```
josueetcom: You'll want to keep this but rename it to `children`.

src/components/FilterBar.jsx
```diff
@@ -15,7 +15,8 @@ import sortOptions, { DEADLINE_ASC } from '../lib/sortOptions';
 import MajorFilter from './MajorFilter';
 import ArrowDropDownIcon from '@mui/icons-material/ArrowDropDown';
 
-const Filter = ({ title, filter }) => {
+const FilterButton = (props) => {
+  const { title } = props;
```
josueetcom, line 19: You can still use the other form
```suggestion
const FilterButton = ({title, children}) => {
```

src/components/FilterBar.jsx
josueetcom: Sorry I meant the width is arbitrary when multiple majors are selected:
Screen Shot 2022-01-25 at 5 37 06 PM

I think a quick fix would be to additionally specify `maxWidth` to something reasonable that fits the needed widths for the other filters. 

Or actually, since only the major filter needs this, you could add an optional `width` (or `popoverWidth` since it's the width of the popover) property to the `FilterButton` component that defaults `undefined`.

josueetcom (approved):

src/components/FilterBar.jsx
```diff
@@ -55,7 +60,7 @@ export default function FilterBar({ setQueryParam, queryParams }) {
       justifyContent="space-between"
       sx={{ flexGrow: 1 }}>
       <Grid item>
-        <FilterButton title="Majors">
+        <FilterButton title="Majors" popoverWidth={{ minWidth: 280 }}>
```
josueetcom, line 63: Not minWidth. You could've made `popoverWidth` a number.

Issue #878, opened 2022-01-26
Consider using useSearchParams for retrieving/setting query string


josueetcom:
https://reactrouter.com/docs/en/v6/api#usesearchparams

PR #884, opened 2022-01-28
Implement useQueryParams hook

### LINKED ISSUE #532: Investigate simplifying query param logic using query-params
The way we added query string parameter parsing/setting in #355 is maybe not the most ideal way of doing things. Perhaps a solution like [use-query-params](https://www.npmjs.com/package/use-query-params) might be better suited for this.

1. Investigate what it takes to rewrite query params business logic using `use-query-params`
2. If the code is simpler, send out a PR.
3. If the code is more complicated, leave a comment on this bug detailing why it's more complicated (give examples) and close this issue. Be sure to bring it up at the next meeting.

I imagine the former will probably be the case.

josueetcom:
## Motivation and Context

Fixes #532 and #878 by providing a custom hook better suited for our needs. 

## Types of changes

## How Has This Been Tested?

## Previewing Changes

1. Go to Scholarships page
2. Try setting filters and putting bad values in the URL for known keys.

### DIFF
--- a/src/lib/useQueryParams.ts
+++ b/src/lib/useQueryParams.ts
@@ -1,0 +1,110 @@
+import queryString, { ParseOptions } from 'query-string';
+import { useEffect } from 'react';
+import { useLocation, useNavigate } from 'react-router-dom';
+import GradeLevel from '../types/GradeLevel';
+import sortOptions from './sortOptions';
+
+const options: ParseOptions = {
+  arrayFormat: 'bracket-separator',
+  arrayFormatSeparator: ',',
+  parseNumbers: true,
+};
+
+type SetQueryParamFn = (
+  key: 'minAmount' | 'maxAmount' | 'grades' | 'majors' | 'sortBy',
+  value: any,
+  replace: boolean
+) => void;
+
+/**
+ * Returns the query string parsed as an object, and a function to update individual parameters.
+ * @example
+ * const [params, setQueryParam] = useQueryParam();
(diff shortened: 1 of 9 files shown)


src/lib/useQueryParams.ts
```diff
@@ -0,0 +1,110 @@
...
+import sortOptions from './sortOptions';
+
+const options: ParseOptions = {
+  arrayFormat: 'bracket-separator',
+  arrayFormatSeparator: ',',
+  parseNumbers: true,
+};
+
+type SetQueryParamFn = (
+  key: 'minAmount' | 'maxAmount' | 'grades' | 'majors' | 'sortBy',
```
josueetcom, line 14: I removed `qParams` in favor of this. As we move things to Typescript this'll prevent `setQueryParam` from being called with an unknown key.

src/lib/useQueryParams.ts
```diff
@@ -0,0 +1,110 @@
...
+          ? Array.from(new Set(majors)).filter((s) => s.length > 0)
+          : [];
+        if (params.majors.length === 0) {
+          delete params.majors;
+        }
+      }
+
+      if (JSON.stringify(params) !== JSON.stringify(origParams)) {
+        // Navigate to the current page with the params cleaned up
+        navigate(
```
josueetcom, line 99: One optimization I did here is to avoid navigating until we prune out everything we want to prune out.

edwingl10 (approved):

PR #887, opened 2022-01-30
useAuth hook

### LINKED ISSUE #319: Refactor Auth getting/managing
**Is your feature request related to a problem? Please describe.**
We're going to have a lot of code that conditionally depends on the Auth state. Plus, there's like [3 auth states](https://medium.com/firebase-developers/why-is-my-currentuser-null-in-firebase-auth-4701791f74f0) :(.

**Describe the solution you'd like**
How about [useAuth](https://usehooks.com/useAuth/)? That way the auth state management can be done in one place.

**Describe alternatives you've considered**
We could continue the way we're doing it but it means remembering to distinguish a third value from `true`/`false` (e.g. `undefined`). 

**Additional context**
PR #292 introduces protected routes, which need access to auth state. Additionally, showing/hiding the Login dialog needs to be detached from the Login button. Future redirects/routes may need access as well.

josueetcom:
## Motivation and Context

Fixes #319 by providing a contextual hook that returns the current user and claims for them.

## Types of changes

## How Has This Been Tested?


src/pages/Home.jsx
```diff
@@ -1,21 +1,12 @@
...
 export default function Home() {
-  const [isSignedIn, setIsSignedIn] = useState(undefined);
-  const loading = isSignedIn === undefined;
+  const { currentUser: user } = useAuth();
 
-  useEffect(
-    () => firebase.auth().onAuthStateChanged((user) => setIsSignedIn(!!user)),
-    []
-  );
+  if (user === undefined) return '';
```
josueetcom, line 9: Decided to do this as the currentUser gets resolved pretty quickly. If we use a progress indicator here it barely flashes even on a slow 3G connection.

edwingl10 (approved):

PR #888, opened 2022-01-30
Filter panel styling

edwingl10:
## Motivation and Context

Filter panel styling to reflect figma design

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. go to scholarships page
3. check out filter panel


src/pages/ListScholarships.jsx
```diff
@@ -65,7 +64,7 @@ function ListScholarships() {
         <Typography variant="h4" component="h1" align="center" style={{ p: 1 }}>
           {t('general.scholarships')}
         </Typography>
-        <FilterBar />
```
edwingl10, line 68: was making tests fail since I made `onSliderchange` reauired in `AmountFilter`

src/components/AmountFilter.jsx
```diff
@@ -1,17 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { InputLabel, Grid } from '@mui/material';
+import { InputLabel, Grid, Slider } from '@mui/material';
 import { Remove as RemoveIcon } from '@mui/icons-material';
 import AmountTextField from './AmountTextField';
 
 export default function AmountFilter(props) {
-  const { min, max, onMinChange, onMaxChange } = props;
+  const { min, max, onMinChange, onMaxChange, onSliderChange } = props;
```
josueetcom, line 8: Why not make onSliderChange a function of onMinChange and onMaxChange? The parent component doesn't need to know how min and max filters are implemented underneath. The parent only needs to be notified of changes in min or max

src/components/AmountFilter.jsx
```diff
@@ -11,20 +11,34 @@ export default function AmountFilter(props) {
...
+          value={max || ''}
+          onChange={(e) => onMaxChange(e.target.value)}
+        />
       </Grid>
+
+      <Slider
+        value={[min, max]}
+        min={0}
+        max={10000}
+        step={100}
```
josueetcom, line 36: I just noticed this step value is inconsistent with the one in the AmountTextField

src/components/AmountFilter.jsx
```diff
@@ -11,20 +11,34 @@ export default function AmountFilter(props) {
...
+        onChange={(e) => {
+          onMinChange(e.target.value[0]);
+          onMaxChange(e.target.value[1]);
```
josueetcom, line 39: I'm noticing that if you drag this slider around the page becomes slow because it's always updating the URL and trying to refresh the scholarship list. Maybe before we add the slider we should implement the **Apply Filters** button at the bottom of the filter panel?

src/components/AmountFilter.jsx
```diff
@@ -11,20 +11,34 @@ export default function AmountFilter(props) {
...
-        <AmountTextField value={max || ''} onChange={onMaxChange} />
+        <InputLabel sx={{ pl: 1 }}>Max</InputLabel>
+        <AmountTextField
+          value={max || ''}
+          onChange={(e) => onMaxChange(e.target.value)}
+        />
       </Grid>
+
+      <Slider
+        value={[min, max]}
```
josueetcom, line 33: The initial max value of the slider isn't on the far right.

src/components/AmountTextField.jsx
```diff
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 
 const AmountTextField = (props) => (
   <TextField
-    sx={{ maxWidth: 130 }}
+    sx={{ maxWidth: 120 }}
     variant="outlined"
     placeholder="Unknown"
```
josueetcom, line 9: Instead of always using `Unknown` how about accepting an optional `placeholder` so this can be something like "0" or "Full Tuition" when the filter uses this?

src/components/AmountFilter.jsx
```diff
@@ -11,20 +11,34 @@ export default function AmountFilter(props) {
   return (
     <Grid container>
       <Grid item sx={{ m: 1 }}>
```
josueetcom, line 13: The slider isn't center aligned with the textfields. You can [auto lay it out](https://mui.com/components/grid/#auto-layout) so that there is `space-between` the field, element and dash and get rid of the margins here. Instead of margins use Grid spacing and set the `xs` property accordingly.

If you do need to add margin/padding add it to the Grid container or the `AccordionDetails` element so that the entire filter area (including the slider below) gets spacing to the sides.

src/components/GradeLevelFilter.jsx
```diff
@@ -4,12 +4,42 @@ import {
...
+        <FormControlLabel
+          key={grade}
+          control={
+            <Checkbox
+              color="primary"
+              checked={grades.has(grade)}
+              onChange={() => toggleSelection(grade)}
+            />
+          }
+          label={GradeLevel.toString(grade)}
```
josueetcom, line 26: You could strip out some of the extra words, e.g. "Graduate ", "College ".

src/components/AmountFilter.jsx
edwingl10: I'll hide it behind an experiement guard

src/components/AmountTextField.jsx
edwingl10: when would we display `full tuition`?
On the figma design it shows it when the max input is empty but that might be confusing?

src/components/AmountFilter.jsx
```diff
@@ -1,30 +1,49 @@
...
+        <InputLabel>Max</InputLabel>
+        <AmountTextField
+          value={max || ''}
+          onChange={(e) => onMaxChange(e.target.value)}
+          placeholder="No max"
+        />
       </Grid>
+
+      {experiments.expShowSlider && (
+        <Slider
```
josueetcom, line 36: nit: `sx={{mt: 1}}`

josueetcom (approved):

PR #889, opened 2022-01-31
filter panel for mobile

edwingl10:
## Motivation and Context

Makes Filter panel accssible for mobile.

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. go to scholarships page
3. change oto mobile view and open filter panel


src/pages/ListScholarships.jsx
```diff
@@ -37,21 +45,22 @@ function ListScholarships() {
...
-          display: { xs: 'none', md: 'block' },
           flexShrink: 0,
           width: drawerWidth,
           '& .MuiDrawer-paper': {
             width: drawerWidth,
             position: 'absolute',
             boxSizing: 'border-box',
           },
-          position: 'sticky',
+          position: isDesktop ? 'sticky' : 'static',
```
josueetcom, line 55:
```suggestion
          position: {xs: 'static', md: 'sticky'},
```

src/pages/ListScholarships.jsx
```diff
@@ -1,18 +1,26 @@
...
 import Scholarships from '../models/Scholarships';
 
-const drawerWidth = 360;
-
 function ListScholarships() {
   const { t } = useTranslation();
+  const [drawerOpen, setDrawerOpen] = useState(false);
+
+  const isDesktop = useMediaQuery((theme) => theme.breakpoints.up('md'));
+  const drawerWidth = isDesktop ? 360 : '100%';
```
josueetcom, line 23: You could leave drawerWidth as is and set it to `{xs: '100%', md: 360}`

src/pages/ListScholarships.jsx
```diff
@@ -37,21 +45,22 @@ function ListScholarships() {
...
             position: 'absolute',
             boxSizing: 'border-box',
           },
-          position: 'sticky',
+          position: isDesktop ? 'sticky' : 'static',
           overflowY: 'auto',
         }}
-        variant="permanent"
+        open={drawerOpen || isDesktop}
+        variant={isDesktop ? 'permanent' : 'temporary'}
```
josueetcom, line 59: Do you know if this works?
```suggestion
        variant={{xs: 'temporary', md: 'permanent'}}
```

edwingl10: nope, I get the following error 
 `Failed prop type: Invalid prop `variant` of value `[object Object]` supplied to `ForwardRef(Drawer)`, expected one of ["permanent","persistent","temporary"].`

src/components/FilterPanel.jsx
```diff
@@ -43,8 +49,23 @@ export default function FilterPanel() {
...
     ),
   };
+
   return (
     <Box>
+      <Stack
+        direction="row"
+        justifyContent="space-between"
+        alignItems="center"
+        sx={{ display: { md: 'none' }, my: 1 }}>
```
josueetcom, line 59: Instead of a `<Stack>` I suggest using a [`<Toolbar>`](https://mui.com/api/toolbar/) instead. It should come with the margins you need and display things in a row.

src/components/FilterPanel.jsx
```diff
@@ -60,3 +81,7 @@ export default function FilterPanel() {
     </Box>
   );
 }
+
+FilterPanel.propTypes = {
+  closePanel: PropTypes.func.isRequired,
```
josueetcom, line 86: nit: `onClose` might be a better name. Handlers tend to be named `on` some action. `closePanel` could be misinterpreted as a boolean.

src/pages/ListScholarships.jsx
```diff
@@ -56,36 +58,33 @@ function ListScholarships() {
...
           flexGrow: 1,
-          padding: 2,
           position: 'sticky',
           overflowY: { md: 'auto' },
         }}>
-        <Typography variant="h4" component="h1" align="center" style={{ p: 1 }}>
-          {t('general.scholarships')}
-        </Typography>
-        <FilterBar />
+        <FilterBar openFilter={() => setDrawerOpen(true)} />
```
josueetcom, line 87: The surrounding container introduces margins to this. How about sticking it outside of the container? Though to make it scrollable with the rest of the main content you'll need to surround both this and the container inside a box that has `position: 'sticky'` and `overflowY` set instead of the container here.

josueetcom: Also to keep the filter bar contents aligned with the container it'll need to inside of a container with the same `maxWidth` itself. So in summary
```
Scrollable Box
  - FilterBar
    - Container md
      - buttons
  - Container md
    - ScholarshipList
```

src/components/FilterBar.jsx
```diff
@@ -1,112 +1,72 @@
...
-            onSelect={(m) => setQueryParam('majors', m)}
-            onDelete={(m) =>
-              setQueryParam(
-                'majors',
-                majors.filter((major) => major !== m)
-              )
-            }
-          />
-        </FilterButton>
+    <Box sx={{ pb: 3 }}>
```
josueetcom, line 21: As I mention in one of the other comments, I suggest using a [`<Toolbar>`](https://mui.com/api/toolbar/) here instead.

src/components/FilterBar.jsx
```diff
@@ -1,112 +1,72 @@
...
+        direction="row-reverse"
+        sx={{ flexGrow: 1, bgcolor: 'primary.main', p: 2 }}>
+        <Grid item sx={{ display: 'flex', alignItems: 'center' }}>
+          <IconButton
+            onClick={(e) => setAnchorEl(e.currentTarget)}
+            color="secondary">
+            <ImportExportIcon />
+          </IconButton>
+          <Typography color="secondary">Sort</Typography>
+        </Grid>
```
josueetcom, line 35: How about using a [button with an icon](https://mui.com/components/buttons/#buttons-with-icons-and-label) instead of this?

src/components/FilterBar.jsx
```diff
@@ -1,112 +1,72 @@
...
-        <FilterButton title="Amount">
-          <AmountFilter
-            min={minAmount ?? 0}
-            max={maxAmount ?? 0}
-            onMinChange={(e) => setQueryParam('minAmount', e.target.value)}
-            onMaxChange={(e) => setQueryParam('maxAmount', e.target.value)}
-          />
-        </FilterButton>
+          <Typography color="secondary">Filter</Typography>
+        </Grid>
```
josueetcom, line 45: Again I think a [button with an icon](https://mui.com/components/buttons/#buttons-with-icons-and-label) keeps this simple.

josueetcom (changes requested):

src/components/AmountFilter.jsx
```diff
@@ -10,7 +10,7 @@ export default function AmountFilter(props) {
   const error = max > 0 && max < min;
 
   return (
-    <Grid container justifyContent="space-between">
+    <Grid container spacing={2}>
```
edwingl10, line 13: change necessary for mobile sreens when the filter panel takes the entire width of the screen. Before the `Textbox`s were too spaced out and would span entire width.

edwingl10: Screen Shot 2022-02-02 at 6 07 17 PM

src/components/AmountFilter.jsx
```diff
@@ -20,7 +20,7 @@ export default function AmountFilter(props) {
           placeholder="No min"
         />
       </Grid>
-      <Grid item sx={{ pt: 2, display: 'flex', alignItems: 'center' }}>
+      <Grid item sx={{ mt: 2, display: 'flex', alignItems: 'center' }}>
```
edwingl10, line 23: changed to center the `-`

src/components/FilterBar.jsx
```diff
@@ -1,112 +1,58 @@
...
-        onClick={(e) => setAnchorEl(e.currentTarget)}
-        endIcon={<ArrowDropDownIcon color="primary" />}
-        sx={{ m: 1, height: (theme) => theme.spacing(4) }}>
-        {title}
-      </Button>
-      <Popover
+    <Toolbar sx={{ bgcolor: 'primary.main' }}>
+      <Container
+        maxWidth="md"
+        sx={{ display: 'flex', justifyContent: 'space-between' }}>
```
edwingl10, line 17: sadly once I surrouned the buttons with the `Container`, the `justifyContent` in the `Toolbar` didn't work

edwingl10: also to keep the sort button on the left even when the filter button is hidden, I have to set the direction to `row-reverse` and put the sort button first

src/components/MajorFilter.jsx
```diff
@@ -40,6 +40,7 @@ function MajorFilter({ majors, onSelect, onDelete }) {
                 {isOpen ? <ArrowDropUpIcon /> : <ArrowDropDownIcon />}
               </IconButton>
             }
+            sx={{ maxWidth: 300 }}
```
edwingl10, line 43: added to avoid this
Screen Shot 2022-02-02 at 6 15 39 PM

src/pages/ListScholarships.jsx
```diff
@@ -56,38 +58,41 @@ function ListScholarships() {
...
+      <Box
         sx={{
-          bgcolor: 'background.default',
-          flexGrow: 1,
-          padding: 2,
-          position: 'sticky',
-          overflowY: { md: 'auto' },
+          width: '100%',
+          display: 'flex',
+          flexDirection: 'column',
```
edwingl10, line 81: allows `FilterBar` to scroll with the `scholarshipList`

src/pages/ListScholarships.jsx
```diff
@@ -56,38 +58,41 @@ function ListScholarships() {
...
-      </Container>
+        <FilterBar openFilter={() => setDrawerOpen(true)} />
+
+        <Container
+          maxWidth="md"
+          component="main"
+          ref={ref}
+          sx={{
+            flexGrow: 1,
+            overflowY: 'auto',
```
edwingl10, line 91: Didn't add to Box above so `FilterBar` could scroll with the list of scholarships.
Also made it auto so it could scroll when on mobile.

src/components/FilterPanel.jsx
```diff
@@ -47,6 +52,18 @@ export default function FilterPanel() {
...
+      <Toolbar
+        disableGutters
+        sx={{ display: { md: 'none' }, justifyContent: 'space-between' }}>
+        <IconButton onClick={onClose}>
+          <CloseIcon />
+        </IconButton>
+
+        <Typography>Filters</Typography>
+
+        <Button>Reset</Button>
```
josueetcom, line 64: This doesn't currently work.

Maybe disable the button for now?

src/components/MajorFilter.jsx
josueetcom: Seems fine to me. There are some pretty long majors like "Health information and medical records technology"

src/pages/ListScholarships.jsx
josueetcom: I don't see the FilterBar scrolling with the scholarships?
Screen Shot 2022-02-02 at 9 08 20 PM

I think instead of here you want to set `overflowY` in the parent `Box`

src/pages/ListScholarships.jsx
josueetcom: That's because the `display` was set to `flex`. You don't need `display` or `flexDirection` here. Just `overflowY: auto` suffices (as long as `overflowY` is removed from the Container below).

josueetcom (changes requested):

src/components/AmountFilter.jsx
edwingl10: to prevent the text fields from being spaced out too much

src/components/FilterBar.jsx
edwingl10: nice!

src/components/AmountFilter.jsx
```diff
@@ -10,7 +10,7 @@ export default function AmountFilter(props) {
   const error = max > 0 && max < min;
 
   return (
-    <Grid container spacing={2}>
+    <Grid container justifyContent="space-between" sx={{ maxWidth: 300 }}>
```
josueetcom, line 13: Interesting! I hadn't noticed how off it'd be on small to medium sized screens. Makes me wonder if we should change the transition breakpoint to [sm](https://mui.com/customization/breakpoints/#default-breakpoints) instead 🤔 .

That being said, maybe add `margin: auto`? I'm good either way.

josueetcom (approved):

PR #910, opened 2022-02-05
Apply filters button

### LINKED ISSUE #909: Filter Panel UX
Improve the user experience of the filter panel.

- [x] #911
- [x] #919
- [x] #913

edwingl10:
## Motivation and Context

Fixes #909 

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. Go to scholarships page
3. apply some filters
4. clear filters

### DIFF
--- a/src/components/FilterPanel.jsx
+++ b/src/components/FilterPanel.jsx
@@ -9,6 +9,8 @@ import {
   Typography,
   Button,
   Toolbar,
+  Badge,
+  Grid,
 } from '@mui/material';
 import useQueryParams from '../lib/useQueryParams';
 import AmountFilter from './AmountFilter';
@@ -28,47 +30,54 @@ export default function FilterPanel({ onClose }) {
   });
 
   const filters = {
-    Major: (
-      <MajorFilter
-        majors={filterVals.majors}
-        onSelect={(m) => setFilterVals({ ...filterVals, majors: m })}
-        onDelete={(m) =>
-          setFilterVals({
-            ...filterVals,
-            majors: filterVals.majors.filter((major) => major !== m),
-          })
-        }
-      />
-    ),
-    Amount: (
-      <AmountFilter
-        min={filterVals.minAmount ?? 0}
-        max={filterVals.maxAmount ?? 0}
-        onMinChange={(val) =>
-          setFilterVals({ ...filterVals, minAmount: parseInt(val) })
-        }
-        onMaxChange={(val) =>
-          setFilterVals({ ...filterVals, maxAmount: parseInt(val) })
-        }
-      />
-    ),
-    'Grade Level': (
-      <GradeLevelFilter
-        grades={new Set(filterVals.grades)}
-        changeFn={(e) => setFilterVals({ ...filterVals, grades: e })}
-      />
-    ),
+    Major: {
+      comp: (
+        <MajorFilter
+          majors={filterVals.majors}
+          onSelect={(m) => setFilterVals({ ...filterVals, majors: m })}
+          onDelete={(m) =>
+            setFilterVals({
+              ...filterVals,
+              majors: filterVals.majors.filter((major) => major !== m),
+            })
+          }
+        />
+      ),
+      badge: majors?.length,
+    },
+    Amount: {
+      comp: (
+        <AmountFilter
+          min={filterVals.minAmount ?? 0}
+          max={filterVals.maxAmount ?? 0}
+          onMinChange={(val) =>
+            setFilterVals({ ...filterVals, minAmount: parseInt(val) })
+          }
+          onMaxChange={(val) =>
+            setFilterVals({ ...filterVals, maxAmount: parseInt(val) })
+          }
+        />
+      ),
+      badge: minAmount && maxAmount ? 2 : minAmount || maxAmount ? 1 : 0,
+    },
+    'Grade Level': {
+      comp: (
+        <GradeLevelFilter
+          grades={new Set(filterVals.grades)}
+          changeFn={(e) => setFilterVals({ ...filterVals, grades: e })}
+        />
+      ),
+      badge: grades?.length,
+    },
   };
 
   function resetFilters() {
-    setQueryParams({
+    setFilterVals({
       minAmount: undefined,
       maxAmount: undefined,
       grades: undefined,
       majors: undefined,
     });
-
-    setFilterVals({});
   }
 
   return (
@@ -91,16 +100,28 @@ export default function FilterPanel({ onClose }) {
             expandIcon={<ExpandMoreIcon />}
             aria-controls={name + '-content'}
             id={name + '-header'}>
-            <Typography>{name}</Typography>
+            <Grid container alignItems="center" justifyContent="space-between">
+              <Typography>{name}</Typography>
+              <Badge
+                badgeContent={filter.badge}
+                color="primary"
+                sx={{ mr: 2 }}
+              />
+            </Grid>
           </AccordionSummary>
 
-          <AccordionDetails sx={{ m: 1 }}>{filter}</AccordionDetails>
+          <AccordionDetails sx={{ m: 1 }}>{filter.comp}</AccordionDetails>
         </Accordion>
       ))}
 
       <Toolbar
         sx={{ justifyContent: { xs: 'center', md: 'space-between' }, mt: 2 }}>
-        <Button variant="contained" onClick={() => setQueryParams(filterVals)}>
+        <Button
+          variant="contained"
+          onClick={() => {
+            setQueryParams(filterVals);
+            onClose();
+          }}>
           Apply Filters
         </Button>
         <Button
--- a/src/components/FilterPanel.jsx
+++ b/src/components/FilterPanel.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { ExpandMore as ExpandMoreIcon } from '@mui/icons-material';
 import {
   Accordion,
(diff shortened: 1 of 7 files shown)


src/components/FilterPanel.jsx
```diff
@@ -18,38 +20,66 @@ import CloseIcon from '@mui/icons-material/Close';
...
+      badge: grades?.length,
+    },
   };
 
+  function resetFilters() {
+    setFilterVals({
+      minAmount: undefined,
+      maxAmount: undefined,
+      grades: undefined,
+      majors: undefined,
```
edwingl10, line 79: set to `undefined` so they can be pruned. I could also write a separate function that clears the filters, that way I can set this to `{}`

src/components/FilterPanel.jsx
```diff
@@ -70,12 +100,36 @@ export default function FilterPanel({ onClose }) {
             expandIcon={<ExpandMoreIcon />}
             aria-controls={name + '-content'}
             id={name + '-header'}>
-            <Typography>{name}</Typography>
+            <Grid container alignItems="center" justifyContent="space-between">
```
edwingl10, line 103: Tried using `Stack` and it was not styling it how I wanted it despite using the same properties

src/components/FilterPanel.jsx
```diff
@@ -18,38 +20,66 @@ import CloseIcon from '@mui/icons-material/Close';
 import PropTypes from 'prop-types';
 
 export default function FilterPanel({ onClose }) {
-  const [{ minAmount, maxAmount, grades, majors }, setQueryParam] =
+  const [{ minAmount, maxAmount, grades, majors }, setQueryParams] =
     useQueryParams();
+  const [filterVals, setFilterVals] = useState({
```
josueetcom, line 25: How about using useState for each one of these and calling what useQueryParams returns just `params` instead of the individual filter names?

src/lib/useQueryParams.ts
```diff
@@ -32,15 +28,17 @@ export default function useQueryParams(
...
+    Object.keys(params).forEach((k) => {
+      if (params[k] === undefined) {
+        delete params[k];
+      }
+    });
+
     navigate(
       {
-        search: queryString.stringify(origParams, options),
+        search: queryString.stringify(params, options),
```
josueetcom, line 41: This'll end up overwriting everything in the query string. You need to modify origParams instead unless you intend to overwrite the whole query string.

josueetcom (commented):
Initial comments

src/components/FilterPanel.jsx
```diff
@@ -70,12 +100,36 @@ export default function FilterPanel({ onClose }) {
             expandIcon={<ExpandMoreIcon />}
             aria-controls={name + '-content'}
             id={name + '-header'}>
-            <Typography>{name}</Typography>
+            <Grid container alignItems="center" justifyContent="space-between">
+              <Typography>{name}</Typography>
+              <Badge
+                badgeContent={filter.badge}
```
josueetcom, line 106: I'm noticing the badge doesn't appear or disappear until filters are applied. There's also no indicator that filters have or haven't been applied. Some changes to look into:
- Have Badge (or Chip) indicate the number of filters applied/not yet applied. E.g. It could say `3 not yet applied` vs. `2 applied` or simply `2`. 
- The badge could also be a different color if the filter value is different from the last applied value.
- The `Apply Filters` button could be enabled only when there are changes to be applied.
- We could have some text above/below the `Apply Filters` button that indicates filters haven't been applied yet, e.g. 'Your filters have not been applied yet'.
- On mobile before we allow the filter panel to close we should provide a warning confirmation dialog about leaving without applying changes (in which case the local state of the filters here should reset).

More reading: https://baymard.com/blog/how-to-design-applied-filters

src/components/FilterPanel.jsx
```diff
@@ -70,12 +100,36 @@ export default function FilterPanel({ onClose }) {
...
+                color="primary"
+                sx={{ mr: 2 }}
+              />
+            </Grid>
           </AccordionSummary>
 
-          <AccordionDetails sx={{ m: 1 }}>{filter}</AccordionDetails>
+          <AccordionDetails sx={{ m: 1 }}>{filter.comp}</AccordionDetails>
         </Accordion>
```
josueetcom, line 114: What do you think of surrounding `<Accordion>`'s children with a `<Container maxWidth='sm`>`? That way on screens just barely smaller than `md` won't have such a large distance between the filter name and the number of filters applied. It would also help shorten the width of the Majors autocomplete a bit which you were concerned about last time:
Screen Shot 2022-02-05 at 4 07 54 PM

src/components/FilterPanel.jsx
```diff
@@ -70,12 +100,36 @@ export default function FilterPanel({ onClose }) {
...
+            setQueryParams(filterVals);
+            onClose();
+          }}>
+          Apply Filters
+        </Button>
+        <Button
+          sx={{ display: { xs: 'none', md: 'block' } }}
+          onClick={resetFilters}>
+          Clear Filters
+        </Button>
```
josueetcom, line 131: For consistency between mobile and desktop experiences I'd prefer to either keep this around for mobile for consistency OR remove this button entirely and make this bar appear on desktop:
Screen Shot 2022-02-05 at 4 09 20 PM

I'm not fond of the idea of having two different reset buttons if they do the same thing (unless one is like "clear + apply" but the other is just "clear")

josueetcom (changes requested):

src/components/FilterPanel.jsx
edwingl10: Doesn't it look a bit off? What if we use a custom width that is shorter than `xs`
Screen Shot 2022-02-05 at 9 04 27 PM

src/components/FilterPanel.jsx
edwingl10: so I would change `setQueryParam` to how it originally was and then call it on each seperate fitler?

src/components/FilterPanel.jsx
josueetcom: You need to surround both the summary and the detail component in the Container (or maybe have containers within them)

src/components/FilterPanel.jsx
edwingl10: went with the first option

src/components/FilterPanel.jsx
edwingl10: Ok that looks a bit better. I had to surround the Summary and the Details in different containers. 

Screen Shot 2022-02-06 at 11 22 21 AM

src/components/FilterPanel.jsx
josueetcom: You can call `setQueryParams({minAmount, majors, etc.})`

src/components/FilterPanel.jsx
josueetcom: Strange it had to be separate containers 🤔

src/lib/useQueryParams.ts
```diff
@@ -10,18 +10,14 @@ const options: ParseOptions = {
...
-  replace: boolean
-) => void;
+type SetQueryParamFn = (params: Record<string, any>, replace: boolean) => void;
 
 /**
- * Returns the query string parsed as an object, and a function to update individual parameters.
+ * Returns the query string parsed as an object, and a function to update parameters.
  * @example
- * const [params, setQueryParam] = useQueryParam();
+ * const [params, setQueryParams] = useQueryParam);
```
josueetcom, line 18:
```suggestion
 * const [params, setQueryParams] = useQueryParams();
```

src/components/FilterPanel.jsx
josueetcom: Add `disableGutters` to the containers so it doesn't have left and right padding

edwingl10: yeah if I put them in the same container, the Accodrion would stay open and the drop down wouldn't work.

src/components/FilterPanel.jsx
```diff
@@ -9,77 +9,221 @@ import {
...
-
-          <AccordionDetails sx={{ m: 1 }}>{filter}</AccordionDetails>
-        </Accordion>
+        <AccordionFilter name={name} count={filter.count}>
+          {filter.comp}
+        </AccordionFilter>
       ))}
+
+      {filtersChanged && (
+        <Alert severity="warning">Your changes have not yet been applied</Alert>
```
edwingl10, line 205: Didn't remove border since it looks weird when the filter panel spans the entire screen width. 

Screen Shot 2022-02-09 at 4 46 04 PM
VS
Screen Shot 2022-02-09 at 4 47 02 PM

src/components/FilterPanel.jsx
```diff
@@ -9,77 +9,284 @@ import {
...
 import CloseIcon from '@mui/icons-material/Close';
 import PropTypes from 'prop-types';
+import GradeLevel from '../types/GradeLevel';
 
-export default function FilterPanel({ onClose }) {
-  const [{ minAmount, maxAmount, grades, majors }, setQueryParam] =
-    useQueryParams();
+function FilterChip({ label, onDelete, variant }) {
+  return (
+    <Chip
```
josueetcom, line 31: Some suggestions:
- Add a checkmark icon on the left for active filters that will still be applied after any pending changes
- For active filters that will not be applied after pending changes:
  - Gray it out/make it unclickable/undeletable
  - No checkmark icon
  
 Some properties to look into: `clickable`, `disabled`, `color`, `variant`, `onDelete`. You might change some or all of those

PR #912, opened 2022-02-11
apply/cancel button and alert text

### LINKED ISSUE #911: Apply / Cancel button
Instead of changing the query params everytime there is a change in the Filter inputs, there should be an Apply all button to apply those changes. 
Similiary, a Cancel button should cancel all pending changes. We should also show a warning letting the user know that their pending changes have not yet been applied.

edwingl10:
## Motivation and Context

Fixes #911

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. Go to scholarships page
3. choose filters and click apply to make those changes
4. cancel changes 
5. notice the text above the buttons that lets you know changes have not been applied yet

### DIFF
--- a/src/components/AmountFilter.jsx
+++ b/src/components/AmountFilter.jsx
@@ -35,7 +35,7 @@ export default function AmountFilter(props) {
       </Grid>
 
       <Slider
-        defaultValue={[0, 50000]}
+        value={[min || 0, max || 50000]}
         min={0}
         max={50000}
         step={100}
--- a/src/components/AmountFilter.jsx
+++ b/src/components/AmountFilter.jsx
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import { InputLabel, Grid, Slider } from '@mui/material';
 import { Remove as RemoveIcon } from '@mui/icons-material';
 import AmountTextField from './AmountTextField';
-import experiments from '../lib/experiments';
 
 export default function AmountFilter(props) {
   const { min, max, onMinChange, onMaxChange } = props;
@@ -35,19 +34,17 @@ export default function AmountFilter(props) {
         />
       </Grid>
 
-      {experiments.expShowSlider && (
-        <Slider
-          defaultValue={[0, 10000]}
-          min={0}
-          max={50000}
-          step={100}
-          onChange={(e) => {
-            onMinChange(e.target.value[0]);
-            onMaxChange(e.target.value[1]);
-          }}
-          sx={{ mt: 1 }}
-        />
-      )}
+      <Slider
+        defaultValue={[0, 50000]}
+        min={0}
+        max={50000}
+        step={100}
+        onChange={(e) => {
+          onMinChange(e.target.value[0]);
+          onMaxChange(e.target.value[1]);
+        }}
+        sx={{ mt: 1 }}
+      />
     </Grid>
   );
 }
--- a/src/components/FilterPanel.jsx
+++ b/src/components/FilterPanel.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { ExpandMore as ExpandMoreIcon } from '@mui/icons-material';
 import {
   Accordion,
@@ -15,40 +15,60 @@ import AmountFilter from './AmountFilter';
 import GradeLevelFilter from './GradeLevelFilter';
 import MajorFilter from './MajorFilter';
 import CloseIcon from '@mui/icons-material/Close';
+import WarningAmberOutlinedIcon from '@mui/icons-material/WarningAmberOutlined';
 import PropTypes from 'prop-types';
 
 export default function FilterPanel({ onClose }) {
-  const [{ minAmount, maxAmount, grades, majors }, setQueryParam] =
-    useQueryParams();
+  const [params, setQueryParams] = useQueryParams();
+
+  const [minAmount, setMinAmount] = useState(params.minAmount);
+  const [maxAmount, setMaxAmount] = useState(params.maxAmount);
+  const [grades, setGrades] = useState(params.grades);
+  const [majors, setMajors] = useState(params.majors);
 
   const filters = {
-    Major: (
-      <MajorFilter
-        majors={majors}
-        onSelect={(m) => setQueryParam('majors', m)}
-        onDelete={(m) =>
-          setQueryParam(
-            'majors',
-            majors.filter((major) => major !== m)
-          )
-        }
-      />
-    ),
-    Amount: (
-      <AmountFilter
-        min={minAmount ?? 0}
-        max={maxAmount ?? 0}
-        onMinChange={(val) => setQueryParam('minAmount', val)}
-        onMaxChange={(val) => setQueryParam('maxAmount', val)}
-      />
-    ),
-    'Grade Level': (
-      <GradeLevelFilter
-        grades={new Set(grades)}
-        changeFn={(e) => setQueryParam('grades', e)}
-      />
-    ),
+    Major: {
+      comp: (
+        <MajorFilter
+          majors={majors}
+          onSelect={(m) => setMajors(m)}
+          onDelete={(m) => setMajors(majors.filter((major) => major !== m))}
+        />
+      ),
+      wasChanged:
+        JSON.stringify(majors?.length > 0 ? majors : undefined) !==
+        JSON.stringify(params.majors),
+    },
+    Amount: {
+      comp: (
+        <AmountFilter
+          min={minAmount ?? 0}
+          max={maxAmount ?? 0}
+          onMinChange={(val) => setMinAmount(parseInt(val))}
+          onMaxChange={(val) => setMaxAmount(parseInt(val))}
+        />
+      ),
+      wasChanged:
+        (minAmount || undefined) !== params.minAmount ||
+        (maxAmount || undefined) !== params.maxAmount,
+    },
+    'Grade Level': {
+      comp: (
+        <GradeLevelFilter
+          grades={new Set(grades)}
+          changeFn={(e) => setGrades(e)}
+        />
+      ),
+      wasChanged:
+        JSON.stringify(grades?.length > 0 ? grades : undefined) !==
+        JSON.stringify(params.grades),
+    },
   };
+
+  const filtersChanged =
+    filters.Major.wasChanged ||
+    filters.Amount.wasChanged ||
+    filters['Grade Level'].wasChanged;
 
   return (
     <Box>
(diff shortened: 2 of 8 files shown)


src/components/FilterPanel.jsx
```diff
@@ -15,41 +15,61 @@ import AmountFilter from './AmountFilter';
...
+    Major: {
+      comp: (
+        <MajorFilter
+          majors={majors}
+          onSelect={(m) => setMajors(m)}
+          onDelete={(m) => setMajors(majors.filter((major) => major !== m))}
+        />
+      ),
+      wasChanged:
+        JSON.stringify(majors?.length > 0 ? majors : undefined) !==
```
josueetcom, line 39: Doesn't this work?
```suggestion
        JSON.stringify(majors) !==
```

src/components/FilterPanel.jsx
```diff
@@ -15,41 +15,61 @@ import AmountFilter from './AmountFilter';
...
+        <AmountFilter
+          min={minAmount ?? 0}
+          max={maxAmount ?? 0}
+          onMinChange={(val) => setMinAmount(parseInt(val))}
+          onMaxChange={(val) => setMaxAmount(parseInt(val))}
+        />
+      ),
+      wasChanged:
+        (minAmount || undefined) !== params.minAmount ||
+        (maxAmount || undefined) !== params.maxAmount,
```
josueetcom, line 53: Why is `|| undefined` needed? Isn't the initial value already `undefined` if it's not defined? Cant this just be:
```suggestion
        minAmount !== params.minAmount || maxAmount !== params.maxAmount,
```

src/components/FilterPanel.jsx
```diff
@@ -15,41 +15,61 @@ import AmountFilter from './AmountFilter';
...
+    },
+    'Grade Level': {
+      comp: (
+        <GradeLevelFilter
+          grades={new Set(grades)}
+          changeFn={(e) => setGrades(e)}
+        />
+      ),
+      wasChanged:
+        JSON.stringify(grades?.length > 0 ? grades : undefined) !==
```
josueetcom, line 63: Again I'm not understanding why this doesn't work:
```suggestion
        JSON.stringify(grades) !==
```

src/components/FilterPanel.jsx
```diff
@@ -15,41 +15,61 @@ import AmountFilter from './AmountFilter';
...
+      ),
+      wasChanged:
+        (minAmount || undefined) !== params.minAmount ||
+        (maxAmount || undefined) !== params.maxAmount,
+    },
+    'Grade Level': {
+      comp: (
+        <GradeLevelFilter
+          grades={new Set(grades)}
+          changeFn={(e) => setGrades(e)}
```
josueetcom, line 59:
```suggestion
          changeFn={setGrades}
```

src/components/FilterPanel.jsx
```diff
@@ -15,41 +15,61 @@ import AmountFilter from './AmountFilter';
...
-      <GradeLevelFilter
-        grades={new Set(grades)}
-        changeFn={(e) => setQueryParam('grades', e)}
-      />
-    ),
+    Major: {
+      comp: (
+        <MajorFilter
+          majors={majors}
+          onSelect={(m) => setMajors(m)}
```
josueetcom, line 34:
```suggestion
          onSelect={setMajors}
```

josueetcom: I'm confused by how this works. If a major is selected should it be **added** to the `majors` list?  I.e. something like
```suggestion
          onSelect={(m) => {if (!majors.includes(m)) setMajors([...majors, m]);}}
```

Otherwise if the `MajorFilter` component takes care of updating this list why is this called `onSelect` and not something like `onChange`? And why do we have a separate `onDelete` function if `MajorFilter` can just call that change function? The behavior and naming of these functions is confusing and inconsistent.

src/components/FilterPanel.jsx
josueetcom: Nit: Tangential to above comment, `onChange` would be a better name.

src/components/FilterPanel.jsx
```diff
@@ -15,41 +15,61 @@ import AmountFilter from './AmountFilter';
...
+          onDelete={(m) => setMajors(majors.filter((major) => major !== m))}
+        />
+      ),
+      wasChanged:
```
josueetcom, line 38: Nit: just call this `changed`.

src/components/FilterPanel.jsx
```diff
@@ -15,41 +15,61 @@ import AmountFilter from './AmountFilter';
...
+        JSON.stringify(grades?.length > 0 ? grades : undefined) !==
+        JSON.stringify(params.grades),
+    },
   };
 
+  const filtersChanged =
+    filters.Major.wasChanged ||
+    filters.Amount.wasChanged ||
+    filters['Grade Level'].wasChanged;
```
josueetcom, line 71: Simplify this using [Object.keys()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys) and [Array.map()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map):
```suggestion
  const filtersChanged = Object.keys(filters).some(k => filters[k].changed);
```

This also future proofs against future filters.

src/components/FilterPanel.jsx
```diff
@@ -73,9 +93,43 @@ export default function FilterPanel({ onClose }) {
...
+      {filtersChanged && (
+        <Box
+          sx={{
+            display: 'flex',
+            alignItems: 'center',
+            justifyContent: 'center',
+            mt: 2,
+          }}>
+          <WarningAmberOutlinedIcon color="warning" />
+          <Typography>Your changes have not yet been applied</Typography>
```
josueetcom, line 109: Nit: are you able to add a space here?
```suggestion
          <Typography> Your changes have not yet been applied</Typography>
```

src/components/FilterPanel.jsx
```diff
@@ -73,9 +93,43 @@ export default function FilterPanel({ onClose }) {
             <Typography>{name}</Typography>
           </AccordionSummary>
 
-          <AccordionDetails sx={{ m: 1 }}>{filter}</AccordionDetails>
+          <AccordionDetails sx={{ m: 1 }}>{filter.comp}</AccordionDetails>
         </Accordion>
       ))}
+
+      {filtersChanged && (
```
josueetcom, line 100: I'm noticing the buttons shift when this appears/disappears. How about we keep this element but just change the text and/or icon when filters are currently applied? You could use the **InfoOutlined** or **CheckCircleOutLine** icon to match with the warning one, show  and display something like "Your filters are currently applied".

src/components/FilterPanel.jsx
josueetcom: Also a period is missing at the end here.

src/components/FilterPanel.jsx
```diff
@@ -73,9 +93,43 @@ export default function FilterPanel({ onClose }) {
...
+            alignItems: 'center',
+            justifyContent: 'center',
+            mt: 2,
+          }}>
+          <WarningAmberOutlinedIcon color="warning" />
+          <Typography>Your changes have not yet been applied</Typography>
+        </Box>
+      )}
+
+      <Toolbar sx={{ justifyContent: 'space-evenly', mt: 2 }}>
```
josueetcom, line 113: Why was a Toolbar chosen?

josueetcom: What do you think of using a **Stack** instead and not centering the buttons and text but having them on the left? We could also add a margin of 2*spacing between the buttons so they're not so close to each other.

src/components/FilterPanel.jsx
```diff
@@ -73,9 +93,43 @@ export default function FilterPanel({ onClose }) {
...
+        <Button
+          variant="contained"
+          disabled={!filtersChanged}
+          onClick={() => {
+            setQueryParams({ minAmount, maxAmount, grades, majors });
+            onClose();
+          }}>
+          Apply
+        </Button>
+        <Button
```
josueetcom, line 123: Hmm I know we talked about keeping this always enabled for mobile but I'm noticing that on desktop it doesn't make sense as it doesn't do anything if there are no pending changes. Maybe disable this one too on `!filtersChanged`?

src/lib/useQueryParams.ts
```diff
@@ -10,18 +10,14 @@ const options: ParseOptions = {
   parseNumbers: true,
 };
 
-type SetQueryParamFn = (
-  key: 'minAmount' | 'maxAmount' | 'grades' | 'majors' | 'sortBy',
-  value: any,
-  replace: boolean
-) => void;
+type SetQueryParamFn = (params: Record<string, any>, replace: boolean) => void;
```
josueetcom, line 13:
```suggestion
type SetQueryParamsFn = (params: Record<string, any>, replace: boolean) => void;
```

josueetcom (changes requested):

src/components/AmountFilter.jsx
```diff
@@ -35,19 +34,17 @@ export default function AmountFilter(props) {
...
-          step={100}
-          onChange={(e) => {
-            onMinChange(e.target.value[0]);
-            onMaxChange(e.target.value[1]);
-          }}
-          sx={{ mt: 1 }}
-        />
-      )}
+      <Slider
+        defaultValue={[0, 50000]}
```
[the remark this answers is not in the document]
edwingl10: There is a few problems with this. In order for it to pick up the min/max values, it needs to be controlled but  `defaultValue` stops working if I set a `value` property. 

I could do something like `value={[min, max || RANGE_MAX]}` but im wondering if this could cause some unexpected behavior if the user doesn't set a max

src/components/FilterPanel.jsx
edwingl10: when the user adds a major and then removes it, it becomes `[]` while `params.majors` is `undefined`

src/components/FilterPanel.jsx
edwingl10: similar to the `majors`, once an amount is inputted and deleted, it turns into `NaN` while the params amounts stay as `undefined`

src/components/FilterPanel.jsx
edwingl10: nice!

src/components/FilterPanel.jsx
edwingl10: No need to check if majors includes the major. Not sure how, but it already does that on its own.

src/components/AmountFilter.jsx
```diff
@@ -35,19 +34,17 @@ export default function AmountFilter(props) {
...
-          step={100}
-          onChange={(e) => {
-            onMinChange(e.target.value[0]);
-            onMaxChange(e.target.value[1]);
-          }}
-          sx={{ mt: 1 }}
-        />
-      )}
+      <Slider
+        value={[min || 0, max || 50000]}
```
edwingl10, line 38: did this in case either `min` or `max` are removed and it becomes `NaN`

src/components/FilterPanel.jsx
edwingl10: doesn't seem to be adding a space, but I can include a margin of 8px

src/components/FilterPanel.jsx
edwingl10: Made the change but don't you think it looks off?

src/components/FilterPanel.jsx
```diff
@@ -73,9 +81,44 @@ export default function FilterPanel({ onClose }) {
...
+
+      <Stack direction="row" spacing={1} sx={{ m: 2 }}>
+        {filtersChanged ? (
+          <WarningAmberOutlinedIcon color="warning" />
+        ) : (
+          <CheckCircleOutlineIcon color="success" />
+        )}
+        <Typography>
+          {filtersChanged
+            ? "Your changes haven't yet been applied."
```
edwingl10, line 96: shortend `have not` to `haven't` to avoid the text to stack when on desktop

josueetcom: nit: this should use single quotes.

josueetcom: Makes sense, you could also shorten it to `Your changes are not yet applied`, which seems to be closer in length to the other message (look how much screen width each of these take up on your screen):

Your changes haven't yet been applied.
Your filters are currently applied.
Your changes are not yet applied.

src/components/FilterPanel.jsx
josueetcom: Looks good to me?

josueetcom (approved):
Just minor comments but otherwise looks pretty good 👍

PR #933, opened 2022-02-21
Autocomplete form fields

### LINKED ISSUE #932: Autofill requirement values in scholarhip form
We should implement a feature that allows users to auto fill values found in the description. 

This will also help us update scholarships with lint issues.

edwingl10:
## Motivation and Context

Fixes #932 

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. Add new scholarship & put some requirements in the description field
3. should see an alert when in the requirements section 
4. click autofill button
5. Edit a scholarship with lint issues & click autofill button

### DIFF
--- a/src/components/ScholarshipForm.jsx
+++ b/src/components/ScholarshipForm.jsx
@@ -28,7 +28,7 @@ import { SCHOOLS, STATES, MAJORS } from '../types/options';
 import GradeLevel from '../types/GradeLevel';
 import Ethnicity from '../types/Ethnicity';
 import ScholarshipsContext from '../models/ScholarshipsContext';
-import { lint } from '../lib/lint';
+import { lintReqs } from '../lib/lint';
 
 const labelStyle = { marginBottom: 2 };
 
@@ -62,6 +62,9 @@ function ScholarshipForm({ scholarship }) {
         .finally(() => setSubmitting(false));
     },
   });
+
+  let lintIssues = {};
+  if (activeStep === 1) lintIssues = lintReqs(formik.values);
 
   // Initially requirements is null but is set to {} when the "no requirements"
   // checkbox is explicitly set.
@@ -128,6 +131,38 @@ function ScholarshipForm({ scholarship }) {
         'Include information that is required for applicants to have.',
       content: (
         <Grid container spacing={3}>
+          {openAlert && lintIssues?.messages?.length > 0 && (
+            <Alert
+              severity="warning"
+              sx={{ mt: 2 }}
+              onClose={() => {
+                setOpenAlert(false);
+              }}>
+              <AlertTitle>
+                <strong>
+                  We found the following potential requirements in the
+                  description. Would you like to populate these values?
+                </strong>
+              </AlertTitle>
+              <Box component="ul">
+                {lintIssues.messages?.map((m, i) => (
+                  <Typography key={i} component="li">
+                    {m}
+                  </Typography>
+                ))}
+              </Box>
+              <Button
+                onClick={() => {
+                  const vals = formik.values.requirements;
+                  formik.setFieldValue('requirements', {
+                    ...vals,
+                    ...lintIssues.reqs,
+                  });
+                }}>
+                Autofill
+              </Button>
+            </Alert>
+          )}
           <Grid item xs={12}>
             <FormControlLabel
               control={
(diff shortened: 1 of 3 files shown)


src/components/ScholarshipForm.jsx
```diff
@@ -239,59 +242,97 @@ function ScholarshipForm({ scholarship }) {
   }
 
   const onLastStep = activeStep == Object.keys(stepperItems).length - 1;
+  const lintIssues =
+    activeStep === 1 ? lint(formik.values) : { issues: {}, lintVals: {} };
```
edwingl10, line 246: had to do this to avoid an error when using Object.keys() below

src/components/ScholarshipForm.jsx
```diff
@@ -239,59 +242,97 @@ function ScholarshipForm({ scholarship }) {
...
-                    {submissionError.toString()}
-                  </Alert>
-                )}
-              </StepContent>
-            </Step>
-          )
-        )}
-      </Stepper>
-    </form>
+      {openAlert && Object.keys(lintIssues?.issues).length > 0 && (
```
edwingl10, line 302: I was thinking of making a separate function that returns `{messages, values}` specific to requirements, that way this alert wont show when there are lint errors not related to the requirements

src/components/ScholarshipForm.jsx
josueetcom: You can avoid this by embedding the alert within Requirements step. Between these two lines would be a good place I think:
Screen Shot 2022-02-22 at 12 03 50 AM

src/components/ScholarshipForm.jsx
josueetcom: Hmm, yeah there are some lint issues that aren't automatically addressable. What if you return something that looks like`{messages: string[], fixes: {requirements: {gpa: number, ...}}}`?

Or you could make `fixes` an array of `{key: string, values: any[]}` objects.

src/components/ScholarshipForm.jsx
```diff
@@ -239,59 +242,97 @@ function ScholarshipForm({ scholarship }) {
...
+          </AlertTitle>
+          <Box component="ul">
+            {lintIssues.issues.map((m, i) => (
+              <Typography key={i} component="li">
+                {m}
+              </Typography>
+            ))}
+          </Box>
+          <Button
+            variant="contained"
```
josueetcom, line 323: I suggest making this `outlined`

src/components/ScholarshipForm.jsx
```diff
@@ -60,6 +63,9 @@ function ScholarshipForm({ scholarship }) {
     },
   });
 
+  let lintIssues = {};
+  if (activeStep === 1) lintIssues = lintReqs(formik.values);
```
edwingl10, line 67: not necessary but It's probably better than calling this everytime there is a change in the description.

src/components/ScholarshipForm.jsx
```diff
@@ -125,6 +131,40 @@ function ScholarshipForm({ scholarship }) {
...
+                    <Typography key={i} component="li">
+                      {m}
+                    </Typography>
+                  ))}
+                </Box>
+                <Button
+                  onClick={() => {
+                    const vals = formik.values.requirements;
+                    formik.setFieldValue('requirements', {
+                      ...vals,
```
edwingl10, line 159: tried doing `...[formik.values.requirements]` but it would override all other values that weren't being autofilled

src/components/ScholarshipForm.jsx
```diff
@@ -128,6 +131,40 @@ function ScholarshipForm({ scholarship }) {
...
+                      ...lintIssues.reqs,
+                    });
```
josueetcom, line 161: Probably want to do something like formik.setFieldValue('requirements.X', [...reqs.X, lintReqs.X]) for each requirement. Or change this to do that. E.g.:
```suggestion
                    const vals = formik.values.requirements;
                    formik.setFieldValue('requirements', {
                      ...vals,
                      requirements = {
                        grades: [...reqs.grades, ...lintReqs.grades], 
                        ...
                      },
                    });
```

src/components/ScholarshipForm.jsx
```diff
@@ -63,8 +61,30 @@ function ScholarshipForm({ scholarship }) {
     },
   });
 
-  let lintIssues = {};
-  if (activeStep === 1) lintIssues = lintReqs(formik.values);
+  const lintIssues = activeStep === 1 ? lintReqs(formik.values) : {};
+  function autoFill() {
```
josueetcom, line 65:
```suggestion
  const autoFill = () => {
```

src/components/ScholarshipForm.jsx
```diff
@@ -63,8 +61,30 @@ function ScholarshipForm({ scholarship }) {
...
+    const vals = formik.values.requirements;
+    const lintVals = lintIssues.reqs;
+    const updatedReqs = {};
+
+    const grades = [...(vals.grades || []), ...(lintVals.grades || [])];
+    const schools = [...(vals.schools || []), ...(lintVals.schools || [])];
+    const states = [...(vals.states || []), ...(lintVals.states || [])];
```
josueetcom, line 72: Could maybe skip the length check below and do something like:
```suggestion
    const updatedReqs = {
      grades: lintVals.grades ? [...(vals.grades || []), ...(lintVals.grades || [])] : val.grades,
      schools: lintVals.schools ? [...(vals.schools || []), ...(lintVals.schools || [])] : val.schools,
      ...
```

I'm indifferent about this though as I'm not sure how `undefined` values would be treated in the else case.

josueetcom (approved):

src/components/ScholarshipForm.jsx
edwingl10: I'm going to keep it as is.

josueetcom (approved):

PR #953, opened 2022-03-03
add tests for HeaderNavMenu component

Jobhdez:
# Testing
I ran `yarn test` and the tests passed!

### DIFF
--- a/src/components/HeaderNavMenu.test.jsx
+++ b/src/components/HeaderNavMenu.test.jsx
@@ -23,7 +23,10 @@ describe('HeaderNavMenu', () => {
         <HeaderNavMenu links={links} />
       </Router>
     );
-    expect(document.querySelector('a').getAttribute('href')).toBe('/topology');
+    expect(screen.getByText('topology').closest('a')).toHaveAttribute(
+      'href',
+      '/topology'
+    );
     expect(screen.getByText('abstract-algebra').closest('a')).toHaveAttribute(
       'href',
       '/abstract-algebra'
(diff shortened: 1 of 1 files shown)


src/components/HeaderNavMenu.test.jsx
```diff
@@ -0,0 +1,39 @@
...
+};
+
+describe('HeaderNavMenu', () => {
+  test('renders App component', () => {
+    render(
+      <Router>
+        <HeaderNavMenu links={links} />
+      </Router>
+    );
+  });
```
edwingl10, line 19: This is not testing anything. 
Also, since `HeaderNavMenu` returns `Tabs` with links, we can ensure this component renders with the test below.

src/components/HeaderNavMenu.test.jsx
```diff
@@ -0,0 +1,39 @@
...
+  test('renders Header links', () => {
+    render(
+      <Router>
+        <HeaderNavMenu links={links} />
+      </Router>
+    );
+    expect(screen.getByText('topology').closest('a')).toHaveAttribute(
+      'href',
+      '/topology'
+    );
```
edwingl10, line 29: You can use `getByRole` and do something like

`expect(
      screen.getByRole('tab', { name: 'topology' }).getAttribute('href')
    ).toBe('/topology');`

edwingl10 (changes requested):

src/components/HeaderNavMenu.test.jsx
Jobhdez: what's wrong with what I had?

edwingl10 (approved):

PR #958, opened 2022-03-11
Full state name 

edwingl10:
## Motivation and Context

Displays full state instead of just the abbreviation. 

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. add a scholarship
3. add a state requirement


src/components/ScholarshipCard.tsx
```diff
@@ -174,7 +175,7 @@ export default function ScholarshipCard({
               </Typography>
               <DetailCardCell
                 label="State"
-                text={reqs?.states?.join(', ') || 'All'}
+                text={reqs?.states?.map(State.toString).join(', ') || 'All'}
```
josueetcom, line 178: Can you sort these while you're at it?

src/types/States.ts
```diff
@@ -0,0 +1,81 @@
...
+const toStringMappings: Record<string, string> = States.reduce(
+  (arr, s) => ({
+    ...arr,
+    [s.abbr]: `${s.name} (${s.abbr})`,
+  }),
+  {}
+);
+
+export namespace State {
+  export interface StateType {
```
josueetcom, line 72: Is this not possible?
```ts
interface State {}
namespace State {}
export default State;
```

edwingl10: yes! I dont know what I was missing but it was acting weird previously.

josueetcom: Maybe you were exporting State twice.

src/types/States.ts
```diff
@@ -0,0 +1,83 @@
...
+Wisconsin	WI
+Wyoming	WY`;
+
+interface State {
+  /** Full name of the state or territory, e.g. `'Alabama'`. */
+  name: string;
+  /** State abbreviation, e.g. `'AL'`. */
+  abbr: string;
+}
+export const States: State[] = RAW_STATES.split('\n')
```
josueetcom, line 65: Add a separating line and upper case the exported constant:
```suggestion

export const STATES: State[] = RAW_STATES.split('\n')
```

src/types/States.ts
edwingl10: yeah that was the problem 😂

josueetcom (approved):

src/types/States.ts
```diff
@@ -0,0 +1,84 @@
...
+  {}
+);
+
+namespace State {
+  export function toString(state: string): string {
+    return toStringMappings[state];
+  }
+}
+
+export default State;
```
josueetcom, line 84: Oh I forgot to mention that since the default exported name is State then they should go in the file called State.js

PR #961, opened 2022-03-13
tags support

edwingl10:
## Motivation and Context

Adds tag support. 

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. add new scholarship with tags

### DIFF
--- a/src/components/FormikAutocomplete.jsx
+++ b/src/components/FormikAutocomplete.jsx
@@ -5,15 +5,16 @@ import PropTypes from 'prop-types';
 
 /* eslint-disable react/jsx-props-no-spreading */
 function FormikAutocomplete(props) {
-  const { label, id, labelStyle, formik, placeholder, ...otherProps } = props;
+  const {
+    label,
+    id,
+    labelStyle,
+    formik,
+    placeholder,
+    onChange,
+    ...otherProps
+  } = props;
   const values = getIn(formik.values, id, []);
-
-  const formatTags = (vals) => {
-    const newVals = new Set(
-      vals.map((v) => v.toLowerCase().replace(/\s+/g, '-'))
-    );
-    formik.setFieldValue(id, Array.from(newVals));
-  };
 
   return (
     <>
@@ -23,7 +24,7 @@ function FormikAutocomplete(props) {
         multiple
         value={values}
         onChange={(e, val) =>
-          id === 'tags' ? formatTags(val) : formik.setFieldValue(id, val)
+          onChange ? onChange(val) : formik.setFieldValue(id, val)
         }
         renderInput={(params) => (
           <TextField
@@ -45,10 +46,12 @@ FormikAutocomplete.propTypes = {
   labelStyle: PropTypes.object,
   formik: PropTypes.object.isRequired,
   placeholder: PropTypes.string,
+  onChange: PropTypes.func,
 };
 FormikAutocomplete.defaultProps = {
   label: '',
   labelStyle: {},
   placeholder: '',
+  onchange: undefined,
 };
 export default FormikAutocomplete;
--- a/src/components/FormikAutocomplete.jsx
+++ b/src/components/FormikAutocomplete.jsx
@@ -5,8 +5,7 @@ import PropTypes from 'prop-types';
 
 /* eslint-disable react/jsx-props-no-spreading */
 function FormikAutocomplete(props) {
-  const { label, id, labelStyle, formik, placeholder, options, ...otherProps } =
-    props;
+  const { label, id, labelStyle, formik, placeholder, ...otherProps } = props;
   const values = getIn(formik.values, id, []);
 
   const formatTags = (vals) => {
@@ -23,7 +22,6 @@ function FormikAutocomplete(props) {
         id={id}
         multiple
         value={values}
-        options={options}
         onChange={(e, val) =>
           id === 'tags' ? formatTags(val) : formik.setFieldValue(id, val)
         }
@@ -47,12 +45,10 @@ FormikAutocomplete.propTypes = {
   labelStyle: PropTypes.object,
   formik: PropTypes.object.isRequired,
   placeholder: PropTypes.string,
-  options: PropTypes.array,
 };
 FormikAutocomplete.defaultProps = {
   label: '',
   labelStyle: {},
   placeholder: '',
-  options: [],
 };
 export default FormikAutocomplete;
--- a/src/components/FormikAutocomplete.jsx
+++ b/src/components/FormikAutocomplete.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { getIn } from 'formik';
 import { Autocomplete, InputLabel, TextField } from '@mui/material';
 import PropTypes from 'prop-types';
@@ -16,6 +16,8 @@ function FormikAutocomplete(props) {
   } = props;
   const values = getIn(formik.values, id, []);
 
+  const [inputValue, setInputValue] = useState('');
+
   return (
     <>
       <InputLabel sx={labelStyle}>{label}</InputLabel>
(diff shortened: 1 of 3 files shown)


src/types/options.ts
```diff
@@ -2149,3 +2149,16 @@ export const SCHOOLS: School[] = RAW_SCHOOLS.split('\n')
     state,
     website: website.replace('http://', ''),
   }));
+
+const RAW_TAGS = `athletics
```
josueetcom, line 2153: A few comments:
- Instead of a multiline string just make this a list of strings (should be pretty easy with VS Code multi-cursors). The `RAW` values we have elsewhere are a little more complicated as there are multiple values in one line.
- The casing here is inconsistent `DACA only` but `essay`.
- How do we plan on handling similar tags like `Essay` vs. `essay`?
- I'm curious if for some of these we should just have separate checkboxes in the form that users can check if it's the case. We can store the values as tags under the hood.

edwingl10: what tag options could we provide instead then?

josueetcom: I feel like this is more of a design question we should discuss with others but for now I feel like we should have either:
- a blank autocomplete without any preset options. E.g. 
> **Keywords**
> [E.g. athletics, daca, essay, stem, etc.]
- how it is right now but with additional info to explain each keyword a little better. Maybe you can make this a map from `'tag'` to `'description'`, override `getOptionLabel` to return the key and `renderOption` to return something like `<strong>${option}</strong>: ${KEYWORDS[options]}` or `<strong>${option}</strong>: ${KEYWORDS[options].description}`. E.g. you might see this in the options:
  - **daca-only**: Only DACA recipients can apply.
  - **stem**: For students majoring or interested in majoring in STEM.
  - **yearly**: This scholarship is offered yearly.

We can lowercase everything so `Daca` and `DACA` and `daca` aren't separate words and replace spaces with `-` so it's more consistent. When users submit a string enter they should see the lowercase value of what they entered with spaces replaced.

Eventually we could maybe additionally add some checkboxes that'll insert some of these keywords. But maybe if the keyword description is clear enough people will know how to add the right ones.

edwingl10: went with the first option

src/components/FormikAutocomplete.jsx
```diff
@@ -5,16 +5,28 @@ import PropTypes from 'prop-types';
 
 /* eslint-disable react/jsx-props-no-spreading */
 function FormikAutocomplete(props) {
-  const { label, id, labelStyle, formik, placeholder, ...otherProps } = props;
+  const { label, id, labelStyle, formik, placeholder, options, ...otherProps } =
+    props;
   const values = getIn(formik.values, id, []);
+
+  const formatTags = (vals) => {
+    const newVals = new Set(
```
edwingl10, line 13: had to make a set since `DACA only` and `daca only` causes duplicates

src/components/FormikAutocomplete.jsx
```diff
@@ -5,16 +5,28 @@ import PropTypes from 'prop-types';
...
       <InputLabel sx={labelStyle}>{label}</InputLabel>
       <Autocomplete
         id={id}
         multiple
         value={values}
-        onChange={(e, val) => formik.setFieldValue(id, val)}
+        options={options}
+        onChange={(e, val) =>
+          id === 'tags' ? formatTags(val) : formik.setFieldValue(id, val)
+        }
```
josueetcom, line 29: This makes the component a bit fragile as handling other special use cases like this would only complicate it more. Instead of this how about just accepting an optional `onChange` property? Then this can look like:
```suggestion
        onChange={onChange ?? (e, val) => formik.setFieldValue(id, val)}
```

src/components/FormikAutocomplete.jsx
```diff
@@ -5,16 +5,28 @@ import PropTypes from 'prop-types';
...
+
   return (
     <>
       <InputLabel sx={labelStyle}>{label}</InputLabel>
       <Autocomplete
         id={id}
         multiple
         value={values}
-        onChange={(e, val) => formik.setFieldValue(id, val)}
+        options={options}
```
josueetcom, line 26: This doesn't seem to be set by anything. What's the difference between this and `values`?

josueetcom (changes requested):

src/components/FormikAutocomplete.jsx
edwingl10: `values` are the tags inputted into the text area. 
Since it's an `autocomplete`, it requires `options`. 

I could remove this and pass an empty array from the `scholarshipForm`.

src/components/FormikAutocomplete.jsx
edwingl10: did it slightly different

src/components/FormikAutocomplete.jsx
```diff
@@ -1,20 +1,47 @@
...
+
   return (
     <>
       <InputLabel sx={labelStyle}>{label}</InputLabel>
       <Autocomplete
         id={id}
         multiple
         value={values}
-        onChange={(e, val) => formik.setFieldValue(id, val)}
+        inputValue={inputValue}
```
edwingl10, line 28: `inputValue` is the current text that the user is typing. It does not include the `values` already chosen.

src/components/FormikAutocomplete.jsx
```diff
@@ -1,20 +1,47 @@
...
+        onChange={(e, val) =>
+          onChange ? onChange(val) : formik.setFieldValue(id, val)
```
edwingl10, line 30: still needed in order to remove chose values

src/components/FormikAutocomplete.jsx
```diff
@@ -23,10 +25,23 @@ function FormikAutocomplete(props) {
...
         }
+        onInputChange={(event, newInputValue) => {
+          const options = newInputValue.split(',');
+
+          if (options.length > 1) {
+            const vals = values.concat(
+              options.map((x) => x.trim()).filter((x) => x)
+            );
+            setInputValue('');
```
josueetcom, line 39: nit: keep the last value as the input value? Otherwise I can imagine copying and pasting something like "Foo, Bar, and Baz" and `and Baz` become its own value.

src/components/FormikAutocomplete.jsx
```diff
@@ -47,12 +62,12 @@ FormikAutocomplete.propTypes = {
...
   placeholder: PropTypes.string,
-  options: PropTypes.array,
+  onChange: PropTypes.func,
 };
 FormikAutocomplete.defaultProps = {
   label: '',
   labelStyle: {},
   placeholder: '',
-  options: [],
+  onchange: undefined,
```
josueetcom, line 71:
```suggestion
  onChange: undefined,
```

josueetcom (approved):

src/components/FormikAutocomplete.jsx
edwingl10: im not understanding

josueetcom: Whatever the last value parsed is. Looks like it's a one-line change.
```suggestion
            const vals = values.concat(
              options.map((x) => x.trim()).filter((x) => x)
            );
            setInputValue(vals.pop());
```

src/components/FormikAutocomplete.jsx
```diff
@@ -25,7 +36,23 @@ function FormikAutocomplete(props) {
...
+        }
+        onInputChange={(event, newInputValue) => {
+          const options = newInputValue.split(',');
+
+          if (options.length > 1) {
+            const vals = values.concat(
+              options.map((x) => x.trim()).filter((x) => x)
+            );
+            setInputValue('');
+            return onChange ? onChange(vals) : formik.setFieldValue(id, vals);
```
josueetcom, line 51: Just noticed this shows up twice. Instead of having this show up twice just do it once near the top:
```js
const onChange = props.onChange ?? (e, val) => formik.setFieldValue(id, val);
```

josueetcom (approved):

src/components/FormikAutocomplete.jsx
edwingl10: the problem with this is that it doesn't work anymore if you are typing in the tags. You would have to enter the comma in between text that is already there. I'm going to omit this since I don't think many users will copy/paste tags 🤔

src/components/FormikAutocomplete.jsx
edwingl10: change it a bit since it was throwing an error when deleting chips.

PR #969, opened 2022-03-30
Filter chips above results

### LINKED ISSUE #919: Filter chips above results
We should show the filters as `chips` above the results page.

edwingl10:
## Motivation and Context

Fixes #919 

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. go to scolarships page
3. apply and remove filters

### DIFF
--- a/src/components/FilterPanel.jsx
+++ b/src/components/FilterPanel.jsx
@@ -1,4 +1,5 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
+import { useLocation } from 'react-router-dom';
 import { ExpandMore as ExpandMoreIcon } from '@mui/icons-material';
 import {
   Accordion,
@@ -24,6 +25,7 @@ import CloseIcon from '@mui/icons-material/Close';
 import CheckCircleOutlineIcon from '@mui/icons-material/CheckCircleOutline';
 import WarningAmberOutlinedIcon from '@mui/icons-material/WarningAmberOutlined';
 import PropTypes from 'prop-types';
+import queryString from 'query-string';
 
 export default function FilterPanel({ onClose }) {
   const [dialogOpen, setDialogOpen] = useState(false);
@@ -33,6 +35,20 @@ export default function FilterPanel({ onClose }) {
   const [maxAmount, setMaxAmount] = useState(params.maxAmount);
   const [grades, setGrades] = useState(params.grades);
   const [majors, setMajors] = useState(params.majors);
+  const location = useLocation();
+
+  useEffect(() => {
+    const query = queryString.parse(location.search, {
+      arrayFormat: 'bracket-separator',
+      arrayFormatSeparator: ',',
+      parseNumbers: true,
+    });
+
+    setMinAmount(query.minAmount);
+    setMaxAmount(query.maxAmount);
+    setGrades(query.grades);
+    setMajors(query.majors);
+  }, [location]);
 
   const filters = {
     'What are you studying?': {
--- a/src/components/FilterPanel.jsx
+++ b/src/components/FilterPanel.jsx
@@ -52,13 +52,7 @@ export default function FilterPanel({ onClose }) {
 
   const filters = {
     'What are you studying?': {
-      comp: (
-        <MajorFilter
-          majors={majors}
-          onChange={setMajors}
-          majorParams={params.majors}
-        />
-      ),
+      comp: <MajorFilter majors={majors} onChange={setMajors} />,
       changed:
         JSON.stringify(majors || []) !== JSON.stringify(params.majors || []),
     },
--- a/src/components/FilterPanel.jsx
+++ b/src/components/FilterPanel.jsx
@@ -25,14 +25,13 @@ import CheckCircleOutlineIcon from '@mui/icons-material/CheckCircleOutline';
 import WarningAmberOutlinedIcon from '@mui/icons-material/WarningAmberOutlined';
 import PropTypes from 'prop-types';
 
-export default function FilterPanel({ onClose }) {
+export default function FilterPanel({ onClose, majors, setMajors }) {
   const [dialogOpen, setDialogOpen] = useState(false);
   const [params, setQueryParams] = useQueryParams();
 
   const [minAmount, setMinAmount] = useState(params.minAmount);
   const [maxAmount, setMaxAmount] = useState(params.maxAmount);
   const [grades, setGrades] = useState(params.grades);
-  const [majors, setMajors] = useState(params.majors);
 
   const filters = {
     'What are you studying?': {
@@ -157,4 +156,6 @@ export default function FilterPanel({ onClose }) {
 
 FilterPanel.propTypes = {
   onClose: PropTypes.func.isRequired,
+  majors: PropTypes.arrayOf(PropTypes.string),
+  setMajors: PropTypes.func.isRequired,
 };
--- a/src/components/FilterPanel.jsx
+++ b/src/components/FilterPanel.jsx
@@ -35,8 +35,14 @@ export default function FilterPanel({ onClose }) {
   const [majors, setMajors] = useState(params.majors);
 
   const filters = {
-    Major: {
-      comp: <MajorFilter majors={majors} onChange={setMajors} />,
+    'What are you studying?': {
+      comp: (
+        <MajorFilter
+          majors={majors}
+          onChange={setMajors}
+          majorParams={params.majors}
+        />
+      ),
       changed:
         JSON.stringify(majors || []) !== JSON.stringify(params.majors || []),
     },
@@ -87,7 +93,7 @@ export default function FilterPanel({ onClose }) {
             expandIcon={<ExpandMoreIcon />}
             aria-controls={name + '-content'}
             id={name + '-header'}>
-            <Typography>{name}</Typography>
+            <Typography sx={{ fontWeight: 'medium' }}>{name}</Typography>
           </AccordionSummary>
 
           <AccordionDetails sx={{ m: 1 }}>{filter.comp}</AccordionDetails>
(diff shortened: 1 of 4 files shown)


src/components/FilterPanel.jsx
```diff
@@ -25,18 +25,23 @@ import CheckCircleOutlineIcon from '@mui/icons-material/CheckCircleOutline';
 import WarningAmberOutlinedIcon from '@mui/icons-material/WarningAmberOutlined';
 import PropTypes from 'prop-types';
 
-export default function FilterPanel({ onClose }) {
+export default function FilterPanel({ onClose, majors, setMajors }) {
```
edwingl10, line 28: did this so `majors` could be deleted from `listScholarships`

src/components/MajorFilter.jsx
```diff
@@ -6,7 +6,7 @@ import PropTypes from 'prop-types';
 import ArrowDropUpIcon from '@mui/icons-material/ArrowDropUp';
 import SearchIcon from '@mui/icons-material/Search';
 
-function MajorFilter({ majors, onChange }) {
+function MajorFilter({ majors, onChange, majorParams }) {
```
edwingl10, line 9: used in order to distinguish which filteres are currently applied and which ones are still pending.

src/pages/ListScholarships.jsx
```diff
@@ -9,34 +9,50 @@ import {
...
 
+function FilterChip({ label, deleteFn }) {
+  return (
+    <Chip
+      key={label}
+      label={label}
+      color="primary"
+      onClick={deleteFn}
+      onDelete={deleteFn}
+      sx={{ mr: 2, mt: 3 }}
```
josueetcom, line 32: This seems like something a [Stack](https://mui.com/components/stack/#main-content) might be able to help with. 🤔 

Also, since this is a two property thing for items that otherwise go right next to each other, why not do something like this:
```jsx
const filterChips = {};
if (Number.isInteger(minAmount)) { filterChips[`Min ${minAmount}`]: /* delete function */ ;}
// same thing for maxAmount

majors?.forEach((m) => filterChips[m] = /* delete function */; );
// same thing for grades.

return (
  // ...
  {filterChips.items().map(([label, deleteFn]) => 
    <Chip />
  }
  // ...
);
```

This'll also make it a little easier to add future filters because JSX doesn't need to be touched.

src/pages/ListScholarships.jsx
```diff
@@ -85,6 +101,45 @@ function ListScholarships() {
         <Toolbar />
 
         <Container maxWidth="md" sx={{ flexGrow: 1 }}>
+          <Box
+            sx={{
+              overflowX: 'scroll',
+              whiteSpace: 'nowrap',
+              scrollbarWidth: 'none',
+            }}>
```
josueetcom, line 109: Hmm, it's not entirely obvious there are more chips, especially with the padding on the right: 
Screen Shot 2022-03-30 at 7 22 33 PM

Also the use of screen real estate doesn't look that great on desktop. Kind of makes you wonder why there's so much whitespace on the side.

Couple ideas:
- Add an inset shadow to the edges, kind of like [this one](https://css-tricks.com/books/greatest-css-tricks/scroll-shadows/).
- To maximize screen real estate on large screens what if we place this Box outside of the container and center it? 
  - Alternatively, maybe for `md` screens and up maybe we should show everything no matter how many rows it takes?

(This paragraph is tangential) Something else the use of screen real estate makes me wonder is if we should enlarge the container to `lg` or `xl` and start showing 2 or 3 scholarships per row like we had in our original design 🤔 . It might make better use of space. Along those lines making the filter panel collapsible on bigger screens could be cool too.

josueetcom (changes requested):

src/components/FilterPanel.jsx
```diff
@@ -33,10 +35,30 @@ export default function FilterPanel({ onClose }) {
...
+    const query = queryString.parse(location.search, {
+      arrayFormat: 'bracket-separator',
+      arrayFormatSeparator: ',',
+      parseNumbers: true,
+    });
+
+    setMinAmount(query.minAmount);
+    setMaxAmount(query.maxAmount);
+    setGrades(query.grades);
+    setMajors(query.majors);
```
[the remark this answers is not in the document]
edwingl10: the problem with this is that it would require `params` in the dependancies and that causes a `Warning: Maximum update depth exceeded.`

src/pages/ListScholarships.jsx
```diff
@@ -9,34 +9,50 @@ import {
...
     sortField,
     sortDir,
-    minAmount,
-    maxAmount,
-    grades,
-    majors,
+    minAmount: minAmount,
+    maxAmount: maxAmount,
+    grades: grades,
+    majors: majors,
```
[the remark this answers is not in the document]
edwingl10: oops yes, forgot to revert!

src/components/FilterPanel.jsx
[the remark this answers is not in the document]
josueetcom: Hmm alright, there's probably a way we can use `useMemo` for this but I imagine it probably needs to go in `useQueryParams.ts`. I'll have to take a closer look later.

src/pages/ListScholarships.jsx
```diff
@@ -57,6 +45,30 @@ function ListScholarships() {
...
+  if (Number.isInteger(maxAmount)) {
+    filterChips[`Max $${maxAmount}`] = () =>
+      setQueryParams({ maxAmount: undefined });
+  }
+  majors?.forEach(
+    (m) =>
+      (filterChips[m] = () =>
+        setQueryParams({
+          majors: majors?.filter((major) => major !== m),
+        }))
```
josueetcom, line 62: The surrounding parentheses looks a bit hard to read IMO. What do you think of using curly braces instead?
```suggestion
          {
      filterChips[m] = () =>
        setQueryParams({
          majors: majors?.filter((major) => major !== m),
        })
      }
```

src/pages/ListScholarships.jsx
```diff
@@ -57,6 +45,30 @@ function ListScholarships() {
...
+  );
+  grades?.forEach(
+    (g) =>
+      (filterChips[GradeLevel.toString(g)] = () =>
+        setQueryParams({
+          grades: grades?.filter((grade) => grade !== g),
+        }))
```
josueetcom, line 69: How ugly would it look if we mapped the chip name to the query param arguments to apply?
```suggestion
      (filterChips[GradeLevel.toString(g)] =({
          grades: grades?.filter((grade) => grade !== g),
        }))
```

Below in the Chip, we could do `onClick={() => setQueryParams(queryParams)}`.

Though if we do this, we should probably use better names since `queryParams` and `setQueryParams` sound like they set/use the whole thing. Maybe `updateQueryParams`? Or we could keep the `setQueryParams` name, have it replace everything by default (so "set" = "create or replace" by default) and accept an `options` object parameter that's explicit about the merging behavior. E.g. `setQueryParams(queryParams, {merge: true})`

edwingl10: ^ I tried to implement this but it would cause it to crash. It complained about being an unstable operation.

src/pages/ListScholarships.jsx
edwingl10: I kept it inside the container because once I moved it outside, the `overflow` prop would stop working. But I made is so that the tags wrap on `md` screens and higher

src/pages/ListScholarships.jsx
```diff
@@ -92,6 +118,34 @@ function ListScholarships() {
         <Toolbar />
 
         <Container maxWidth="md" sx={{ flexGrow: 1 }}>
+          <Stack
```
edwingl10, line 121: I kept it inside the container because once I moved it outside, the overflow prop would stop working. But I made is so that the tags wrap on md screens and higher

src/pages/ListScholarships.jsx
```diff
@@ -92,6 +118,34 @@ function ListScholarships() {
...
+              overflowX: 'scroll',
+              scrollbarWidth: 'none',
+              backgroundImage:
+                'linear-gradient(to right, #F8F9FA, #F8F9FA), linear-gradient(to right, #F8F9FA, #F8F9FA), linear-gradient(to right, rgba(0, 0, 0, 0.25), rgba(255, 255, 255, 0)), linear-gradient(to left, rgba(0, 0, 0, 0.25), rgba(255, 255, 255, 0))',
+              backgroundPosition:
+                'left center, right center, left center, right center',
+              backgroundRepeat: 'no-repeat',
+              backgroundSize: '20px 100%, 20px 100%, 10px 100%, 10px 100%',
+              backgroundAttachment: 'local, local, scroll, scroll',
+              '::-webkit-scrollbar': { display: 'none' },
```
edwingl10, line 138: to hide scrollbar on chrome & safari

src/pages/ListScholarships.jsx
```diff
@@ -119,22 +114,34 @@ function ListScholarships() {
...
               overflowX: 'scroll',
-              whiteSpace: 'nowrap',
               scrollbarWidth: 'none',
+              backgroundImage:
+                'linear-gradient(to right, #F8F9FA, #F8F9FA), linear-gradient(to right, #F8F9FA, #F8F9FA), linear-gradient(to right, rgba(0, 0, 0, 0.25), rgba(255, 255, 255, 0)), linear-gradient(to left, rgba(0, 0, 0, 0.25), rgba(255, 255, 255, 0))',
+              backgroundPosition:
+                'left center, right center, left center, right center',
+              backgroundRepeat: 'no-repeat',
+              backgroundSize: '20px 100%, 20px 100%, 10px 100%, 10px 100%',
+              backgroundAttachment: 'local, local, scroll, scroll',
```
josueetcom, line 133: Is this supposed to show something that hints there's more to scroll? I can't see anything on mobile nor desktop.

josueetcom (approved):

src/pages/ListScholarships.jsx
edwingl10: yes, it shows the scroll shadow. It is hard to see since the chips are over it but you can see it in between gaps. 
Screen Shot 2022-04-02 at 8 03 11 PM

edwingl10: I added some padding on the top & bottom so it could be more visible 
Screen Shot 2022-04-02 at 8 17 25 PM

PR #983, opened 2022-04-06
font fix on mobile screen

edwingl10:
## Motivation and Context

Fixes font size on mobile screens  

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. re-size window and make sure text changes size

## Screenshots
Screen Shot 2022-04-06 at 12 38 20 AM

### DIFF
--- a/src/components/HomeSection.jsx
+++ b/src/components/HomeSection.jsx
@@ -12,15 +12,11 @@ const HomeSection = ({
   pic,
 }) => {
   const picSize = main ? '95%' : '85%';
-  const textSize = main
-    ? { xs: 'h4', sm: 'h3', md: 'h2' }
-    : { xs: 'h5', sm: 'h4', md: 'h3' };
-
   return (
     <Container sx={{ padding: 4 }}>
       <Grid container spacing={3} direction={direction} alignItems={alignItems}>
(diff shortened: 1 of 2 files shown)


src/components/HomeSection.jsx
```diff
@@ -12,11 +12,15 @@ const HomeSection = ({
   pic,
 }) => {
   const picSize = main ? '95%' : '85%';
+  const textSize = main
```
josueetcom, line 15: This is doesn't really address the root of the problem, which is the font sizes of different header variants.
Have you looked at this?

https://mui.com/customization/typography/#responsive-font-sizes

`responsiveFontSizes(theme)` seems to be what we need. You might need to change the variants of various typography elements across the app after making the font sizes responsive.

src/theme.js
```diff
@@ -1,6 +1,6 @@
-import { createTheme } from '@mui/material';
+import { createTheme, responsiveFontSizes } from '@mui/material';
 
-const theme = createTheme({
+let theme = createTheme({
```
josueetcom, line 3: I suggest leaving this a const and wrapping the createTheme call instead.

josueetcom (approved):

PR #1046, opened 2022-04-21
Show scholarship card on right side of scholarship form

edwingl10:
## Motivation and Context

Fixes #1045. Fixes #985

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] User visible change (users will notice UI or functional changes)

## How Has This Been Tested?

- [x] Existing or new tests cover my changes.
- [ ] Manually tested locally or on the Render PR Server.

## Previewing Changes

1. Click the `onrender.com` URL the **render `bot`** posted below.
2. add / edit a scholarship
4. notice the detail card on the right when viewing through a bigger screen
5. try viewing on mobile

PR #1092, opened 2022-05-07
Fix ScholarshipCard issue and add exhaustive tests

### LINKED ISSUE #705: Component unit tests: ScholarshipCard
We need unit tests for the [UserHome](https://github.com/beyondhb1079/s4us/blob/main/src/components/ScholarshipDetailCard.jsx) component. Consider any edge cases (e.g. how do the grade levels render when there are none, one, and multiple? among other things). See the [Testing Library docs](https://testing-library.com/docs/) and existing tests for reference.

### LINKED ISSUE #1091: [Bug]: Cannot see issues with scholarships I added in dashboard
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

It should be complaining about the outdated school year here:
Screen Shot 2022-05-06 at 5 33 30 PM

### Steps To Reproduce

_No response_

### What devices are you experiencing the problem on?

Desktop

### What browsers are you seeing the problem in?

Chrome

josueetcom:
## Motivation and Context

Fixes #1091, which was a bug introduced by a one-liner change in #1046 🤦 . Lint issues couldn't be seen from the dashboard and any user could see lint issues regardless of permissions as long as they clicked on a scholarship.

Decided to add exhaustive tests for this component since it's fragile so fixes #705. Apparently we were showing the **Report Issue** button in `preview` mode too.

## Types of changes

## How Has This Been Tested?

### DIFF
--- a/src/components/ScholarshipCard.test.tsx
+++ b/src/components/ScholarshipCard.test.tsx
@@ -7,54 +7,146 @@ import ScholarshipAmount from '../types/ScholarshipAmount';
 import { clearFirestoreData, initializeTestApp } from '../lib/testing';
 import i18n from '../i18n';
 import { I18nextProvider } from 'react-i18next';
+import ScholarshipData from '../types/ScholarshipData';
+import GradeLevel from '../types/GradeLevel';
+import Ethnicity from '../types/Ethnicity';
+import State from '../types/States';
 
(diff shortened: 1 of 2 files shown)


Jobhdez (approved):
Looks good to me

PR #1098, opened 2022-05-10
Amount Textfield translations added

edwingl10:
## Motivation and Context

Adds translations to the AmountTextField. 

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. add a scholarship
3. change language
4. notice the amountTextField


josueetcom (approved):

src/components/ScholarshipAmountField.test.jsx
```diff
@@ -12,13 +14,19 @@ const formik = {
 
 function renderWithAmountType(type) {
   formik.values.amount.type = type;
-  return render(<ScholarshipAmountField formik={formik} />);
+  return render(
+    <Suspense fallback="loading">
+      <I18nextProvider i18n={i18n}>
+        <ScholarshipAmountField formik={formik} />
+      </I18nextProvider>
+    </Suspense>
```
josueetcom, line 22: It seems like we need this just about everywhere. We should consider refactoring out this pattern into our [testing.ts](https://github.com/beyondhb1079/s4us/blob/main/src/lib/testing.ts) library and instead import that `render` function here and everywhere (feel free to do this in a separate PR).

src/components/ScholarshipAmountField.test.jsx
```diff
@@ -12,13 +14,19 @@ const formik = {
...
+  );
 }
 
-test('renders select options', () => {
+test('renders select options', async () => {
   renderWithAmountType(AmountType.Fixed);
 
-  const select = screen.getByRole('button');
+  const select = await waitFor(() => screen.getByRole('button'));
```
josueetcom, line 29: I think you can do this:
```suggestion
  const select = await screen.findByRole('button');
```

Don't know if I forgot to suggest this elsewhere in other PRs.

josueetcom (approved):

src/components/ScholarshipAmountField.test.jsx
edwingl10: I tried that and it failed the tests because the only things rendered is `loading`

josueetcom: Weird, `findByRole` _should_ work.

PR #1099, opened 2022-05-10
validation translations

edwingl10:
## Motivation and Context

Adds translation to validation schema

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. add scholarship with errors
3. change language


public/locales/es/validation.json
```diff
@@ -6,8 +6,9 @@
   "validWebsite": "El sitio web debe ser una URL válida",
   "website": "Por favor ingrese al sitio web de becas",
   "option": "Por favor elige una opción arriba",
-  "validFixedAmount": "Ingrese una cantidad válida. Si no conoce la cantidad, seleccione 'Varía' arriba",
+  "validFixedAmount": "Ingrese una cantidad válida. Si no conoce la cantidad, sefleccione 'Varía' arriba",
```
josueetcom, line 9:
```suggestion
  "validFixedAmount": "Ingrese una cantidad válida. Si no conoce la cantidad, seleccione 'Varía' arriba",
```

src/validation/ValidationSchema.ts
```diff
@@ -2,51 +2,37 @@ import * as yup from 'yup';
...
-    .typeError('Please enter a valid date'),
-  description: yup
-    .string()
-    .required('Please enter the scholarship description'),
-  website: yup
-    .string()
-    .url('Website must be a valid URL')
-    .required('Please enter the scholarship website'),
+  name: yup.string().required('name'),
+  deadline: yup.date().required('deadline').typeError('validDate'),
```
josueetcom, line 6: Rather than having the translation keys here (the i18next-parser will not pick it up), how about making this a function that takes in the Translation object itself? E.g.
```ts
const validationSchema = (t) => ({
  name: yup.string().required(t('name')),
  deadline: yup.date().required(t('deadline')).typeError(t('validDate'))
  ...
})
```

public/locales/en/validation.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "Please enter the scholarship name",
+  "deadline": "Please enter the scholarship deadline",
+  "validDate": "Please enter a valid date",
```
josueetcom, line 4: Two things:
1. I suggest naming these `fieldPurpose`. E.g. `nameRequired`, `deadlineRequired`, `dateInvalid`, etc. 
2. I suggest sorting these. I think once you make the changes in the `ValidationSchema.ts` file you should be able to simply run `yarn i18next` and it'll find these keys and sort these files for you.

josueetcom (changes requested):

src/validation/ValidationSchema.ts
edwingl10: hmm doesnt seem to work for all keys. For example, `validAmount` stays as is and doesn't get translated

src/validation/ValidationSchema.ts
```diff
@@ -1,45 +1,55 @@
...
-          gpa === undefined ||
-          (gpa > 0 &&
-            gpa <= 4 &&
-            gpa.toString().match(/^[0-4](\.\d\d?)?$/) !== null)
-      ),
-  }),
-});
+const validationSchema: any = (t: (key: string) => string) =>
+  yup.object({
+    name: yup.string().required(t('validation:name')),
```
josueetcom, line 6: These all have the `validation:` prefix. Why not just pass in `useTranslation(['validation'])`?

edwingl10: the problem with this is that the validation doesn't update when the language changes. Only when the submit button is clicked again.

josueetcom: What if you do this in `ScholarshipForm.tsx`:
```js
const {t: validationT} = useTranslation(['validation']);
```

And then pass in `validationT`?

edwingl10: same problem :(

josueetcom: Ah but it does work if the users presses the **Next** button!

src/components/ScholarshipForm.jsx
```diff
@@ -44,11 +44,12 @@ function ScholarshipForm({ scholarship }) {
...
 
   const isDesktop = useMediaQuery((theme) => theme.breakpoints.up('md'));
-  const { t } = useTranslation(['scholarshipForm', 'common']);
+  const { t } = useTranslation(['scholarshipForm', 'common', 'validation']);
+  const { t: validationT } = useTranslation('validation');
 
   const formik = useFormik({
     initialValues: scholarship.data,
-    validationSchema,
+    validationSchema: validationSchema(validationT),
```
josueetcom, line 52: So I think, even if the errors don't change immediately, this _does_ affect future errors. We can probably just add a listener on the language change that looks like this:
```js
i18n.on("languageChanged", () => formik.errors.length && formik.validateForm());
```

edwingl10: hmm so it works, but it takes a while to re-load.

josueetcom: Another option would be to clear `formik.errors` when the translation change happens so we don't need to do this.

josueetcom: So basically:
```js
i18n.on("languageChanged", () => formik.setErrors({}));
```

josueetcom (approved):

PR #1100, opened 2022-05-11
Get rid of max amount in filter

edwingl10:
## Motivation and Context

Gets rid of max amount in the filterPanel. 

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. go to scholarships page
3. open amount filter accordion

### DIFF
--- a/src/components/AmountFilter.jsx
+++ b/src/components/AmountFilter.jsx
@@ -1,19 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { InputLabel, Grid, Slider } from '@mui/material';
-import { Remove as RemoveIcon } from '@mui/icons-material';
+import { InputLabel, Stack, Slider } from '@mui/material';
 import AmountTextField from './AmountTextField';
 
 export default function AmountFilter(props) {
-  const { min, max, onMinChange, onMaxChange } = props;
-  const error = max > 0 && max < min;
+  const { min, onMinChange } = props;
+  const error = min < 0;
 
   return (
-    <Grid
-      container
-      justifyContent="space-between"
-      sx={{ maxWidth: 300, m: 'auto' }}>
-      <Grid item>
+    <Stack alignItems="center" sx={{ maxWidth: 300, m: 'auto' }}>
+      <>
         <InputLabel>Min</InputLabel>
         <AmountTextField
           error={error}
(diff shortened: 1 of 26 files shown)


src/models/Scholarships.test.ts
```diff
@@ -270,19 +270,6 @@ test('scholarships.list - filters by minAmount', async () => {
   );
 });
 
-test('scholarships.list - filters by maxAmount', async () => {
```
josueetcom, line 273: I don't think I would worry about removing it from `src/lib` and `src/models` yet. It might prove useful later if we do identify a need for it. We can just remove it from the UI for now.

src/components/AmountFilter.jsx
```diff
@@ -1,57 +1,37 @@
...
-      <Grid item>
+    <Stack alignItems="center" sx={{ maxWidth: 300, m: 'auto' }}>
+      <>
         <InputLabel>Min</InputLabel>
         <AmountTextField
           error={error}
           value={min || ''}
           onChange={(e) => onMinChange(e.target.value)}
           placeholder="No min"
         />
```
josueetcom, line 19: I wonder if we should get rid of this too 🤔

src/components/AmountFilter.jsx
```diff
@@ -1,57 +1,37 @@
...
-          value={max || ''}
-          onChange={(e) => onMaxChange(e.target.value)}
-          placeholder="No max"
-        />
-      </Grid>
+      </>
 
       <Slider
-        value={[min || 0, max || 50000]}
+        value={min || 0}
```
josueetcom, line 23: Can the prop default be `0` instead?

src/components/AmountFilter.jsx
```diff
@@ -1,57 +1,37 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { InputLabel, Grid, Slider } from '@mui/material';
-import { Remove as RemoveIcon } from '@mui/icons-material';
+import { InputLabel, Stack, Slider } from '@mui/material';
 import AmountTextField from './AmountTextField';
 
 export default function AmountFilter(props) {
```
josueetcom, line 6: Maybe rename to MinAmountFilter?

src/components/FilterPanel.jsx
```diff
@@ -61,12 +59,10 @@ export default function FilterPanel({ onClose }) {
       comp: (
```
josueetcom, line 59: Change `Amount` key to `'Min Amount'`?

josueetcom (dismissed):

src/components/AmountFilter.jsx
edwingl10: and just leave the slider?

josueetcom: Hmm, my concern is around the space it takes :P.

What if we change the Stack direction, put the text field to the right, and use the `variant="standard"` instead?

src/components/FilterPanel.jsx
```diff
@@ -18,7 +18,7 @@ import {
   Stack,
 } from '@mui/material';
 import useQueryParams from '../lib/useQueryParams';
-import AmountFilter from './AmountFilter';
+import MinAmountFilter from './MinAmountFilter';
```
josueetcom, line 21: You forgot to `git add` this file.

edwingl10: it was added no?

josueetcom (approved):

Issue #1102, opened 2022-05-12
Improve page load performance


josueetcom:
[PageSpeed Insights](https://pagespeed.web.dev/report?url=https%3A%2F%2Fdreamscholars.org%2F) rates our site at 90/100 for Desktop performance and 36/100 for mobile performance:

<img width="994" alt="Screen Shot 2022-05-11 at 8 06 38 PM" src="https://user-images.githubusercontent.com/8023233/167983979-19ee39d8-be7e-4a21-8254-3f0f164b855b.png">

We should really address performance issues by leveraging code-splitting (lazy loading) and Suspense:
https://web.dev/code-splitting-suspense/

josueetcom:
For reference, here is the score after #1108 but before #1114:
<img width="1022" alt="Screen Shot 2022-05-14 at 3 13 41 AM" src="https://user-images.githubusercontent.com/8023233/168421478-04cb8217-310a-4f89-be1d-7b1720f1bd63.png">

Big gains in FCP and SI scores! #1114 should hopefully address LCP. Still gotta figure out how to improve TI and TBT :P

josueetcom:
Looks like we have a better score now after merging those two PRs:

<img width="1012" alt="Screen Shot 2022-05-15 at 1 03 26 AM" src="https://user-images.githubusercontent.com/8023233/168463243-8217b232-cbc1-40b3-8e04-6b9a046c3046.png">

TTI, LCP and TBT have dropped without affecting FCP and SI

josueetcom:
What remains to be improved has to do with the JS we send to the browser. There's a lot of unused code:
<img width="1434" alt="Screen Shot 2022-05-15 at 1 14 35 AM" src="https://user-images.githubusercontent.com/8023233/168463590-e1883a91-294e-4f75-a702-ef08f891d9cc.png">

Unfortunately most of it is has to do with Firebase:
- 15% (175KiB) is for the Firestore library. Loaded for ScholarshipsContext. Not needed for the public home page but is needed for other pages.
- 14% is for the analytics library (green box at the bottom)
- And the `iframe.js` file somewhere is for Firebase Auth. It doesn't always appear but on PageSpeed it says it's 263 KiB and not zipped. In practice though the mobile view in my inspector gives around ~86 KiB and it _is_ gzipped.
- Everything else there is pretty much needed as far as I can tell.

But, these _are_ some things we could try:
[...]

PR #1118, opened 2022-05-15
Move Header and Footer out of content-wrap div

josueetcom:
## Motivation and Context

This fixes two issues related to page load experience (#1102):
- `<Footer>` _should_ be outside of the content-wrap `<div>` so that the Flexbox container can push it away from content. I messed this up in a prior PR.
- But keep `<Footer>` _inside_ the same `<Suspense>` so it doesn't get rendered before the main content.

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. Refresh the page (maybe with simulated slow speeds).
3. The footer should not appear before the main content
4. Zoom out the most you can.
5. The footer should stay at the very bottom.
6. Compare these behaviors with what we currently have at https://dreamscholars.org to see the difference.

### DIFF
--- a/src/pages/PublicHome.tsx
+++ b/src/pages/PublicHome.tsx
@@ -38,7 +38,7 @@ function PublicHome(): JSX.Element {
 
       <ScholarshipsMadeSimple />
 
-      <Box sx={{ background: (theme) => theme.palette.background.paper }}>
+      <Box sx={{ backgroundColor: 'background.paper', flex: 1 }}>
         <HomeSection
           alignItems="center"
           direction="row"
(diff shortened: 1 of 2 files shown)


src/pages/PublicHome.tsx
```diff
@@ -38,7 +38,7 @@ function PublicHome(): JSX.Element {
 
       <ScholarshipsMadeSimple />
 
-      <Box sx={{ background: (theme) => theme.palette.background.paper }}>
```
josueetcom, line 41: A bit miscellaneous but meh, it is slightly cleaner :P

edwingl10 (approved):

PR #1130, opened 2022-05-18
State filter

### LINKED ISSUE #1128: State Filter
Allow users to filter by state: 
things that need to be done for this to be completed: 

- [ ] filter design
- [ ] filter functionality
- [ ]  parameter support

### LINKED ISSUE #1129: State filter design

edwingl10:
## Motivation and Context

Fixes #1128 

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. go to scholarships page 
3. add state filters


src/components/CustomAutocomplete.tsx
```diff
@@ -28,7 +28,8 @@ export default function AutocompleteFilter(
       open={isOpen && !limitReached}
       disabled={limitReached}
       onInputChange={(e, val) => setIsOpen(val.length > 0)}
-      renderTags={() => null}
+      forcePopupIcon={false}
```
edwingl10, line 31: helps with the previous bug I had.
Possible [explanation.](https://stackoverflow.com/questions/61256685/material-ui-autocomplete-styling)

src/components/CustomAutocomplete.tsx
```diff
@@ -51,12 +52,12 @@ export default function AutocompleteFilter(
...
+CustomAutocomplete.propTypes = {
   limitReached: PropTypes.bool,
   placeholder: PropTypes.string,
 };
 
-AutocompleteFilter.defaultProps = {
+CustomAutocomplete.defaultProps = {
   limitReached: false,
   placeholder: undefined,
 };
```
josueetcom, line 63: Why does `CustomAutoCompete` have propTypes and defaultProps?

src/components/FilterPanel.jsx
```diff
@@ -50,15 +53,16 @@ export default function FilterPanel({ onClose }) {
...
 
   const filters = {
     'What are you studying?': {
       comp: (
         <>
-          <AutocompleteFilter
+          <CustomAutocomplete
             freeSolo
-            value={majors}
+            value={majors || []}
```
josueetcom, line 65: Why is this necessary? Also why not have the `|| []` logic in the component itself?

src/lib/useQueryParams.ts
```diff
@@ -95,6 +96,19 @@ export default function useQueryParams(
         delete params.majors;
       }
     }
+
+    if (states !== undefined) {
```
josueetcom, line 100: You need to update the tests

src/models/Scholarships.ts
```diff
@@ -97,6 +97,7 @@ export interface FilterOptions {
   maxAmount?: number;
   grades?: GradeLevel[];
   majors?: String[];
```
josueetcom, line 99: Tangential but can you lower-case the `s` here for consistency?
```suggestion
  majors?: string[];
```

src/components/CustomAutocomplete.tsx
josueetcom: What does `forcePopupIcon` do? I understand it does something with the up/down icon on the right but I couldn't make heads or tail of the API description when it says "Force the visibility display"

src/components/CustomAutocomplete.tsx
edwingl10: yeah I guess it doesn't need them.

src/components/FilterPanel.jsx
edwingl10: because it was complaining about the following: `MUI: A component is changing the uncontrolled value state of Autocomplete to be controlled.`

edwingl10: I tried doing it in the component itself but interestingly, it wouldn't let me type in the state filter. It's the same bug that we have in the scholarship form

josueetcom (approved):

src/components/CustomAutocomplete.tsx
edwingl10: I think besides visibility, the styles are also not forced in this case.

Issue #1174, opened 2022-05-22
[Bug]: Investigate leaked open handles


josueetcom:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Running `yarn test` it complains about resources not being properly cleaned up and suggests using `--detectOpenHandles` to determine what it is. However that doesn't do anything for us. I did my own investigation and it seems to be coming from the tests in the `src/models` folders, particularly from any tests there that interact with Firestore. My guess is that either the tests aren't given enough time to tear down (perhaps the `@jest-environment node` comment at the top has something to do with it?) or there's a leakage in the Firestore library. I doubt it's the latter though as other tests that use Firestore don't leak.

### Steps To Reproduce

1. Run `yarn test src/`
2. Notice it complains about open handles
3. Run the command again with `--detectOpenHandles`
4. Notice it doesn't catch anything

### What devices are you experiencing the problem on?

Desktop

PR #1175, opened 2022-05-22
Address --detectOpenHandles issue so test check runs faster

josueetcom:
## Motivation and Context
Originally we added `--detectOpenHandles` because there were statements about it in the test runs. However, it never gives us anything useful. I did however track down that the tests that supposedly leak open handles are in `src/models`, particularly around interacting with Firestore (see #1174 for context).

Anyhow, it turns out the issue was the `@jest-environment node`. Removing that allows us to run `yarn test` without any complaints about leaked handles.

Also with this, the validation test check might be faster (at least it is on my machine).


Jobhdez (approved):
Looks good

PR #1180, opened 2022-05-28
School filter 

### LINKED ISSUE #1179: [FR]: Schools filter
### Is there an existing feature request for this?

- [X] I have searched the existing feature requests.

### What is your feature idea?

Allow users to filter by schools 

### What alternatives have you considered?

_No response_

edwingl10:
## Motivation and Context

Fixes #1179 

## Types of changes

## How Has This Been Tested?

## Previewing Changes

2. go to scholarships page
3. fitler by schools


src/components/FilterPanel.tsx
```diff
@@ -44,6 +45,7 @@ export default function FilterPanel({
   const [grades, setGrades] = useState(params.grades);
   const [majors, setMajors] = useState(params.majors);
   const [states, setStates] = useState(params.states);
+  const [schools, setSchools] = useState(params.schools);
```
josueetcom, line 48: If we do this we could probably get rid of the `|| []` elsewhere.
```suggestion
  const [grades, setGrades] = useState(params.grades || []);
  const [majors, setMajors] = useState(params.majors || []);
  const [states, setStates] = useState(params.states || []);
  const [schools, setSchools] = useState(params.schools || []);
```

src/components/FilterPanel.tsx
```diff
@@ -44,6 +45,7 @@ export default function FilterPanel({
...
   const location = useLocation();
 
   useEffect(() => {
```
josueetcom, line 51: Could you remind me why we have this effect? This really seems like something `useQueryParams` should be doing. I'm surprised we haven't filed a bug to address this.

src/lib/useQueryParams.ts
```diff
@@ -109,6 +110,15 @@ export default function useQueryParams(
...
     }
+
+    if (schools !== undefined) {
+      params.schools = Array.isArray(schools)
+        ? Array.from(new Set(schools)).filter((s) => s.length > 0)
+        : [];
+      if (params.schools.length === 0) {
+        delete params.schools;
+      }
+    }
```
josueetcom, line 121: I actually wouldn't worry about pruning unknown schools (or majors for that matter) since there are more schools and majors than we're aware about.

src/models/Scholarships.ts
```diff
@@ -191,6 +192,11 @@ class Scholarships extends FirestoreCollection<ScholarshipData> {
                 data.requirements?.states,
                 opts.states
               ) &&
+              // school filter
+              requirementMatchesFilter(
+                data.requirements?.schools,
+                opts.schools
```
josueetcom, line 198: Lower case these so we're not case sensitive. Or maybe `requirementMatchesFilter()` should do `.map((s) => s?.toLowerCase() || s)` on both lists so we lower case both sides when it's dealing with string lists.

josueetcom (changes requested):

src/components/FilterPanel.tsx
```diff
@@ -134,6 +137,35 @@ export default function FilterPanel({
...
+              key={school}
+              onClick={() =>
+                setSchools(schools.filter((s: string) => s !== school))
+              }
+              sx={{ mx: 1, mt: 1 }}
+            />
+          ))}
+        </>
+      ),
+      expanded: false,
```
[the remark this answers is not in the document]
edwingl10: I had set it because it was complaining about it.

src/components/FilterPanel.tsx
edwingl10: so the filters dont get applied right away when the user choose them.

src/components/FilterPanel.tsx
edwingl10: problem is that it complains that they are `undefined`. I checked in the console and they go from being an empty array to undefined. 

I could do this change inside the component itself.

src/lib/useQueryParams.ts
edwingl10: It just prunes if the schools is an empty array or is an array with empty values. Like the majors params.

public/locales/en/filters.json
```diff
@@ -7,6 +7,7 @@
   "deadlineEarliestToLatest": "Deadline (Earliest to Latest)",
   "deadlineLatestToEarliest": "Deadline (Latest to Earliest)",
   "enterMajorFilter": "Enter a major to filter by",
+  "enterSchoolFilter": "Enter a school to filter by",
   "enterStateFilter": "Enter a state to filter by",
```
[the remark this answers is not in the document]
edwingl10: I think [interpolation ](https://www.i18next.com/translation-function/interpolation)might be a better choice but I think we would have to do something like 
```
t('key', {word: t('school')));
```

so that the phrase AND the word get translated.

src/components/FilterPanel.test.tsx
```diff
@@ -45,12 +49,41 @@ test('message change when filter option chosen', async () => {
...
+    name: 'University of California Irvine (CA)',
+  });
+  fireEvent.click(school);
+
+  expect(
+    screen.getByText('University of California Irvine (CA)')
+  ).toBeInTheDocument();
+});
+
+test('Min Amount slider reflects input value', async () => {
```
edwingl10, line 76: test added to meet threshold 😛

src/components/FilterPanel.tsx
```diff
@@ -40,32 +40,19 @@ export default function FilterPanel({
...
-      arrayFormat: 'bracket-separator',
-      arrayFormatSeparator: ',',
-      parseNumbers: true,
-    });
-
-    setMinAmount(query.minAmount);
-    setGrades(query.grades);
-    setMajors(query.majors);
-    setStates(query.states);
-  }, [location]);
```
edwingl10, line 60: I dont remember why we had this but I dont think we need it either way.

public/locales/en/filters.json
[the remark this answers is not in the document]
josueetcom: Ohh that looks nice! We should maybe do that here then

src/lib/useQueryParams.ts
josueetcom: Ah you're right! I forgot about that haha

src/components/FilterPanel.tsx
josueetcom: The effect runs when the location gets changed, e.g. let's say the user clicks on one of the chips from the FilterBar. What I don't remember is why `useQueryParams` doesn't take care of this itself.

josueetcom: I found the [comment](https://github.com/beyondhb1079/s4us/pull/969#discussion_r839102427) where I commented on this but I guess I never asked why this was needed in the first place 🤔 . Yeah I don't think this is necessary as `useQueryParams` uses `useLocation` which should cause a re-render and a re-initialization of state.

src/components/FilterPanel.tsx
```diff
@@ -40,32 +40,19 @@ export default function FilterPanel({
...
+  const [minAmount, setMinAmount] = useState(params.minAmount || 0);
```
josueetcom, line 43: This is maybe the one that shouldn't be defaulted as there is a semantic difference between explicit `0` and `undefined` :/

src/components/FilterPanel.tsx
```diff
@@ -92,11 +79,11 @@ export default function FilterPanel({
     [t('minAmount')]: {
       comp: (
         <MinAmountFilter
-          min={minAmount ?? 0}
```
josueetcom, line 95: I'm curious why this was like this before... You'd think with `undefined` it'd default to 0 or something 🤔

josueetcom (approved):

src/components/FilterPanel.tsx
edwingl10: It's so the console wouldn't complain about changing an uncontrolled component.

src/components/FilterPanel.tsx
edwingl10: It wouldn't matter though because in the `minAmountFilter` we would do `minAmount ?? 0`

public/locales/en/filters.json
```diff
@@ -6,18 +6,19 @@
...
   "deadlineEarliestToLatest": "Deadline (Earliest to Latest)",
   "deadlineLatestToEarliest": "Deadline (Latest to Earliest)",
-  "enterMajorFilter": "Enter a major to filter by",
-  "enterStateFilter": "Enter a state to filter by",
   "filters": "Filters",
   "filtersCurrentlyApplied": "Your filters are currently applied.",
   "gradeLevel": "Grade Level",
   "highSchool": "High School",
   "limitReached": "Limit reached",
+  "major": "major",
```
josueetcom, line 14: Should this be capitalized like the other two?

I wonder if we should formally establish a paradigm for the capitalization of single words

josueetcom (approved):

public/locales/en/filters.json
edwingl10: The reason the others are capitalize is because they are used as the Filter title and major is only used for the placeholder.

edwingl10: I'll capitilize though, for consistency reasons.

